PR #122, opened 2017-02-21
Strategy Refactoring

IARI:
removed duplicates and encapsulated a context for strategies


IARI:
For the record: Daniel doesn't like the context object and wants to refactor all the strategy code at some unspecified point in the future (which is probably goig to be to late for my master thesis).

PR #165, opened 2017-05-10
Ultimate Server and Interactive Interface

IARI:
This PR contains the Ultimate Server and Interactive Interface and introduces the Active-Verification Paradigm for Ultimate.

It includes in implementation for the TraceAbstraction Plugin.

---

In the following, Assume you want to use the interactive interface in your existing Ultimate Plugin _AnUltimatePlugin_.

The Dependency Graph:

architecture

## Plugins:

- **UltimateServer** This is a new controller. It starts a TCP Server that listens for a Client implementing the _ActiveUltimate Protocol_ which is protobuf based.
- **_AnUltimatePlugin_** Your existing Ultimate Plugin: this could be any Source, Analysis, Generator or Outpug Plugin.
- **Library-Protobuf** Protobuf for Java requires a runtime library. that's that.
- **Library-InteractiveModel** In here the core interface for Interaction between a Client and your Plugin is defined. Your Plugin only requires this Library.
- **Library-Interactive**_**AUPlugin**_ In here you define converters for custom data types that you want to exchange with the Client.
- **Library-InteractiveCommon** This defines some common datatypes and converters (String messages, boolen (ask a client for confirmation)) and stuff like that, which might generally be of interest for users of this interactive interface.

---
This change is [Reviewable](https://reviewable.io/reviews/ultimate-pa/ultimate/165)


danieldietsch:
Many of the codacy checks can probably be ignored (at least the ones concerning generated classes). But you should fix the maven build.

IARI:
@danieldietsch ins response to your comment above:
In the long Term, I suggest to remove the generated files and integrate protoc with the build chain.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
```diff
@@ -287,8 +287,14 @@ public IcfgProgramExecution getRcfgProgramExecution() {
 	public String errorLocs() {
 		return mErrorLocs.toString();
 	}
-
+	
 	public final Result iterate() {
```
IARI: This is not beautiful.
I want to send the result of the Cegar loop to the client instantly after the loop is done.

- Originally, I did this in TraceAbstractionStarter (after the call of iterate)
- now its here - in a second method, because the original method has so many exit points.

Reason its not in TraceAbstractionStarter anymore:
I got rid of the interactive interface there, less footprint.

Alternative possibility:
introduce a `final Result result` in the original iterate method.

if anyone has a better suggestion, or a clear preerence for another solution please tell me.

danieldietsch:
I merged this pull request manually via squashing rebasing.

PR #174, opened 2017-05-16
Decreased Interactive Footprint and improved Strategy Selection

IARI:
The name of commits that touch TA Files ends with with (TAFiles)


IARI:
Great - when i start Pull requests all the olds commits get added.
I guess I have to make a new branch then.

PR #560, opened 2021-04-22
Fix thread IDs in C translation

maul-esel:
Previously, each `pthread_create()` call was translated to a Boogie statement `fork X ...` for a fixed constant `X` (e.g. `fork 0 ... `). This is unsound if the statement can be executed multiple times, as several threads end up with the same thread ID and `pthread_join()` calls then become ambiguous.

Instead, we now maintain a global counter that is used as thread ID and then incremented.


danieldietsch (approved):
Looks good to me, but I dont know about the general idea -- @Heizmann may be the judge of that ^^

Heizmann (approved):

PR #569, opened 2021-07-01
Support for multiple results from one CEGAR loop

### LINKED ISSUE #323: Improve "stop after first violation" behavior
If "stop after first violation" is false, we currently check each error location in a separate trace abstraction run. 

We can be faster if we start with checking all assertions and keep checking as long as none is violated. 
If one is violated, we could do one of the following. 
 * Remember which assertions are safe and which are unsafe, generate results for them, restart trace abstraction with the remaining assertions for which the status is not yet determined. 
 * Generate an automaton that recognizes all traces that lead to the violating location and refine with this automaton. Continue as normal. 

Both methods may have their own advantages and disadvantages, we should implement at least one of them.

danieldietsch:
This PR adds support for multiple results in one CEGAR loop and for re-using the abstraction for multiple error locations, even if some of them are reachable. 
When an error location is reachable, CEGAR now refines by subtracting a simple error automaton that recognizes all traces that contain the transition to this error location. This can be combined with fault localization and error partitioning. 
This PR adds two new settings: ONE_ERROR_PER_CEGAR is a boolean that controls if a CEGAR instance should be used for more than one error location, and ERROR_AUTOMATON_MODE controls what kind of automaton should be used to remove reachable counterexamples (SIMPLE_ERROR_AUTOMATON guarantees removal in one iteration, the other two can be used in conjunction with fault localization). 

This approach requires a more complicated handling of timeouts, in particular during trace checks and proof generation. To prevent that one error location takes the complete timeout or the timeout per analysis, the trace analysis phase (``isCounterexampleFeasible``) should be limited to 1/<# of remaining error locations> * <remaining time> . Implementing this required some changes to the current timers, and timeout handling in additional places. 

As a side-effect, this PR removes the ``VerificationResult`` from various CEGAR statistics,i.e., a CEGAR loop does not have a single result anymore. 

This PR closes #323.


danieldietsch:
Regression tests and a complete SV-COMP run do not reveal any more problems. I will merge this PR this afternoon.

PR #570, opened 2021-07-11
Immutable Sets

maul-esel:
As discussed, implements a class `ImmutableSet` that caches the hash code of a set and prevents all modifications of the set.

Open points:
- [x] Future of `de.uni_freiburg.informatik.ultimate.util.datastructures.CachedHashSet`
     This class is similar in that it caches the hash. However, it allows modifications. The only usage is in `de.uni_freiburg.informatik.ultimate.util.datastructures.UnionFind`. Can we replace this by `ImmutableSet` ?
- [x] Replacing calls of `Collections.unmodifiableSet` /  `Collections.unmodifiableCollection`
     This is not a 100% straightforward search-and-replace, because these methods allow the underlying set to be modified, and these modifications show up in the returned set. We might have to look at individual classes to see if and how we can replace these calls with `ImmutableSet`.
- [x] Ensuring elements of sets and keys of maps are immutable
      Beyond improving performance due to cheaper `hashCode()`, this should also prevent (or perhaps uncover) bugs.
- [x] Is it worth computing the hash lazily instead of eagerly?

/cc @danieldietsch @Heizmann @schuessf


Heizmann (commented):
Two points regarding: ImmutableSet#copy 
* P1: Maybe Java already supports this: 
https://docs.oracle.com/javase/10/docs/api/java/util/Set.html#copyOf(java.util.Collection)
I was however unable to find out if it caches the hashCode.
* P2: A problem is that it uses only HashSets. Maybe a user wants to keep his kind of set. (The only scenario that I have in mind is that the user wants to preserve his iteration order.) A solution could be to use Set#clone. (Can only be implemented with warnings)

danieldietsch:
* P1: no, this does not what we want: it does not create "our" type, and the type it creates does not cache hash codes
* P2: true, it would be nice to have an additional method that supplies the underlying type. I would not use ``clone()`` because its semantics are too murky for me -- in particular as we want to allow all implementors of ``Set``, and this does not force an implemtation of ``clone()``. E.g., SMTInterpols ``CuckooHashSet`` does not have a ``clone()`` implementation. 

Then: 
* P3 I had a look at ``CachedHashSet``. I think we can replace it. 
* P4 (Replacing calls of Collections.unmodifiableSet / Collections.unmodifiableCollection): Should we just do that in this branch until we are satisfied? 
* P5 (Ensuring elements of sets and keys of maps are immutable): I agree, but it will widen the gap between test and benchmark performance even more. One way to deal with it would be to disable it during certain runs with ``-da:de.uni_freiburg.informatik.ultimate.util.datastructures.ImmutableSet`` (see https://docs.oracle.com/javase/7/docs/technotes/guides/language/assert.html), but this introduces other problems. I would say we keep the assertion for some time and see what kind of bugs we find. 
* P6 (lazy hash computation) Yes, I think this would be better

And even more: 
* P7: It might be possible use the precomputed hashcode in ``equals`` to gain more performance during set comparisons -- but we dont have a lot of them, so... 
* P8: What about ``ImmutableMap``?

maul-esel:
* P5: By "ensuring elements of sets and keys of maps are immutable", I meant we should look through all usages of `Set<Set<?>>` and `Map<Set<?>, ?>` and consider wrapping the elements / keys in `ImmutableSet` instances, possibly even changing the type to `Set<ImmutableSet<?>>` / `Map<ImmutableSet<?>, ?>` (analogous to looking at usages of `Collections.unmodifiableSet`).

   This is separate from the `assert` statement in `hashCode()`.

MartinSpiessl:
Stupid question: why not use https://guava.dev/releases/19.0/api/docs/com/google/common/collect/ImmutableSet.html?

danieldietsch:
@MartinSpiessl : because Guava's ImmutableSet does not cache the hashcode of its elements.

maul-esel:
@Heizmann Among the remaining cases of `Set<Set<?>>` are [LookaheadPartitionConstructor](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/minimization/LookaheadPartitionConstructor.java) and [FinitePrefix2PetriNet.TransitionSet](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/FinitePrefix2PetriNet.java#L539-L572) in the automata library. Both of these are unused, and `LookaheadPartitionConstructor` is marked as deprecated. Can we just remove them?

maul-esel:
As I see it, the following points are open:

* P2 : We can add that. It's just as easy though to just call `ImmutableSet.of(new LinkedHashSet<>(set))`, i.e., do the copying yourself, rather than something like `ImmutableSet.copyOf(set, LinkedHashSet::new)`.
* P4 / P5 :
  > Should we just do that in this branch until we are satisfied?
  
  At what point are we "satisfied"? I've done this in some parts of the code, but going through the entire code base will take too much time. Are there areas where this is particularly relevant?
* P8 : I think we have rather few immutable maps, and even fewer on which `hashCode()` is called.

danieldietsch:
* P2: Makes sense. Perhaps just add an ``ofOrdered`` that uses ``LinkedHashSet``?
* P4/P5: I agree -- perhaps we should try and merge this as fast as possible. Il'l run some tests and report back with the results tomorrow.
* P8: Makes sense.

Heizmann:
> @Heizmann Among the remaining cases of `Set<Set<?>>` are [LookaheadPartitionConstructor](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/minimization/LookaheadPartitionConstructor.java) and [FinitePrefix2PetriNet.TransitionSet](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/FinitePrefix2PetriNet.java#L539-L572) in the automata library. Both of these are unused, and `LookaheadPartitionConstructor` is marked as deprecated. Can we just remove them?

I can't decide this soon. Could you maybe add a TODO in these classes?

Heizmann:
P2: ordered (LinkedHashSet) and memory efficient (HashSet) might be sufficient
P4: main branch
P5: Keep it for some time is a good idea. Maybe add a boolean flag and a TODO with a pointer to this issue.
P7: I agree that set comparisons occur typically only while adding sets to sets or maps.
P8: ImmutableMaps would be definitely helpful for the rank-based complementation of Büchi automata and are possibly helpful in some data structure for nested word automata. I don't know if they would be helpful in other places or if the improvement is significant.

PR #576, opened 2021-12-08
Add typechecking for attributes on Boogie assume and assert statements

martin-neuhaeusser:
Boogie assume and assert statements may have attributes. However, these attributes
are not type-checked. In case the attributes contain expressions, this results in
a null value in their type field and causes a null-pointer dereference during
Boogie preprocessing.


danieldietsch (approved):

danieldietsch:
Thank you for the fix

PR #579, opened 2022-01-10
Witness fix: Use "assumption.resultfunction"

maul-esel:
Per witness specification, assumptions must not contain function call expressions. Instead, a variable `\result` can be used and the corresponding function can be given via the key `assumption.resultfunction`.

Using functions directly caused CPAChecker to reject the witness with `ERROR (invalid witness file)`. In SV-COMP'22 (for Automizer), CPAChecker crashed with this error 32 times [1] -- many (but not all) of these seem to be due to this bug.

The solution implemented below feels slightly hacky; feel free to suggest improvements.

[1] https://sv-comp.sosy-lab.org/2022/results/results-verified/uautomizer.results.SV-COMP22.All.table.html#/table?hidden0=1,2,3,5,4&hidden3=1,2,4,3&hidden=1,2,4,5,6,7,8,9,10,11,12&filter=0(0*status*(category(notIn(error,unknown)))),3(0*status*(status(in(ERROR%20%28invalid%20witness%20file%29))))


danieldietsch:
I am not sure if we want to implement that; we should discuss it with @Heizmann. 

With the current witness specification, I would prefer not to use the `\return` keyword in assumptions at all, because I do not believe there are many situations where it is beneficial to (our) verification. We should instead lobby for support of contracts and of something like `\old` in the witness specification. 

Did you try what happens with CPAChecker if we just remove these assumptions from our witnesses?

maul-esel:
If we omit such assumptions, the witness should be considered syntactically correct. I seem to remember I tried this for one of them, and CPAChecker then successfully validated it... but let me check that again (tomorrow), for a few of the witnesses.

maul-esel:
OK, this took much longer, but I finally got to it. For all the witnesses I was able to reproduce with a timeout of 300s, it made no difference to CPAChecker whether we use the version from this PR (using `assumption.resultfunction`) or we just omit return values of C function calls from our witnesses.

I don't know how it is for the other validators, e.g. those based on executing counterexamples. But if you prefer not to implement this, I can change the PR to always omit return values.

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/graphml/GeneratedWitnessEdge.java
```diff
@@ -175,11 +197,15 @@ public String getSourceCode() {
 	}
 
 	private void appendValidExpression(final E var, final E val, final StringBuilder sb) {
-		final String varStr = mStringProvider.getStringFromExpression(var);
-
-		if (varStr.contains("\\") || varStr.contains("&")) {
-			// is something like read, old, etc.
-			return;
+		final String varStr;
+		if (var == mResultFunction) {
```
danieldietsch: It might be better to use Objects.equals() here; also, `var` is a keyword ;)

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/graphml/GeneratedWitnessEdge.java
```diff
@@ -116,6 +123,13 @@ public String getAssumption() {
 
 		final StringBuilder sb = new StringBuilder();
 		for (final E var : mState.getVariables()) {
+			if (mStringProvider.isProcedureCall(var) && var != mResultFunction) {
```
danieldietsch: Same as below; although the `var` here is probably from before `var` was a keyword

danieldietsch:
I have no strong feelings either way, but perhaps the others can say something. But having syntactical correct witnesses is definitively desirable -- strange that the linter did not catch that.

maul-esel:
I think the linter mostly checks the XML structure, it probably does not parse the expressions. Might be a feature to add.

schuessf:
I think it would be a good idea to merge this, to achieve syntactically more correct witnesses, with 9319baf by just ommiting the function calls (so the title is not quite correct anymore). We should just make sure to add this check also for YAML-witness, when supported 🙂

schuessf (approved):

PR #584, opened 2022-04-25
Refactor BuchiAutomizer (Termination)

schuessf:
The goal of this PR is to allow to use `BuchiAutomizer` with an abstraction of a different type than `INestedWordAutomaton`

- The former `BuchiCegarLoop` was split into two classes:
  - `AbstractCegarLoop` that has several abstract methods (mainly refinement and emptiness check) that are dependant on the actual type of the abstraction
  - `BuchiAutomatonCegarLoop` that implements those methods for the abstraction of the type `INestedWordAutomaton`
- Similar to #583: Instances of `AbstractCegarLoop` are created in `BuchiCegarLoopFactory` and get the initial abstraction passed via the constructor using `IInitialAbstractionProvider`. The time needed for the initial abstraction is also included in the statistics.
- Other minor refactoring

PR #589, opened 2022-07-01
BuchiAutomizer for concurrent programs

schuessf:
This pull request allows to use the `BuchiAutomizer` for concurrent programs. It is based on the work from pull request #584 and consists of the following changes:
*  It is now possible to analyse the termination of concurrent programs represented by finite automata (the "product automata") using the already exisiting `BuchiAutomatonCegarLoop`. This needed some minor changes in order to work as expected.
* This requires the petrification of the existing icfg with an increasing number of thread instances (similar to the trace abstraction). In order to detect insufficient thread instances, the petrification was slightly modified and now adds self-loops to the insufficient-thread-location
* Additionally `AbstractBuchiCegarLoop` was refactored. Now it deals with the `LassoCheck` and creates interpolant automata out of it (including the iterations over different construction styles), instead of the implementing class. This makes it easier to implement this class, since you can mostly use the automata-theoretic operations and should also reduce code duplication in the future.


trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiInterpolantAutomatonBuilder.java
```diff
@@ -0,0 +1,209 @@
...
+		final NestedWord<LETTER> stem = counterexample.getStem().getWord();
+		final NestedWord<LETTER> loop = counterexample.getLoop().getWord();
+		final NestedWordAutomaton<LETTER, IPredicate> result =
+				new NestedWordAutomaton<>(new AutomataLibraryServices(mServices), vpAlphabet, emptyStateFactory);
+		if (stem.length() == 0) {
+			result.addState(true, true, honda);
+		} else {
+			result.addState(true, false, precondition);
+			for (int i = 0; i < stemInterpolants.length; i++) {
+				addState(stemInterpolants[i], result);
```
danieldietsch, line 96: why do you call `addState` sometimes directly on the NWA and sometimes via the `addState` wrapper?

danieldietsch:
Can you check why the various checks failed? 

It will take some time to review a PR this large -- I try and do it piece-wise. 

In general, it would be nice to squash large PRs so that they dont have so many commits, e.g., combine all cc fixes into one, combine new examples into one, try and keep features together (i.e., if you have something that took a couple of commits with various local fixes and rewrites, it would be nice to have it in one commit).

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiInterpolantAutomatonBuilder.java
schuessf: Not sure, this is just some old unchanged code that was moved to an extra class.

schuessf:
> Can you check why the various checks failed?

The log [here](https://jenkins.sopranium.de/job/Ultimate/job/Ultimate/view/change-requests/job/PR-589/2/console) yields this exception:
`Unable to read repository at https://monteverdi.informatik.uni-freiburg.de/nexus/content/repositories/eclipse-4.17. https://monteverdi.informatik.uni-freiburg.de/nexus/content/repositories/eclipse-4.17 is not a valid repository location`
I am not sure, why it failed this morning, it did build successfully right after the creation of this PR.

> It will take some time to review a PR this large -- I try and do it piece-wise.
> 
> In general, it would be nice to squash large PRs so that they dont have so many commits, e.g., combine all cc fixes into one, combine new examples into one, try and keep features together (i.e., if you have something that took a couple of commits with various local fixes and rewrites, it would be nice to have it in one commit).

Yes, sorry, required more refactoring than I expected. Basically the changes in [BuchiCegarLoopFactory.java](https://github.com/ultimate-pa/ultimate/pull/589/files#diff-ebb8b35ee436d840b1d75dd4ed329a4ef1f8b823e7a2e1456896b91c6984b87a) cover the first point, [BuchiAutomizerObserver.java](https://github.com/ultimate-pa/ultimate/pull/589/files#diff-39347dab6222b2d1a3a0b956e0e406c3ae09e26c6783c80d76f077145ceabcc0) the second point and the rest (which is obviously a lot!) is just refactoring that should make things easier for future changes (i.e. the last point).

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
```diff
@@ -131,41 +138,71 @@ public boolean process(final IElement root) throws IOException {
...
+				final IcfgPetrifier icfgPetrifier = new IcfgPetrifier(mServices, icfg, numberOfThreadInstances);
+				mServices.getBacktranslationService().addTranslator(icfgPetrifier.getBacktranslator());
+				final IIcfg<IcfgLocation> petrified = icfgPetrifier.getPetrifiedIcfg();
+				final Set<IcfgLocation> insufficientThreadLocs = new HashSet<>(
+						petrified.getCfgSmtToolkit().getConcurrencyInformation().getInUseErrorNodeMap().values());
+				final AbstractBuchiCegarLoop<IcfgEdge, ?> currentCegarLoop =
+						factory.constructCegarLoop(petrified, witnessAutomaton);
+				final Result currentResult = currentCegarLoop.runCegarLoop();
+				if (currentResult != Result.NONTERMINATING
+						|| !containsInsufficientThreadLocation(currentCegarLoop.getCounterexample(),
```
maul-esel, line 157: I'm not sure I understand this check. If the result is that the program terminates (1st disjunct true), why do we not increase the number of threads?

maul-esel: Also, a less crucial comment: If a nonterminating execution contains an "insufficient threads" location, do we always need to increase the thread count? Particularly, if the location appears only in the stem?

**Edit:** Ah, I see `containsInsufficientThreadLocation` looks only at the loop. Perhaps rename the method to make that clear?

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
```diff
@@ -324,9 +358,16 @@ private void reportLTLPropertyIsViolated(final AbstractBuchiCegarLoop<IcfgEdge,
...
 				IcfgProgramExecution.create(loop, partialProgramStateMapping, new Map[loop.size()]);
-		reportResult(new LTLInfiniteCounterExampleResult<>(position, Activator.PLUGIN_ID,
-				mServices.getBacktranslationService(), stemPE, loopPE, ILocation.getAnnotation(position),
-				ltlAnnot.getUltimateLTLProperty()));
+		reportResult(new LTLInfiniteCounterExampleResult<>(getProgramPoint(counterexample), Activator.PLUGIN_ID,
+				mServices.getBacktranslationService(), stemPE, loopPE, ltlAnnot.getUltimateLTLProperty()));
+	}
+
+	private static IcfgLocation
+			getProgramPoint(final NestedLassoRun<? extends IIcfgTransition<?>, IPredicate> counterexample) {
```
maul-esel: Maybe make `honda` part of the name? The ctex contains many program points.

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
```diff
@@ -324,9 +358,16 @@ private void reportLTLPropertyIsViolated(final AbstractBuchiCegarLoop<IcfgEdge,
...
+		final IPredicate pred = counterexample.getLoop().getStateAtPosition(0);
+		// Return any program point from the predicate
+		// There is not really a "correct" answer here for concurrent programs
```
maul-esel: What role does the program point play in the counterexample-results? Should these be changed to take an `IPredicate` (i.e., an `SPredicate` or `MLPredicate`) instead?

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/cegar/AbstractBuchiCegarLoop.java
```diff
@@ -204,47 +242,75 @@ public AbstractBuchiCegarLoop(final IIcfg<?> icfg, final RankVarConstructor rank
 		mRefinementStrategyFactory = new StrategyFactory<>(mLogger, mPref, taCheckAndRefinementPrefs, icfg,
 				mPredicateFactory, mDefaultStateFactory, transitionClazz);
 		mAbstraction = initialAbstraction;
+		mInterpolantAutomatonBuilder = new BuchiInterpolantAutomatonBuilder<>(mServices, mCsToolkitWithRankVars,
+				SIMPLIFICATION_TECHNIQUE, XNF_CONVERSION_TECHNIQUE, predicateFactory, mDefaultStateFactory);
+		mBiaConstructionStyleSequence =
+				baPref.getEnum(BuchiAutomizerPreferenceInitializer.LABEL_BIA_CONSTRUCTION_STRATEGY,
+						BuchiInterpolantAutomatonConstructionStrategy.class).getBiaConstrucionStyleSequence(baPref);
+		mAutomataMinimizationAfterFeasbilityBasedRefinement = baPref.getEnum(
```
maul-esel: typo: Feasbility

maul-esel (commented):
I really like the idea of this PR.

I took a look, but can't give any guarantees (I'm unfamiliar with this part of the code). I left a few comments and some questions.

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
schuessf: * I added a self loop to the insufficient thread loactions in order to find, if we need more thread instances. If the program terminates, we know that we did not reach this location.
* If the program reaches this location, it cannot leave it anymore (the only outgoing edge is a self loop). I could also add the stem to the check, it does not make a difference.

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
schuessf: If the program does not terminate, we get a `NonterminatingLassoResult` that specifies the line, where the loop starts. For sequential programs this is just the loop head, but for concurrent programs there is no really such a thing, since there can be multiple concurrent loops. Therefore I just return any of those states contained in the predicate. This should be one loop head of the involved loops. But I just found out that this is not always true, so I will change that.

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
schuessf: Good idea, I will change that.

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/cegar/AbstractBuchiCegarLoop.java
schuessf: Ah, another old, yet unspotted typo 😄

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
maul-esel: >  If the program terminates, we know that we did not reach this location.

Ah ok, of course.

> If the program reaches this location, it cannot leave it anymore (the only outgoing edge is a self loop). I could also add the stem to the check, it does not make a difference.

Even if one thread reaches such a location, other threads may still move, right?

schuessf: Yes, the other threads could still move, the self-loop may not even occur in the lasso-loop. But we still reached the insufficient-thread-location. This could be a bit imprecise in some (hopefully rare) cases (similar for `TraceAbstraction`), but should be still correct.

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
schuessf: Changed it in this commit: 5a81a7ce46a14f1fbbb2ca9b259ca8214dbf3eac
Now there is a method `private static IcfgEdge getHondaAction(...)` and we use this edge as an identifier. This should output better line numbers (we use just the source of this edge) for concurrent programs and should not make a difference for sequential programs.

danieldietsch (approved):

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
maul-esel: Yes, I was thinking about this case. I'm (reasonably) convinced that the current solution is sound.

(We can have a look in the future if in such cases we want to modify the nonterminating counterexample before we output it, but let's leave it for now.)

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
maul-esel: :+1:

maul-esel (approved):

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
schuessf: Yes, I think modifying the counterexample is a good idea. First we want to remove all self-loops in the insufficient thread from it. We could safely do that, since it does not change the location. If the loop is empty then, we need to increase the number of thread instances, otherwise we found a "real" counterexample and can stop.

We probably also want to remove the edge leading to the insufficient thread. This edge should only occur in the stem. There we need to slightly modifiy the locations, but I guess that should not be an issue.

Removing these edges should also be good, since we don't have a backtranslation for them (I think).

PR #590, opened 2022-08-05
Abstract Independence & Combined Abstract/Concrete POR

maul-esel:
This pull request merges GemCutter changes from the last few months back into `dev`.

## Abstract Independence
The PR  adds abstract independence relations (using variable-based abstraction by @lawki, and looper independence) for GemCutter. These independence relations allow more (or in general, *other*) statements to commute than the `SemanticIndependenceRelation` and `SyntacticIndependenceRelation` implemented so far, while still allowing for sound verification.

## New POR Algorithm
Furthermore, the PR adds a newly developed partial order reduction algorithm (called `SleepMapReduction`) that can take into account multiple independence relations at the same time. This still needs optimization work to be competitive, but the hope is that it will become GemCutter's new default reduction algorithm in the future.

## Overview of the Changes

* The changes in `trunk/examples` add a couple of new benchmarks and reorganize the GemCutter settings files.
* The changes in `Library-Automata` implement the new reduction algorithm, along with a few minor changes in `TotalizeNwa` and `ProductNwa`.
* In `Library-ModelCheckerUtils`, @Heizmann and I made some changes to support transferring `TransFormula`s with fresh term variables between scripts.
* In `Library-TraceCheckerUtils` we find the new independence relations and a lot of support infrastructure.
* `UltimateUtil` has some new code relating to lattices (needed for abstract independence refinement)
* `BasicCegarLoop` has a new feature useful for debugging: A set of SMT assertions can be read from a file and added to the proof, before the first iteration. This allows to e.g. check if the proof check actually succeeds on a proof that is expected to suffice for a given program.

***

I will do some more minor cleanup work in the next few days (mostly on the new benchmarks), and hope to merge this by Wednesday next week. Reviews are welcome :)


schuessf (approved):
I am not really familiar with the partial order reduction implementation, but I support the general idea to combine multiple independence relations. At least on the parts I understand I could not spot any issue.

danieldietsch:
This will take some time, but I look into it this evening, perhaps I can at least skim ;)

danieldietsch:
:)

PR #592, opened 2022-08-19
Minor Petri net improvements

maul-esel:
* Classes that implement `hashCode` should also implement `equals`
* More efficient implementation of the method `BranchingProcess::conflictPathCheck` (only called in an `assert` though)
* add util method and fix one-safety check in `Marking`


schuessf (approved):
Looks good to me!
Just two questions out of curiosity:
* Until now no one is actually calling the `equals` methods, right?
* Couldn't we get rid of [TransitionUnifier](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/TransitionUnifier.java) and just call `Transition.equals` instead (which is now implemented)?

maul-esel:
We've discussed the PR internally, as documentation for the future:

* `equals` is called internally, when we e.g. have sets of transitions, conditions or events
* Unlike `Transition::equals`, the `TransitionUnifier` must not take the ID into account; hence it must remain separate

PR #593, opened 2022-08-19
Petri-net refactorings

schuessf:
* Introduce new interface `IPetriNetTransitionProvider` between `IPetriNet` and `IPetriNetSuccessorProvider` for an on-demand construction.
* Move `getSuccessors(PLACE)` and `getPredecessors(PLACE)` from `IPetriNet` to `IPetriNetTransitionProvider`.
* Remove the interface `ITransition` and use its only implementation `Transition` everywhere, since we need its successors and predecessors. This should remove unnecessary casts and make the code cleaner.
* Remove `getSuccessors(Transition)` and `getPredecessors(Transition)` from `IPetriNet` and just call the methods on the `Transition` itself (connected to the previous point).
* Remove `getSuccessorTransitionProviders(HashRelation)` from `IPetriNetSuccessorProvider`, since it is never called except for implementing itself again. `IPetriNetSuccessorProvider` has another method `getSuccessorTransitionProviders(Set<PLACE>, Set<PLACE>)` that probably replaced the other method.
* Replace the last call to a method of `SetOperations` by `DatastructureUtils` and remove this deprecated class.
* Some attempts to be as general as possible with petri-nets (prefer `IPetriNetSuccessorProvider` over  `IPetriNet` and `IPetriNet` over `BoundedPetriNet`)
* Some CC fixes and minor code simplifications (e.g. use streams in some simple cases)
* Additionally this branch also contains the Büchi-Petri-Nets as work-in-progress to simplify the merge process. You can safely ignore all newly created files for now 😉


trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/Difference.java
```diff
@@ -500,25 +492,6 @@ private void syncWithAnySelfloop(final ITransition<LETTER, PLACE> oldTrans) {
...
-		}
-		for (final PLACE oldPlace : mMinuend.getSuccessors(trans)) {
-			final PLACE newPlace = mOldPlace2NewPlace.get(oldPlace);
-			if (newPlace == null) {
-				throw new IllegalArgumentException("no copy for minuend place: " + oldPlace);
-			}
-			succs.add(newPlace);
-		}
-	}
-
```
maul-esel, line 521: What is the reason for removing this method and the map `mOldPlace2NewPlace`?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/DifferencePetriNet.java
```diff
@@ -62,30 +61,22 @@
  * @author Matthias Heizmann (heizmann@informatik.uni-freiburg.de)
  */
 public class DifferencePetriNet<LETTER, PLACE> implements IPetriNetSuccessorProvider<LETTER, PLACE> {
-
-	private static final String EXACTLY_ONE_STATE_OF_SUBTRAHEND = "query for successors must contain exactly one state of subtrahend";
-	private static final String AT_MOST_ONE_STATE_OF_SUBTRAHEND = "query for successors must contain at most one state of subtrahend";
-	private static final String EMPTY_INITIAL_ERROR_MESSAGE = "Subtrahend has no initial states! We could soundly return the minuend as result (implement this if required). However we presume that in most cases, such a subtrahend was passed accidentally";
```
maul-esel, line 68: Why inline all these error messages?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/DifferencePetriNet.java
```diff
@@ -581,56 +400,62 @@ public DifferenceSynchronizationInformation<LETTER, PLACE> computeDifferenceSync
...
+	@Override
+	public Set<Transition<LETTER, PLACE>> getSuccessors(final PLACE place) {
+		return mYetConstructedResult.getSuccessors(place);
+	}
+
+	@Override
+	public Set<Transition<LETTER, PLACE>> getPredecessors(final PLACE place) {
+		return mYetConstructedResult.getPredecessors(place);
+	}
+
```
maul-esel: Will this behave as expected? I.e., could it not be necessary to construct a place's successors (or perhaps predecessors) when this method is called (on-demand)?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/PrefixProduct.java
```diff
@@ -73,7 +72,7 @@
 	private final Map<PLACE, PLACE> mOldPlace2newPlace = new HashMap<>();
```
maul-esel: You removed a similar map in `PrefixProduct` -- would it make sense to do the same here?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/BranchingProcess.java
```diff
@@ -357,6 +323,38 @@ public boolean inCausalRelation(final Condition<LETTER, PLACE> condition, final
...
+		for (final Condition<LETTER, PLACE> e1Pred : e1.getPredecessorConditions()) {
+			for (final Condition<LETTER, PLACE> e2Pred : e2.getPredecessorConditions()) {
+				final Set<Object> e2Ancestors = ancestorNodes(e2);
+				result = result || conflictPathCheck(e1Pred, e2Pred, e2Ancestors);
+			}
+		}
+		return result;
+
+	}
+
```
maul-esel, line 357: I assume this code is moved from somewhere else?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/BranchingProcess.java
```diff
@@ -406,7 +404,7 @@ private boolean conflictPathCheck(final Condition<LETTER, PLACE> c1, final Condi
 	 * @return Set containing all Conditions and Events which are (strict) ancestors of a Condition. The dummyRoot is
 	 *         not considered as an ancestor.
 	 */
-	private Set<Object> ancestorNodes(final Condition<LETTER, PLACE> condition) {
+	public Set<Object> ancestorNodes(final Condition<LETTER, PLACE> condition) {
```
maul-esel: What is the reason to make this public?

maul-esel (commented):
I've looked through the changes to the existing PN code. It mostly looks good to me, I've added some questions / comments.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/BranchingProcess.java
schuessf: No, this is just new code needed for Büchi-petri-nets.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/BranchingProcess.java
schuessf: I am pretty sure, I did not change this, at least not on purpose. I can revert it then.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/Difference.java
schuessf: This map was not really useful. It has only one put-operation with the same key and value. So I just replaced `mOldPlace2NewPlace.get(x)` by `x`.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/PrefixProduct.java
schuessf: I guess you mean `Difference`. If it behaves similarly, I can also remove it here.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/BranchingProcess.java
maul-esel: Ah, ok. I'm not sure though this is the best way to do it, though -- in particular I see the method `conflictPathCheck` here, which is (more efficient now than it was but still) not very efficient, and best only used for debugging / in `assert` statements.

Can you not use the `ICoRelation` provided by the `BranchingProcess` ?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/PrefixProduct.java
maul-esel: Yes, I meant `Difference` -- sorry

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/Difference.java
schuessf: And the reason to remove this method is that these for-loops are not really necessary. You could simply call `Collection::addAll`. Therefore I removed this method and replaced its calls by `addAll`

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/PrefixProduct.java
schuessf: It is indeed similar, so I changed it in b7a2ab5d623d48b6fa5745c1b09d9aa6686c7b08. There were even two identity maps that I could simply remove.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/BranchingProcess.java
schuessf: I am also not sure. @dkuechler added this code and this is still work-in-progress that we also discussed today.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/DifferencePetriNet.java
schuessf: I just inlined the last message. Sonar complained that the line was too long. Since this message is only used at one single position, I decided to inline it. But I could also revert it.
The first two messages are not used anymore. They only occured in the method `getSuccessorTransitionProviders(HashRelation)` (defined in `IPetriNetSuccessorProvider`) that was removed.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/operations/DifferencePetriNet.java
schuessf: I guess you are correct! The transitions are only added in `getOrConstructTransition(Transition<LETTER, PLACE>, PLACE)` that is only called by `getTransitions`. We should probably call it for all transitions with the corresponding place as a successor/predecessor.
Since it is not used (at least for now), we could also just throw an exception there.

schuessf: Okay, I guess it is more complicated, since `getOrConstructTransition` is only a method of an inner class...
So we should rather throw an exception here.

schuessf: See 5ecb9d5f500d43f5f2c2332b55e821961b69ebbd: Now there is an interface `IPetriNetTransitionProvider` (there is still the option to change the name...) between `IPetriNet` and `IPetriNetSuccessorProvider`. Therefore `DifferencePetriNet` does not need to implement these methods anymore.

PR #594, opened 2022-08-21
remove LETTER type parameter from Marking

maul-esel:
A marking is a set of places -- so why is it parameterized in a type of letters? It is sufficient to parameterize those methods that deal with transitions.

Removing this parameter simplifies code: It avoids the need to carry around the additional type parameter LETTER when it is not needed, or using wildcards.


Heizmann (approved):
Makes sense.
I think the historical reason for the LETTER is the following.
In the past we had dedicated Place<LABEL,LETTER> objects that wrapped an LABEL object (which is now the PLACE object itself) and getters for outgoing and incoming transitions.

schuessf (approved):
Looks good to me!
I think the bestis to merge it before #593, since I expect some minor conflicts again.

PR #604, opened 2022-11-02
Remove problematic loop invariant in example program

schillic:
The user experience is suboptimal because the first thing you see is that there are problems with this example, but it is actually correct (only the annotation is "not completely right"). I guess that showcase examples should not contain problematic annotations. The tool can handle this example without the annotation.

Screenshot_20221102_171903


schillic:
Thanks @schuessf!

PR #609, opened 2022-12-08
Add support for varargs

### LINKED ISSUE #390: Support variadic functions
We want to support variadic functions as described in https://en.cppreference.com/w/c/variadic . 

Until recently this was not an issue because most examples that contain variadic functions do not use them. 

Since we retired ``determineNecessaryDeclarations``, we fail on these examples without this support.

schuessf:
This PR adds support for varargs (closes #390). The idea is as follows:
* If we have a call to a varargs-function, we write all the varargs to a pointer and pass it to the function. There is one exception to that: If the varargs are not used anyway, we ommit the creation of this pointer and just dispatch the arguments.
* `va_start(v, a)` is just translated to an assignement of the varargs-pointer-argument to `v`.
* `va_arg(v, type)` is translated using a macro complicated macro `*((typeof(type) *)((v += sizeof(type)) - sizeof(type)))`, since it is not a function (because of the type) and we cannot handle it otherwise. The idea behind this macro is that we cast `v` to a `type*` and increment it afterwards by `sizeof(type)` (since `+=` yields the value after the increment, we need to subtract it...).
* `va_end(v)` basically deallocates `v`, but at its "original" offset 0.


trunk/examples/CToBoogieTranslation/regression/all/varargsDifferentTypes2.c
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+void check(int count,...) {
+   int result = 0;
+   
+   __builtin_va_list valist;
+   __builtin_va_start(valist, count);
+   
+   char* string = __builtin_va_arg(valist, char*);
+   int actuallyChar = __builtin_va_arg(valist, int);
```
maul-esel, line 14: Is this defined behaviour?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
```diff
@@ -396,17 +400,20 @@ public Result handleFunctionDefinition(final IDispatcher main, final MemoryHandl
...
 	}
 
-	private static CFunction updateVarArgsUsage(final IASTFunctionDefinition node, final CFunction oldFunType) {
-		final CFunction funType;
+	private static CFunction updateVarArgsUsage(final ILocation loc, final IASTFunctionDefinition node,
+			final CFunction funType, final String funName) {
 		// update varags usage
-		if (oldFunType.hasVarArgs() && oldFunType.getVarArgsUsage() == VarArgsUsage.UNKNOWN) {
+		if (funType.hasVarArgs() && funType.getVarArgsUsage() == VarArgsUsage.UNKNOWN) {
 			// if the function body creates a va_list object it uses its varargs
```
maul-esel: update the comment

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
```diff
@@ -396,17 +400,20 @@ public Result handleFunctionDefinition(final IDispatcher main, final MemoryHandl
...
-			final ASTNameCollector vaListFinder = new ASTNameCollector("va_list");
+			final ASTNameCollector vaListFinder = new ASTNameCollector("__builtin_va_start");
```
maul-esel: would it make sense to check against both `va_start` and `__builtin_va_start` ?

I guess my general point is, the `__builtin_va_*`-Macros depend on a particular preprocessing -- can we also support files without that preprocessing?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
```diff
@@ -716,6 +688,57 @@ private Result handleFunctionCallGivenNameAndArguments(final IDispatcher main, f
 			translatedParams.add(in.getLrValue().getValue());
 			functionCallExpressionResultBuilder.addAllExceptLrValue(in);
 		}
+		if (calleeProcCType != null && calleeProcCType.hasVarArgs()) {
+			final boolean hasUsedVarargs = calleeProcCType.getVarArgsUsage() == VarArgsUsage.USED;
+			// For varargs we need a special handling:
+			// - If the varargs are not used, we simply dispatch the arguments, without passing them to the function.
+			// - If they are used, we create a pointer for all the remaining arguments and pass them to the function.
+			final AuxVarInfo auxvarinfo = mAuxVarInfoBuilder.constructAuxVarInfo(loc,
+					mTypeHandler.constructPointerType(loc), SFO.AUXVAR.VARARGS_POINTER);
```
maul-esel, line 700: I think you'll probably need to call some allocation function to get the pointer. If you simply use a nondeterministic pointer, wouldn't that

- possibly overwrite the contents of any other pointer?
- violate memsafety?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
```diff
@@ -716,6 +688,57 @@ private Result handleFunctionCallGivenNameAndArguments(final IDispatcher main, f
...
+			final CPrimitive pointerType = mExpressionTranslation.getCTypeOfPointerComponents();
+			BigInteger currentOffset = BigInteger.ZERO;
+			final List<Statement> writes = new ArrayList<>();
+			final Expression originalBase = MemoryHandler.getPointerBaseAddress(auxvarinfo.getExp(), loc);
+			final Expression originalOffset = MemoryHandler.getPointerOffset(auxvarinfo.getExp(), loc);
+			for (final ExpressionResult va : varargs) {
+				final ExpressionResult param;
+				int size;
+				if (va.getCType().getUnderlyingType() instanceof CPrimitive) {
+					// All smaller types (char, short) are promoted to int
```
maul-esel: Is that part of the standard? If so, could you add a reference link here?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
```diff
@@ -716,6 +688,57 @@ private Result handleFunctionCallGivenNameAndArguments(final IDispatcher main, f
...
+					param = mExprResultTransformer.doIntegerPromotion(loc, va);
```
maul-esel: You don't actually check if the type is smaller. How does this behave on e.g. a `long long`?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
```diff
@@ -716,6 +688,57 @@ private Result handleFunctionCallGivenNameAndArguments(final IDispatcher main, f
...
+					size = mTypeSizes.getSize(((CPrimitive) param.getCType()).getType());
+				} else {
+					param = va;
+					size = mTypeSizes.getSizeOfPointer();
```
maul-esel: Are primitives and pointers the only possible types?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
maul-esel: Also, I'm not sure  what the current effect is -- if the call appears in the loop, does using ther auxVar (without e.g. havocing it) not mean that the same pointer is used in every iteration? Is this intended, or can it have harmful effects?

maul-esel: Oh wait, I see you initialize the pointer below. My bad.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
```diff
@@ -880,7 +909,6 @@ private void handleFunctionsInParams(final IDispatcher main, final ILocation loc
 			assert false : "are we missing a type of function declarator??";
 		}
 
-		assert inparamVarListArray.length == paramDecs.length;
```
maul-esel, line 883: Would it make sense to keep this assert, but modify it so it also covers the case with varargs?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
maul-esel: I see `StandardFunctionHandler` already considers this case (so we should do the same here, right?)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
```diff
@@ -396,17 +400,20 @@ public Result handleFunctionDefinition(final IDispatcher main, final MemoryHandl
...
-			funType = oldFunType.updateVarArgsUsage(usesVarArgs);
-		} else {
-			funType = oldFunType;
+			final int numberOfVaStarts = vaListFinder.getNames().length;
+			if (numberOfVaStarts > 1) {
+				// TODO: This requires a different handling of va_start and va_end.
+				// Currently they are simply handled by assignment and deallocation.
+				throw new UnsupportedSyntaxException(loc, funName + " has multiple calls to va_start.");
+			}
+			return funType.updateVarArgsUsage(numberOfVaStarts != 0);
```
maul-esel, line 415: Aren't there also other cases where varargs are used, e.g. by only passing them on to another call?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -886,6 +892,65 @@ private ExpressionResult getNondetStringOrNull(final ILocation loc) {
...
+		final ExpressionResult arg0 =
+				mExprResultTransformer.transformDispatchDecaySwitchRexBoolToInt(main, loc, arguments[0]);
+		builder.addAllExceptLrValue(arg0);
+		// TODO: We do not actually use this parameter, this could be unsound.
+		final ExpressionResult arg1 =
+				mExprResultTransformer.transformDispatchDecaySwitchRexBoolToInt(main, loc, arguments[1]);
+		builder.addAllExceptLrValue(arg1);
+		final Expression dst = arg0.getLrValue().getValue();
+		if (!(dst instanceof IdentifierExpression)) {
+			throw new UnsupportedSyntaxException(loc, "The first argument of " + name + " has to be an identifier.");
```
maul-esel, line 910: Probably ok for now. If we encounter cases where this fails, we could perhaps change the code, so that `dst` is not assigned the pointer, but instead points to a memory location where we write the pointer.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -886,6 +892,65 @@ private ExpressionResult getNondetStringOrNull(final ILocation loc) {
...
 
+	private Result handleVaStart(final IDispatcher main, final IASTFunctionCallExpression node, final ILocation loc,
+			final String name) {
+		final IASTInitializerClause[] arguments = node.getArguments();
+		checkArguments(loc, 2, name, arguments);
+		final ExpressionResultBuilder builder = new ExpressionResultBuilder();
+		final ExpressionResult arg0 =
+				mExprResultTransformer.transformDispatchDecaySwitchRexBoolToInt(main, loc, arguments[0]);
+		builder.addAllExceptLrValue(arg0);
+		// TODO: We do not actually use this parameter, this could be unsound.
```
maul-esel: Is there a particular reason?

maul-esel (commented):
Looks good overall, great that we're finally getting this feature! :tada: 

I left some comments, mostly questions.

trunk/examples/CToBoogieTranslation/regression/all/varargsDifferentTypes2.c
schuessf: You mean "casting" the argument to `int` that has actually type `char` in the call? Yes, since all types smaller than int are promoted to int.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: Yes, we could do something similar here, it might however make the code slightly more complicated. In the `StandardFunctionHandler` this was pretty easy, so I added it there.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: I don't think so, since only with `va_start` you create an actual object from the varargs, you cannot pass them otherwise.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: Yes, it is part of the standard, I will do so.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: The method itself checks this (see. `ExpressionResultTransformer::integerPromotionNeeded`).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: Probably not. However I couldn't find another way to extract the size of a general `CType` as an integer. Is there one?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: We could make it `>=` instead, but if the assertion is violated then, we get an exception in the loop afterwards anyway.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: Yes. `va_start(v, p)` has a strange semantic. All parameters after `p` are considered vargars and are therfore copied to `v`. Usually you just use the last parameter before `...` for `p` and I am neither sure, what happens otherwise, nor how we could handle it, since we would need to refer to the function signature that we don't have here.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: Yes, this could be done. I did however not encounter such an example yet and I am not sure if anyone actually does that 🙃 (it would be really strange code imo)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: I just updated it. However to support varargs for files without preprocessing, we need also macros (in `GccStaticLanguageSettingsProvider`) for `va_list` and `va_arg`. Should I also add them or should we leave that for now?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: Okay, actually I don't need the size as integer, an expression should be sufficient.  Therefore I can use `TypeSizeAndOffsetComputer::constructBytesizeExpression` and support all possible `CType`s.

schuessf: I just did that in 407bfbf410f6773d46022c1fcf7e36ffd9da63bf, just using `MemoryHandler::calculateSizeOf`, since we already have a `MemoryHandler` in this method. This is also a much nicer solution than before.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
maul-esel: Is there a reason against it? If not, I'd suggest so -- it's better than leaving it half-implemented.

That brings up another question I had: You say we need a macro for `va_list`, but I don't see one for `__builtin_va_list`. How is that handled?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
maul-esel: I'm nitpicking, but do you happen to have an actual web link for that? or if not, maybe put `see C standard, section 7.6.11.2`)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
maul-esel: :+1: Great. I expect that in almost all cases, this should result in literals in the Boogie code, right?

maul-esel: Also, do you have a small regression test for that? E.g. by passing a struct?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
```diff
@@ -661,7 +661,12 @@ private Result handleFunctionCallGivenNameAndArguments(final IDispatcher main, f
 			} else if (calleeProcInfo.getCType() != null) {
 				// we already know the parameters: do implicit casts and bool/int conversion
 				if (i >= calleeProcCType.getParameterTypes().length && calleeProcCType.hasVarArgs()) {
+					if (in.getCType() instanceof CPrimitive) {
```
maul-esel: Shouldn't it be the underlying type here?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: > 👍 Great. I expect that in almost all cases, this should result in literals in the Boogie code, right?

For all the previous cases (primitive, pointer) this behaves exactly the same like before. The `ExpressionFactory` ensures that e.g. the addition of two literals is a literal again (and other simplifications).

> Also, do you have a small regression test for that? E.g. by passing a struct?

I will add a test for struct.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
maul-esel: Ok, I'll leave that up to you

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
maul-esel: Ah, ok, that is indeed very strange. So could I theoretically, in a function `void f(int a, int b, int c)` use `va_start(v, a)` to get a pointer pointing to the values of `b` and `c` ?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: Sure, I can just add it. For `__builtin_va_list` there is already a macro in the CDT parser, the same holds for `__builtin_va_arg` (but this is unfortunately not correct).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java
schuessf: I added a better check in 73c48b98b4d586f3fb469fd05d434677c9e0bb9a (not sure why it's not shown as outdated here...)

schuessf: Also I think it would be good, if `doIntegerPromotion` does not throw an exception for non-primitive types, but just returns the operand, but this should not be changed in this PR.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: So I just checked it: The compiler does not allow `va_start` for fixed-argument-functions. But if you have a function `f(int a, int b, ...)` and you call `__builtin_va_start(v, a)`, you still get only the `...`-arguments. The C-standard is also not quite precise about that:
>The parameter parmN is the identifier of the rightmost parameter in the variable parameter list in the function definition (the one just before the , ...).

What is supposed to happen, if parmN is not the rightmost parameter? So I think, it is okay, how we currently handle it.

maul-esel: :+1: Sounds good. It seems like all other cases are unspecified, so hopefully they shouldn't occur.

schuessf: Okay, so I will replace the TODO with a comment instead.

maul-esel (approved):

PR #626, opened 2023-04-17
Support parsing SMT models

maul-esel:
This PR adds support for parsing SMT models as output by Z3. We need this in particular for models returned by Z3's CHC solver, spacer.

There is currently no support for models that define auxiliary functions. Z3 tends to do this for models containing arrays, as well as some other cases. We leave this support to future work ;-)

### DIFF
--- a/trunk/source/SMTSolverBridge/src/de/uni_freiburg/informatik/ultimate/smtsolver/external/ModelDescription.java
+++ b/trunk/source/SMTSolverBridge/src/de/uni_freiburg/informatik/ultimate/smtsolver/external/ModelDescription.java
@@ -45,7 +45,7 @@ import de.uni_freiburg.informatik.ultimate.logic.TermVariable;
  *
  */
 public class ModelDescription implements Model {
-	private final Map<String, FunctionDefinition> mDefinitions;
+	private final Map<FunctionSymbol, FunctionDefinition> mDefinitions;
 
 	public ModelDescription(final Set<FunctionDefinition> definitions) {
 		mDefinitions = definitions.stream().collect(Collectors.toMap(FunctionDefinition::getName, Function.identity()));
@@ -63,9 +63,7 @@ public class ModelDescription implements Model {
 
 	@Override
 	public Set<FunctionSymbol> getDefinedFunctions() {
-		// TODO how to go from function name strings to FunctionSymbols?
-		// return mDefinitions.keySet();
-		throw new UnsupportedOperationException();
+		return mDefinitions.keySet();
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


trunk/source/SMTSolverBridge/src/de/uni_freiburg/informatik/ultimate/smtsolver/external/ModelDescription.java
```diff
@@ -0,0 +1,96 @@
...
+		throw new UnsupportedOperationException();
+	}
+
+	@Override
+	public Set<FunctionSymbol> getDefinedFunctions() {
+		return mDefinitions.values().stream().map(FunctionDefinition::getName).collect(Collectors.toSet());
+	}
+
+	@Override
+	public Term getFunctionDefinition(final String func, final TermVariable[] args) {
```
schuessf, line 72: This is of course related to the super class, but why do the args need to be TermVariables? If we want to calculate invariants per location, we want to insert constants as arguments here.

trunk/source/SMTSolverBridge/src/de/uni_freiburg/informatik/ultimate/smtsolver/external/scriptResult.cup
```diff
@@ -580,3 +584,23 @@ goResponse ::= TRUE {: RESULT = true; :}
...
 			 | errorResponse;
+
+/* get-model response, currently only for Horn solvers */
+gmResponse ::= LPAR funDef*:d {: RESULT = new ModelDescription(Set.of(d)); :} RPAR;
+
+funDef ::= LPAR DEFINEFUN {: localVars.beginScope(); :} symbol:fun
+                      LPAR sortedVar*:vars RPAR sort:resultSort term:body RPAR
+            {: localVars.endScope();
+               try {
+                       // TODO Support definition of auxiliary functions
```
schuessf, line 595: Just a minor comment: Support for auxiliary functions should be also possible (requires some work of course) in a similar way, right?

trunk/source/SMTSolverBridge/src/de/uni_freiburg/informatik/ultimate/smtsolver/external/ModelDescription.java
schuessf: ... But this requires changes in SmtInterpol (where the Model interface is), so it is fine for this PR.

maul-esel: So let's leave it for today.

trunk/source/SMTSolverBridge/src/de/uni_freiburg/informatik/ultimate/smtsolver/external/scriptResult.cup
maul-esel: It's a bit more complicated: We need to teach the `Script` about the auxiliary function symbol. Perhaps we could push, create the terms, and pop? Seems ugly.

maul-esel: Or perhaps one could directly inline the auxiliaries?

Let's see if this poses a problem in practice first.

schuessf (approved):

PR #641, opened 2023-07-26
Rework C Translation settings for assert/reach_error

schuessf:
This PR changes the C translation settings s.t. we can handle asserts and reach_error seperately. Previously there have been two interleaved settings `Translation Mode` (a mostly deprecated settings...) and `Check unreachability of error function in SV-COMP mode`. These are now replaced by the new settings `Check unreachability of reach_error function` and `Check assertions from assert.h`. These settings were also replaced accordingly in all the settings files (therefore the big diff...).


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -710,13 +710,7 @@ public Result visit(final IDispatcher main, final IASTArraySubscriptExpression n
...
 
 	public Result visit(final IDispatcher main, final IASTASMDeclaration node) {
-		if (mSettings.isSvcompMode()) {
-			// workaround for now: ignore inline assembler instructions
-			return new SkipResult();
-		}
-		final String msg = "CHandler: Not yet implemented: \"" + node.getRawSignature() + "\" (Type: "
-				+ node.getClass().getName() + ")";
-		throw new UnsupportedSyntaxException(mLocationFactory.createCLocation(node), msg);
+		return new SkipResult();
```
danieldietsch, line 714: we now always ignore assembler without any warning? 
at least add a log message

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/PreprocessorHandler.java
```diff
@@ -144,12 +141,6 @@ public Result visit(final IDispatcher main, final IASTPreprocessorMacroDefinitio
 
 	@Override
 	public Result visit(final IDispatcher main, final IASTPreprocessorPragmaStatement node) {
-		if (mIgnorePreprocessorPragmas) {
-			return new SkipResult();
-		}
-		final String msg = "PreprocessorHandler: Not yet implemented: " + node.toString();
-		final ILocation loc = mLocationFactory.createCLocation(node);
-		mReporter.unsupportedSyntax(loc, msg);
 		return new SkipResult();
```
danieldietsch, line 145: same here: we should not silently ignore preprocessor pragmas

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/preferences/CACSLPreferenceInitializer.java
```diff
@@ -46,11 +45,14 @@ public class CACSLPreferenceInitializer extends UltimatePreferenceInitializer {
 	private static final String MAINPROC_DESC =
 			"Specify the entry function of the program. " + "Use an empty string for library mode "
 					+ "(i.e., assume all globals are non-deterministic and verify each function in isolation).";
-	public static final String LABEL_MODE = "Translation Mode:";
+	public static final String LABEL_ERROR = "Check unreachability of reach_error function";
+	private static final String DESC_ERROR =
+			"Check if every call to reach_error is unreachable. This is used for the ReachSafety category of SV-COMP";
```
danieldietsch: add a note that this will override any body of a function called reach_error()

danieldietsch (changes requested):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: The same happened before in SV-COMP mode. So we should probably not throw an exception here. But always logging the previous error message seems like a good idea!

danieldietsch: yeah, but in "not svcomp mode" we threw an exception ;) -- this helps when we try and find things that we cannot handle.

maul-esel: The solution would be a boolean setting `Ignore inline assembler`, right?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/PreprocessorHandler.java
```diff
@@ -58,15 +58,12 @@
  */
 public class PreprocessorHandler implements IPreprocessorHandler {
 
-	private final boolean mIgnorePreprocessorPragmas;
 	private final CTranslationResultReporter mReporter;
 	private final LocationFactory mLocationFactory;
 
-	public PreprocessorHandler(final CTranslationResultReporter reporter, final LocationFactory locFac,
```
maul-esel, line 65: So preprocessor directives do indeed have an effect... we crash (as we should)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -2421,12 +2419,12 @@ private Result handleErrorFunction(final IDispatcher main, final IASTFunctionCal
...
 	 */
-	private Statement createReachabilityAssert(final ILocation loc, final String functionName) {
-		final boolean checkSvcompErrorfunction = mSettings.checkSvcompErrorFunction();
+	private Statement createReachabilityAssert(final ILocation loc, final String functionName,
+			final boolean checkProperty, final Spec spec) {
 		final boolean checkMemoryleakInMain = mSettings.checkMemoryLeakInMain()
 				&& mMemoryHandler.getRequiredMemoryModelFeatures().isMemoryModelInfrastructureRequired();
 		final Expression falseLiteral = ExpressionFactory.createBooleanLiteral(loc, false);
-		if (!checkSvcompErrorfunction && !checkMemoryleakInMain) {
+		if (!checkProperty && !checkMemoryleakInMain) {
```
maul-esel, line 2424: Does this mean that, if we don't check asserts, we assume they hold? If so, is that the right behaviour? And if so, shouldn't the same be done in `handleAssert` ?

schuessf: > Does this mean that, if we don't check asserts, we assume they hold? If so, is that the right behaviour?

Seems like this is the case. I did not introduce the support for this, but just adapted the settings to handle it separately. We could also use the three-value-enum (IGNORE, ASSUME, CHECK) here, if we want to be more flexible.

> And if so, shouldn't the same be done in handleAssert ?

Yes, you are right, `handleAssert` should be handled similarly.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: Yes, I think those are the two options: Always succeed with a warning or add a setting to crash. I am fine with both

danieldietsch: We should not add another setting for that. I would propose we generate a result that warns a user that we encountered such an instruction and ignored it (i.e., the result became potentially unsound). Same for preprocessor pragmas.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
danieldietsch: good catch!

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: Okay, `CTranslationResultReporter.warn` does exactly that, but only if the setting `Report unsoundness warnings` is enabled, per default it is however disabled. Should we just change the default to enabled?

danieldietsch: Probably a good idea. Shouldnt be that many, right?

danieldietsch (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
maul-esel: Please update the documentation comment for the method.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: I don't think so. And in all cases, where this is used, it should be helpful to report the unsoundness (with a logger-warning and a result).

bahnwaerter (approved):
Your changes look good to me.

Once we have support for macros, we can remove the explicit support for the `assert(...)` macro from `assert.h`. It's great that you noted that as a todo in the code so that we don't forget it.

PR #643, opened 2023-08-10
Add support for YAML-witnesses

schuessf:
The main goal of this PR is to add support for the new YAML-witness format. This includes the following:
* Parsing witnesses (`YamlWitnessParser`) mostly done by @bahnwaerter. There are several classes for the datastructures of the yaml witnesses entries (LoopInvariant, Metadata, ...). They all implement the interface `IYamlProvider` that translates them back to an EO-YAML-Object (useful for the witness generation).
* Generate witnesses (`YamlCorrectnessWitnessGenerator`): The ICFG is traversed to extract invariants. This works similar to the GraphML format. There is a setting, if you want to produce YAML or GraphML witnesses (or both).
* Validate (`YamlCorrectnessWitnessExtractor`) witnesses: The witnesses are first parsed and then translated to `ExtractedWitnessInvariant` (same as for GraphML). The validation works based on the file ending, so we are able to validate GraphML- and YAML-witnesses now.

Additionally I made some other changes/refactorings:

* The witness instrumentation does not happen anymore in `MainDispatcher`, but there is now an interface `IExtractedWitnessEntry` that allows to transform an `ExpressionResult` (e.g. can add statements,...). This should simplify the instrumentation and allows also more witness entries in the future more easily (ghost variables, procedure contracts,...). Currently there are two implementation of this interface: `ExtractedLocationInvariant` (instruments an assert) and `ExtractedLoopInvariant` (instruments a loop invariant).
* The witness extraction now returns a `HashRelation` instead of `Map`, since a location can correspond to multiple witness entries. This holds for now (and eliminates merging of witness-invariants; still used in the old format though), but especially if we have more entries like ghost variables.
* Add `WitnessRegressionTestSuite` to check the validation of pre-generated YAML/GraphML-witnesses (to check that we do not break the witness validation). For now I added mostly some random SV-COMP witnesses that could be validated on `dev` (so looks like nothing was broken 😉)

There are still some limitations/open issues:

* The YAML-format allows column numbers, we should also use them, otherwise we might produce faulty witnesses. The problem is that `IASTNode` does not contain them, but an offset we could use to extract the columns. I made a first attempt here: 8b45bde21bed1e175683b99fccd1003b0c207ed9.
* The YAML-format distincts `loop_invariant` and `location_invariant`. Is there any mechanism to distinct this when generating witnesses?
* YAML-violation-witnesses are not supported, but there is also no official format yet.


trunk/examples/witness-checking/regression/correctness/for_loop.c
```diff
@@ -0,0 +1,9 @@
+void main() {
+  int x = 0;
+  for (int i=0; i<=3; i++) {
+    x += i;
+  }
+  if (x != 6) {
+    reach_error();
+  }
+}
```
bahnwaerter: Please add a newline at the end of the file.

trunk/examples/witness-checking/regression/correctness/cohendiv-ll.c-witness.yaml
```diff
@@ -0,0 +1,54 @@
...
+  metadata:
+    format_version: 0.1
+    uuid: "58baa2ef-2a2f-4a83-894b-ec573c79e3db"
+    creation_time: "2023-08-02T17:02:11+02:00"
+    producer:
+      name: UNUSED
+      version: UNUSED
+    task:
+      input_files:
+        - "C:\ultimate\trunk\examples\witness-checking\regression\correctness\cohendiv-ll.c"
```
bahnwaerter, line 12: Does the implementation support validation as well as generation for witness files containing Unix file paths?

I know that there is no official specification for the witness file format available. So we have to deal with platform-specific file paths at the moment.

trunk/examples/witness-checking/regression/correctness/cohendiv-ll.c-witness.yaml
```diff
@@ -0,0 +1,54 @@
+-
```
bahnwaerter, line 1: The witness entry list is not generated correctly (syntax is not human-readable). Can you check the `eo-yaml` output options?

trunk/examples/witness-checking/regression/correctness/const.c-witness.yaml
```diff
@@ -0,0 +1,27 @@
+-
```
bahnwaerter, line 1: Since we create only one YAML document in a physical file, we have to add three dashes at the beginning of each witness file to be comaptible with the YAML standard. I'm sure, that `eo-yaml` supports this feature already. Therefore, check the output options from `eo-yaml`!

trunk/examples/witness-checking/regression/correctness/const.c-witness.yaml
```diff
@@ -0,0 +1,27 @@
...
+  location:
+    file_name: "C:\ultimate\trunk\examples\witness-checking\regression\correctness\const.c"
+    file_hash: UNUSED
+    line: 13
+    column: 0
+    function: main
+  loop_invariant:
+    string: s / 4294967296 * 4294967296 == s
+    type: assertion
+    format: C
```
bahnwaerter: Should we generate newlines at the end of witness files? At least for command line users, this would be more convenient when previewing witness files ...

bahnwaerter (commented):
Thanks @schuessf for implementing the support for YAML-based correctness witnesses. Your code changes are huge but look good to me. Two minor remarks, where the second raises some future design questions (to @maul-esel, @Heizmann, @danieldietsch) rather than criticizing your code changes:

- You can check if `eo-yaml` provides support for annotations which we could use to annotate fields of a YAML witness element class to automatically build YAML serializers. This would reduce further maintenance if YAML generation must be extended for new/changed entry types. Is such a feature also available for parsing?
- Open technical questions regarding the YAML-based witness entry format:
  * How should platform-specific elements (e.g., file paths) encoded in a YAML witness files?
  * Should YAML multi-document witness files also supported?

trunk/examples/witness-checking/regression/correctness/cohendiv-ll.c-witness.yaml
schuessf: We just take the filename from the ICFG that origins from the CDT Parser, so I guess that should get the platform dependant path. This is also the same as for GraphML witnesses. And for validation we don't really care about the path 😉

trunk/examples/witness-checking/regression/correctness/cohendiv-ll.c-witness.yaml
schuessf: You mean because the dash is in a new line? According to [here](https://yaml.org/spec/1.2.2/#21-collections) this should be fine.

trunk/examples/witness-checking/regression/correctness/const.c-witness.yaml
schuessf: No, according to [here](https://yaml.org/spec/1.2.2/#913-bare-documents) these dashes can be ommitted. It is only necessary if the file contains directives which we don't have.

schuessf:
> You can check if `eo-yaml` provides support for annotations which we could use to annotate fields of a YAML witness element class to automatically build YAML serializers. This would reduce further maintenance if YAML generation must be extended for new/changed entry types. Is such a feature also available for parsing?

I am not fully aware how to do this. I however just found out that objects can be dumped to YAML at the very least: https://github.com/decorators-squad/eo-yaml/wiki/Java-Bean-To-YAML

> How should platform-specific elements (e.g., file paths) encoded in a YAML witness files?

I guess platform-specific file paths are fine and I don't think anything else in the witnesses are platform-specific. Also it is more an info for the user, we don't use this for validation.

> Should YAML multi-document witness files also supported?

Why should we generate multiple YAML files? We just want to have one witness and if (in the very future) multi-files projects are supported, the format already has the option to provide multiple source files.

trunk/examples/witness-checking/regression/correctness/cohendiv-ll.c-witness.yaml
schuessf: But I agree that it isn't nice. Unfortunately this is an issue with eo-yaml: https://github.com/decorators-squad/eo-yaml/issues/545
In 5eb799a I customized the output (somewhat hacky, since I need to work on strings) to fix this issue in our implementation.

schuessf:
> > You can check if `eo-yaml` provides support for annotations which we could use to annotate fields of a YAML witness element class to automatically build YAML serializers. This would reduce further maintenance if YAML generation must be extended for new/changed entry types. Is such a feature also available for parsing?
> 
> I am not fully aware how to do this. I however just found out that objects can be dumped to YAML at the very least: https://github.com/decorators-squad/eo-yaml/wiki/Java-Bean-To-YAML

While this works, it is not quite suitable for us. It extracts all methods using reflections and adds them as keys to YAML, therefore we also have e.g. `hashCode` and `toString` keys. ~~Therefore I decided to implement this on my own in 5eb799a. This extracts all getters using reflections and builds a map with them (slightly renamed) as keys.~~

trunk/examples/witness-checking/regression/correctness/cohendiv-ll.c-witness.yaml
bahnwaerter: Okay, then this shouldn't be a problem.

trunk/examples/witness-checking/regression/correctness/const.c-witness.yaml
bahnwaerter: Sorry, I missed that in the YAML specification. Then everything should be fine.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/CLocation.java
```diff
@@ -176,4 +177,12 @@ private IASTNode getMergedNode(final CLocation otherCloc) {
 		}
 	}
 
+	@Override
+	public String getFunction() {
+		final IASTFunctionDefinition scope = CdtASTUtils.findScope(mNode);
+		if (scope == null) {
+			return null;
```
maul-esel, line 184: What does this case signify? Should we throw an error instead?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -1480,15 +1477,13 @@ public Result visit(final IDispatcher main, final IASTFunctionDefinition node) {
...
-			}
-		}
 		final String[] name = { node.getName().toString() };
 		stmt.add(new GotoStatement(mLocationFactory.createCLocation(node), name));
-		return new ExpressionResult(stmt, null);
+		final ExpressionResult result = new ExpressionResult(stmt, null);
+		if (mIsPrerun) {
+			return result;
+		}
+		return ((MainDispatcher) main).handleWitnessEntries(node, result);
```
maul-esel: Why do we need this cast? (I.e. shouldn't this method be part of the interface?)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainDispatcher.java
```diff
@@ -424,8 +409,21 @@ public Result dispatch(final IASTNode n) {
...
+	}
+
+	public Result handleWitnessEntries(final IASTNode node, final Result result) {
+		if (mWitnessEntries == null) {
+			return result;
+		}
+		Result rtr = result;
+		final ILocation loc = mLocationFactory.createCLocation(node);
+		for (final IExtractedWitnessEntry entry : mWitnessEntries.getImage(node)) {
+			if (mNodeLabelsOfAddedWitnesses.add(entry.getNodeLabels())) {
```
maul-esel, line 423: nitpick: side-effects in `if` conditions are not so nice; moving it into a separate line (and storing the result in a boolean variable) would be better

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/CACSL2BoogieTranslatorObserver.java
```diff
@@ -60,19 +62,21 @@ public class CACSL2BoogieTranslatorObserver implements IUnmanagedObserver {
 	private final ILogger mLogger;
 	private final IUltimateServiceProvider mServices;
 	private final ACSLObjectContainerObserver mAdditionalAnnotationObserver;
-	private final CorrectnessWitnessExtractor mWitnessExtractor;
+	private final GraphMLCorrectnessWitnessExtractor mWitnessExtractor;
```
maul-esel: rename for consistency?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/CorrectnessWitnessExtractor.java
```diff
@@ -27,610 +27,94 @@
...
-		}
-
-		final Map<IASTNode, ExtractedWitnessInvariant> rtr = new HashMap<>(mapA.size());
-		for (final Entry<IASTNode, ExtractedWitnessInvariant> entryB : mapB.entrySet()) {
-			final ExtractedWitnessInvariant aWitnessInvariant = mapA.get(entryB.getKey());
-			if (aWitnessInvariant == null) {
-				rtr.put(entryB.getKey(), entryB.getValue());
-			} else {
-				rtr.put(entryB.getKey(), mergeAndWarn(aWitnessInvariant, entryB.getValue()));
+	public HashRelation<IASTNode, IExtractedWitnessEntry> getMatchingWitnessEntries() {
```
maul-esel: nitpick: I find "matching" a bit confusing here, as the signature does not tell me what is matched against.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/CorrectnessWitnessExtractor.java
```diff
@@ -27,610 +27,94 @@
...
-import de.uni_freiburg.informatik.ultimate.witnessparser.graph.WitnessNodeAnnotation;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.HashRelation;
 import de.uni_freiburg.informatik.ultimate.witnessparser.preferences.WitnessParserPreferences;
 
 /**
  *
  * @author Daniel Dietsch (dietsch@informatik.uni-freiburg.de)
  *
  */
-public class CorrectnessWitnessExtractor {
```
maul-esel, line 75: Could we use the opportunity to add some documentation of the class?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedLoopInvariant.java
```diff
@@ -0,0 +1,72 @@
...
+		final List<Statement> statements = new ArrayList<>();
+		boolean hasLoop = false;
+		for (final Statement st : expressionResult.getStatements()) {
+			if (st instanceof WhileStatement) {
+				assert !hasLoop;
+				hasLoop = true;
+				final WhileStatement whileOld = (WhileStatement) st;
+				statements.add(new WhileStatement(loc, whileOld.getCondition(), DataStructureUtils
+						.concat(whileOld.getInvariants(), extractLoopInvariants(invariantExprResult, loc)),
+						whileOld.getBody()));
```
maul-esel, line 41: Once a `while` has been instrumented, should we not  check that there is no second `while` and throw an error if there is?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedLoopInvariant.java
```diff
@@ -0,0 +1,72 @@
...
+			return new ExpressionResultBuilder(expressionResult).resetStatements(statements).build();
+		}
+		// We might identify sth. that is not a loop (e.g. goto) as a loop.
+		// We cannot annotate it with a a LoopInvariantSpecification, so we just insert an assert in that case.
+		return new ExpressionResultBuilder(invariantExprResult).addAllIncludingLrValue(expressionResult).build();
+	}
+
+	private static LoopInvariantSpecification[] extractLoopInvariants(final ExpressionResult result,
+			final ILocation loc) {
+		final List<LoopInvariantSpecification> specs = new ArrayList<>();
```
maul-esel, line 56: I don't fully understand what happens here:

- Could we not directly return the needed data structures, rather than retroactively extracting them?
- How hard would it be to support other statements here that are perhaps required for the subsequent `assert`? Would it make sense?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedWitnessInvariant.java
```diff
@@ -1,119 +1,85 @@
...
-	public String toStringWithWitnessNodeLabel() {
-		return getBAACode() + " [L" + getStartline() + "-L" + getEndline() + "]["
-				+ mNodeLabels.stream().collect(Collectors.joining(", ")) + "]" + mInvariant;
+	protected abstract String getLocationDescription();
+
+	protected ExpressionResult instrument(final ILocation loc, final IDispatcher dispatcher) {
+		ACSLNode acslNode = null;
+		try {
+			checkForQuantifiers(mInvariant);
+			acslNode = Parser.parseComment("lstart\n assert " + mInvariant + ";", getStartline(), 1);
```
maul-esel, line 66: Just as a fun thought: If `mInvariant` is `true;\nwhile(true)`, would we then validate the witness even potentially for an incorrect program?

trunk/source/Library-UltimateUtil/src/de/uni_freiburg/informatik/ultimate/util/CoreUtil.java
```diff
@@ -271,9 +271,6 @@ private static File writeFile(final String filename, final String[] content, fin
 
 	private static File writeFile(final String filename, final String content, final boolean append)
 			throws IOException {
-		if (content == null || content.isEmpty()) {
-			return null;
-		}
```
maul-esel, line 276: :question:

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/WitnessParser.java
```diff
@@ -51,7 +53,7 @@
  */
 public class WitnessParser implements ISource {
 
-	private static final String[] FILE_TYPES = new String[] { "graphml" };
+	private static final String[] FILE_TYPES = new String[] { "graphml", "yaml" };
```
maul-esel, line 56: The typical file extension for YAML is `.yml`, isn't it?

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
```diff
@@ -0,0 +1,142 @@
+package de.uni_freiburg.informatik.ultimate.witnessparser;
```
maul-esel, line 30: Add license header and authors (perhaps in other files as well, I didn't check); don't forget Katie

trunk/examples/witness-checking/regression/correctness/apache-get-tag.i-witness.yaml
```diff
@@ -0,0 +1,54 @@
...
+  metadata:
+    format_version: 0.1
+    uuid: "072e87df-48b4-476f-ac99-67b3aa304c45"
+    creation_time: "2023-08-02T17:01:58+02:00"
+    producer:
+      name: UNUSED
+      version: UNUSED
+    task:
+      input_files:
+        - "C:\ultimate\trunk\examples\witness-checking\regression\correctness\apache-get-tag.i"
```
danieldietsch, line 12: do we need absolute paths here? And wouldnt that lead to problems during testing in CI or on other machines?

trunk/examples/witness-checking/regression/correctness/doubleLoop.c-witness.yaml
```diff
@@ -0,0 +1,54 @@
...
+    uuid: "e92e8d2b-0ac6-4d6d-b6bd-6598f9843c13"
+    creation_time: "2023-08-02T17:29:09+02:00"
+    producer:
+      name: ULTIMATE
+      version: 0.2.3
+    task:
+      input_files:
+        - "C:\ultimate\trunk\examples\witness-checking\regression\correctness\doubleLoop.c"
+      input_file_hashes:
+        "C:\ultimate\trunk\examples\witness-checking\regression\correctness\doubleLoop.c": hash
```
danieldietsch, line 14: has the string `hash` a special meaning for us?

trunk/examples/witness-checking/regression/correctness/gsv2008.i-witness.yaml
```diff
@@ -0,0 +1,27 @@
...
+    uuid: "c294a826-5505-4952-9e7d-e00df27ec731"
+    creation_time: "2023-08-02T17:02:16+02:00"
+    producer:
+      name: UNUSED
+      version: UNUSED
+    task:
+      input_files:
+        - "C:\ultimate\trunk\examples\witness-checking\regression\correctness\gsv2008.i"
+      input_file_hashes:
+        "C:\ultimate\trunk\examples\witness-checking\regression\correctness\gsv2008.i": UNUSED
```
danieldietsch, line 14: here it is `UNUSED` instead of `hash` -- any particular reason? Do both strings have the same effect?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/CLocation.java
danieldietsch: I guess null is the global scope, and then this cannot be part of a function, right?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: Yes, this is much nicer, I will change it.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/CLocation.java
schuessf: Exactly, global locations cannot be assigned any function, so I think returning null should be fine here.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/CorrectnessWitnessExtractor.java
schuessf: I will add some documentation here to help.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedLoopInvariant.java
schuessf: I already added an assert to check this, but throwing an exception is a better idea

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedLoopInvariant.java
schuessf: First we create an assert of the loop invariant, then we extract the assertion and make it a loop invariant in Boogie. Alternatively we could also directly instrument an ACSL loop invariant, but I have not looked to that in detail and just kept thr previous functionality.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedWitnessInvariant.java
schuessf: I am not entirely sure, how parsing ACSL comments works, but I think we just crash with invalid syntax in that case.

trunk/source/Library-UltimateUtil/src/de/uni_freiburg/informatik/ultimate/util/CoreUtil.java
schuessf: It is possible that we create empty YAML witnesses (e.g. if there are no loops), so we should still create the file. I checked the usage of the method, it is only used for witnesses and some stat outputs.

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/WitnessParser.java
schuessf: Both are fine, I think. We might need to check the naming for SV-COMP (also for generating).

trunk/examples/witness-checking/regression/correctness/apache-get-tag.i-witness.yaml
schuessf: We don't use any of the metadata for validation. I just used the filename from CDT parser here. For our regression tests wr could use some dummy string, it does not really matter.

trunk/examples/witness-checking/regression/correctness/doubleLoop.c-witness.yaml
schuessf: I think I created this witness manually, so I have probably overwritten the hash setting. But as already said: metadata are simply ignored for validation.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/CorrectnessWitnessExtractor.java
maul-esel: :+1: What about renaming the method to `getWitnessEntries` ?

trunk/source/Library-UltimateUtil/src/de/uni_freiburg/informatik/ultimate/util/CoreUtil.java
maul-esel: :+1: I was a bit surprised there was even a check like this in the first place -- I'd say a condition like this should be up to the caller

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/WitnessParser.java
maul-esel: :+1: It seems both are used. Ideally then, we should support both for validation, and fix one for generating witnesses. Currently, the examples in the  sv-witnesses repo use both extensions.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainDispatcher.java
danieldietsch: I disagree for this particular case. Adding something to a container and checking if it was already present is IMHO nicer if its done like that.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/util/CdtASTUtils.java
```diff
@@ -174,11 +173,11 @@ private static IASTStatement getFirstOrSuccessorStatement(final IASTCompoundStat
 		return children.length == 0 ? findSuccessorStatement(cmpStmt) : getFirstOrSuccessorStatement(children[0]);
 	}
 
-	public static IASTDeclaration findScope(final IASTNode current) {
+	public static IASTFunctionDefinition findScope(final IASTNode current) {
 		IASTNode parent = current.getParent();
 		while (parent != null) {
 			if (parent instanceof IASTFunctionDefinition) {
-				return (IASTDeclaration) parent;
+				return (IASTFunctionDefinition) parent;
```
danieldietsch, line 180: :+1:

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlCorrectnessWitnessExtractor.java
```diff
@@ -0,0 +1,121 @@
...
+	protected HashRelation<IASTNode, IExtractedWitnessEntry> extract() {
+		final HashRelation<IASTNode, IExtractedWitnessEntry> rtr = new HashRelation<>();
+		for (final WitnessEntry entry : mWitness.getEntries()) {
+			int line = -1;
+			if (entry instanceof LocationInvariant && !mCheckOnlyLoopInvariants) {
+				line = ((LocationInvariant) entry).getLocation().getLine();
+			}
+			if (entry instanceof LoopInvariant) {
+				line = ((LoopInvariant) entry).getLocation().getLine();
+			}
```
danieldietsch, line 47: Is it possible that an  entry is a LocationInvariant and a LoopInvariant? 
What happens if the LineMatchingVisitor gets -1 as line?

trunk/source/Library-UltimateModel/src/de/uni_freiburg/informatik/ultimate/core/model/models/ILocation.java
```diff
@@ -67,6 +67,11 @@ public interface ILocation extends IAnnotations {
 	 */
 	int getEndColumn();
 
+	/**
+	 * @return The function that contains this {@code Location} (i.e. its scope), null if unknown.
```
danieldietsch: Oh, I thought null indicates the global scope. What distinguishes unknown from global scope?

trunk/source/Library-UltimateUtil/src/de/uni_freiburg/informatik/ultimate/util/CoreUtil.java
danieldietsch: I agree for content.isEmpty(), but for content == null, this will crash

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/Witness.java
```diff
@@ -0,0 +1,44 @@
...
+
+	@Override
+	public YamlNode toYaml() {
+		final YamlSequenceBuilder builder = Yaml.createMutableYamlSequenceBuilder();
+		mEntries.forEach(x -> builder.add(x.toYaml()));
+		return builder.build();
+	}
+
+	public String toYamlString() {
+		return toYaml().toString()+"\n";
```
danieldietsch: this looks like it is not formatted

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/Witness.java
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	public boolean isCorrectnessWitness() {
+		// TODO: Check this, when we also support violation witnesses
+		return true;
```
danieldietsch, line 72: Is there currently no witness type in the yaml spec?

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/WitnessEntry.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author Manuel Bentele
+ */
+public abstract class WitnessEntry implements IYamlProvider {
+
+	private final String mName;
+	protected final Metadata mMetadata;
+
+	public WitnessEntry(final String name, final Metadata metadata) {
+		mName = name;
```
danieldietsch: if you base hashcode on that, either ensure that mName is never null or handle the case in hashCode()

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/preferences/PreferenceInitializer.java
```diff
@@ -53,12 +53,14 @@ public enum WitnessVerifierType {
 	public static final String LABEL_GRAPH_DATA_PROGRAMHASH = "Graph data programhash";
 	public static final String LABEL_GRAPH_DATA_ARCHITECTURE = "Graph data architecture";
 	public static final String LABEL_GRAPH_DATA_PRODUCER = "Graph data producer";
+	public static final String LABEL_GRAPH_DATA_PRODUCER_VERSION = "Graph data producer version";
```
danieldietsch: We can determine our version without the wrapper script hack, why dont we?

danieldietsch (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainDispatcher.java
maul-esel: For me, storing that result in a nicely named variable also helps me to remember the meaning of the boolean returned by `add()`

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/Witness.java
schuessf: As far as I know, no. My plan is (once there is a yaml spec for violation witnesses) to check if the witness contains a violation sequence. Maybe the spec might change and some witness type is added.

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/preferences/PreferenceInitializer.java
schuessf: How can this be done?

trunk/source/Library-UltimateModel/src/de/uni_freiburg/informatik/ultimate/core/model/models/ILocation.java
schuessf: Good point! Maybe I can just say "null if unknown or not present" (or something similar) in the docs?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlCorrectnessWitnessExtractor.java
schuessf: > Is it possible that an entry is a LocationInvariant and a LoopInvariant?

No

>  What happens if the LineMatchingVisitor gets -1 as line?

It skips the witness entry (so the result does not contain it), because there are (hopefully) no IASTNodes in that line.

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/preferences/PreferenceInitializer.java
danieldietsch: `UltimateCore.getUltimateVersionString()`

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlCorrectnessWitnessExtractor.java
danieldietsch: So maybe you could use if-else and else instead of the two checks?

PR #644, opened 2023-08-14
Add support for C11 and C23 static assertions

bahnwaerter:
The following signatures of C11 and C23 static assertions will be supported by this patch:

  - `_Static_assert(expr)`
  - `_Static_assert(expr, msg)`
  -  `static_assert(expr)`
  -  `static_assert(expr, msg)`

The handling of static assertions corresponds to regular assertions except the specified error `msg` parameter. This parameter must be a string literal describing a violation of the static assertion in more detail.


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -1832,6 +1839,55 @@ private Result handleAssert(final IDispatcher main, final IASTFunctionCallExpres
...
+				final ExpressionResult result = mExprResultTransformer
+						.transformSwitchRexIntToBool((ExpressionResult) main.dispatch(arguments[0]), loc, node);
+				return new ExpressionResultBuilder()
+						.addAllExceptLrValue(result).addStatement(createReachabilityAssert(loc, name,
+								mSettings.checkAssertions(), Spec.ASSERT, result.getLrValue().getValue(), errorMsg))
+						.build();
+			} else {
+				/* WARNING: this case should be never reached since the msg should be always a string literal */
+				/* TODO: Fallback: handle as regular assertion without any msg */
+				throw new IncorrectSyntaxException(loc, "Message parameter of static assert is not a string literal");
```
schuessf, line 1880: Not really a problem for now, but you could handle this quite easily, right? Just dispatch `arguments[1]` and create an `ExpressionResult` with that and `handleAssert(main, node, loc, name)`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -611,10 +613,15 @@ private Map<String, IFunctionModelHandler> getFunctionModels() {
 				(main, node, loc, name) -> handleVerifierNonDet(main, loc, new CPrimitive(CPrimitives.USHORT)));
 
 		/** from assert.h */
+		/** C standard library functions (from assert.h) to define the 'assert' macro */
 		fill(map, "__assert_fail", this::handleAssertFail);
 		fill(map, "__assert_func", this::handleAssertFail);
 		// TODO: This should not occur in the preprocessed file, but we handle it for now
 		fill(map, "assert", this::handleAssert);
+		/** C11 static assertion (C language keyword, deprecated in C23) */
+		fill(map, "_Static_assert", this::handleStaticAssert);
```
schuessf, line 620: Just to make sure: Are these the functions that occur in the preprocessed files?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -2435,16 +2529,25 @@ private Statement createReachabilityAssert(final ILocation loc, final String fun
...
+
+			/* overwrite result message for reachable error functions */
+			posMsgProvider.registerMessageOverride(Spec.ERROR_FUNCTION,
+					() -> String.format("a call to %s is unreachable", functionName));
+			negMsgProvider.registerMessageOverride(Spec.ERROR_FUNCTION,
+					() -> String.format("a call to %s is reachable", functionName));
+
+			/* overwrite result message for assertions with error messages */
+			if (errorMsg != null && !errorMsg.isEmpty()) {
+				negMsgProvider.registerMessageOverride(Spec.ASSERT,
```
schuessf: Do we only want to override the message for asserts (for now this is of course true 😉) or could you just override the message for `spec`?

trunk/source/Library-UltimateCore/src/de/uni_freiburg/informatik/ultimate/core/lib/models/annotation/CheckNegativeMessageProvider.java
```diff
@@ -0,0 +1,93 @@
...
+ * Negative default message provider for {@link Check} specifications that are checked.
+ * 
+ * @author Manuel Bentele
+ */
+public class CheckNegativeMessageProvider extends CheckMessageProvider {
+
+	@Override
+	public String getDefaultMessage(final Spec spec) {
+		switch (spec) {
+		case ARRAY_INDEX:
```
schuessf: I like your refactoring there, this moves these huge switch-cases out of check. However, one issue is that, if you want to add a new Spec, you need to change three files (`Check`, `CheckPositiveMessageProvider`, `CheckNegativeMessageProvider`) and these switch-cases are still there (just somewhere else). So since you started this, is there a nicer solution? Maybe it is better if you can just add two methods to `Spec` for positive/negative message (along with a constructor)? This eliminates these switch-cases and you only need one class for `CheckMessageProvider`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
bahnwaerter: You're right! This would be an easy solution for the proposed TODO.

At the moment, the intended behavior from the C11 and C23 standard is implemented. The proposed TODO is only one possibility to weaken this exceptional case, but is not covered by any C standard. So it is up to you: Should we weaken this exceptional case or not?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
bahnwaerter: No, these functions are C language keywords defined by the C standard. Currently, we cannot handle C standard-specific keywords in the CDT parser, so we treat them somewhat hacky as functions.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: Ah, okay, if there is only a string literal allowed by the C standard, then `IncorrectSyntaxException` is fine here. I just misread your comment here, so you could please update it and state that this case is not allowed (instead of the TODO)?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
bahnwaerter: Each specific `Check` should implement its own message customization. For example, a `ReqCheck` defines other messages for a `Spec` than a regular `Check`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: Okay, but i mean, if I preprocess a file, it still contains `_Static_assert` and `static_assert` afterwards, right? So these are not macros that are replaced (like `assert(...)`). They are of course specific keywords, but the "look like" function, so our handling should be fine.

trunk/source/Library-UltimateCore/src/de/uni_freiburg/informatik/ultimate/core/lib/models/annotation/CheckNegativeMessageProvider.java
bahnwaerter: You're right. If a new `Spec` should be added, three files have to be edited. Your proposed solution would be one possibility to get rid of the switch-case statements.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: Yes, but I mean just here. Wouldn't it be better to to use `negMsgProvider.registerMessageOverride(spec,...)` instead of `negMsgProvider.registerMessageOverride(Spec.ASSERT,...)` here? Not crucial, should just allow us to have messages for sth. else than assertions here.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
bahnwaerter: I will remove the TODO to be compliant with the C standard.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
bahnwaerter: A C file still contains both static assertion keywords after preprocessing. As you mentioned, our handling should be fine here.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
bahnwaerter: No you can't do that, because each `Spec` requires its own result message. Since the function `createReachabilityAssert(...)` defines behavior for various `Spec`s, we have to distinguish the type of `Spec` here.

Note that default messages are automatically used as fallback messages, i.e. if there is no overwrite for a specific `Spec` type is defined, the default message is used.

bahnwaerter: Got it! I changed the implementation to cover your idea to customize the default message for all `Spec` types if an error message is provided.

schuessf (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -2407,18 +2460,56 @@ private Result handleMemCopyOrMove(final IDispatcher main, final IASTFunctionCal
...
+	 * @param checkProperty
+	 *            enables checking of the reachability specification, otherwise an assumption is made.
+	 * @param spec
+	 *            type of the reachability specification that should be checked.
+	 * @param expr
+	 *            expression for checking the reachability specification.
+	 * 
+	 * @see {@link #createReachabilityAssert(ILocation, String, boolean, Spec, Expression, String)}
+	 */
+	private Statement createReachabilityAssert(final ILocation loc, final String functionName,
```
danieldietsch: I think the comment is somewhat misleading. The function creates an assume statement or an assert statement annotated with a check annotation. It is somewhat strange that it is called "createReachabilityAssert". Also, a "reachability specification" is a) not a really well defined term, and b) just not necessary here. 
Then, the parameter function name is only relevant for certain types of Specs, so the comment is probably also misleading.

trunk/source/Library-UltimateModel/src/de/uni_freiburg/informatik/ultimate/core/model/models/annotation/IMessageProvider.java
```diff
@@ -0,0 +1,111 @@
...
+
+	/**
+	 * Returns a positive default message for a given {@link ISpec.Type}.
+	 * 
+	 * @param spec
+	 *            a specification that is checked.
+	 * 
+	 * @return positive default message for the given {@code spec}.
+	 */
+	public default String getDefaultPositiveMessage(final ISpec.Type spec) {
```
danieldietsch: public is not necessary in a public interface

trunk/source/Library-UltimateCore/src/de/uni_freiburg/informatik/ultimate/core/lib/models/annotation/MessageProvider.java
```diff
@@ -0,0 +1,131 @@
...
+		if (!spec.isInGroups(mSpecGroups)) {
+			throw new IllegalArgumentException(String.format(
+					"Messages for specification type '%s' are not handled by this message provider for '%s'!", spec,
+					mSpecGroups));
+		}
+	}
+
+	@Override
+	public void registerPositiveMessageOverride(final ISpec.Type spec, final Supplier<String> msgProviderFunc) {
+
```
danieldietsch: why the blank line?

trunk/source/Library-UltimateCore/src/de/uni_freiburg/informatik/ultimate/core/lib/models/annotation/MessageProvider.java
```diff
@@ -0,0 +1,131 @@
...
+	public void registerNegativeMessageOverride(final ISpec.Type spec, final Supplier<String> msgProviderFunc) {
+
+		mNegMsgProviderOverrideFuncs.put(spec, msgProviderFunc);
+	}
+
+	@Override
+	public String getPositiveMessage(final ISpec.Type spec) {
+		final Supplier<String> posMsgProviderFunc = mPosMsgProviderOverrideFuncs.get(spec);
+
+		checkSpecificationGroups(spec);
```
danieldietsch, line 106: perhaps call this function first?

trunk/source/Library-UltimateModel/src/de/uni_freiburg/informatik/ultimate/core/model/models/annotation/ISpec.java
```diff
@@ -0,0 +1,332 @@
...
+		/**
+		 * Not further specified or unknown.
+		 */
+		UNKNOWN(Group.GENERIC,
+				"unknown kind of specification holds",
+				"unknown kind of specification may be violated"),
+		/**
+		 * An LTL property.
+		 */
+		LTL(Group.GENERIC),
```
danieldietsch: why does LTL not have default messages?

trunk/source/Library-UltimateModel/src/de/uni_freiburg/informatik/ultimate/core/model/models/annotation/ISpec.java
```diff
@@ -0,0 +1,332 @@
...
+package de.uni_freiburg.informatik.ultimate.core.model.models.annotation;
+
+import java.util.Set;
+
+/**
+ * Defines specification types for arbitrary checks and analyzes.
+ * 
+ * @author Manuel Bentele
+ */
+public interface ISpec {
```
danieldietsch: I am not quite sure that the new architecture for specification types is an improvement above the old one. Could you explain to me why it is better like this?

danieldietsch (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
bahnwaerter: You're right. The function was already named this way before my changes. It's probably better if we refactor it all ...

trunk/source/Library-UltimateModel/src/de/uni_freiburg/informatik/ultimate/core/model/models/annotation/IMessageProvider.java
bahnwaerter: But it's not forbidden either ;-) Jokes aside, I will remove it.

trunk/source/Library-UltimateCore/src/de/uni_freiburg/informatik/ultimate/core/lib/models/annotation/MessageProvider.java
bahnwaerter: Despite the autoformatter, the blank line remained. I will remove it.

trunk/source/Library-UltimateCore/src/de/uni_freiburg/informatik/ultimate/core/lib/models/annotation/MessageProvider.java
bahnwaerter: Thanks for the hint. You're proposed calling order makes more sense.

trunk/source/Library-UltimateModel/src/de/uni_freiburg/informatik/ultimate/core/model/models/annotation/ISpec.java
bahnwaerter: Previously, there was no default message for LTL specifications. I have preserved this behavior. Should we add one?

trunk/source/Library-UltimateModel/src/de/uni_freiburg/informatik/ultimate/core/model/models/annotation/ISpec.java
bahnwaerter: I quickly moved the `Spec` types to an interface because they have no place in the implementation. In my opinion these are part of the generic software model as they can be implemented anywhere in Ultimate (see program checks in `Check` as well as requirement checks in `ReqCheck`).

However, we don't necessarily need the interface. On the one hand, it would be currently sufficient if only the `ISpec.Type` enum is made public available in the model. On the other hand, additional functions could be added to the interface if the `Check` class is cleaned up. Currently, the connections between `ReqCheck`, `Check` and `Spec`s looked a bit hacky to me and were not properly encapsulated.

@schuessf also wanted the addition of new `ISpec.Type`s to be easier and the default messages to be defined centrally. Instead of having to edit three files, you now have to edit only one file.

Do you have another idea for the general software structure regarding `Check`s and `Spec` types?

PR #649, opened 2023-10-04
Add ACSL support for ghost variables

schuessf:
This PR adds support for (some) ACSL-Ghost-Statements, specifically:

* `//@ ghost T x;` (local or global context)
* `//@ ghost T x = expr;` (local or global context)
* `//@ ghost x = expr;` (local context only)

This new language feature can then also help to validate witnesses with ghost variables (see [here](https://github.com/ultimate-pa/ultimate/compare/wip/fs/acsl-ghosts...wip/fs/witness-ghost); a PR of that will be created this one is merged 🙂).

In our translation we create and declare a new Boogie-Variable `#ghost~x` and handle the assignment accordingly. Additionally we make sure that only ghost-variables are assigned in ghost-statements (so it should not have any side-effects on program-variables).

However, we need to handle the declaration of global variables (parsed as `GlobalGhostDeclaration`) and local variables (parsed as `GhostDeclaration`) slightly different, because the first is handled by `CHandler` and the latter by `ACSLHandler`. This leads to some duplicate code, but the handling is also slightly different.


trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/acsl.ast
```diff
@@ -21,13 +21,22 @@ CodeForBehavior ::=
...
+	    identifier : String
+	    expr : Expression
+	| { GlobalGhostDeclaration }
+ 		type : ACSLType
+	    identifier : String
+	    expr : Expression
+	| { GhostUpdate }
+	    identifier : String
+	    expr : Expression
+ 	;
```
bahnwaerter: Please use same indentation as existing lines in AST model file:

```suggestion
 	{ Assertion }
		formula : Expression
 	| { CodeInvariant }
		formula : Expression
 	| { GhostDeclaration }
		type : ACSLType
		identifier : String
		expr : Expression
	| { GlobalGhostDeclaration }
		type : ACSLType
		identifier : String
		expr : Expression
	| { GhostUpdate }
		identifier : String
		expr : Expression
 	;
```

bahnwaerter (changes requested):
Thanks @schuessf for your great work on ghost variable support. Your code changes looks fine to me. Feel free to merge your changes after considering my little comment about code formatting/indentation.

bahnwaerter (approved):

bahnwaerter (approved):

trunk/examples/programs/regression/c/ACSL-ghostInitGlobal.c
```diff
@@ -0,0 +1,11 @@
+//#Safe
+// Author: schuessf@informatik.uni-freiburg.de
+// Date: 2023-09-18
+
+//@ ghost int i = 0;
+
+int main() {
+  int x = 0;
+  //@ assert x == i;
```
maul-esel, line 9: Wie ist denn die Semantik für die Nutzung von Ghost-Variablen in Asserts? Beeinflusst man damit nicht das Programm? Sowohl in dem Sinn, dass ein failure auftreten kann; als auch in dem Sinn, dass wir ggf eine `assume x!=i`-Kante einfügen.

trunk/examples/programs/regression/c/ACSL-ghostUpdateGlobal.c
```diff
@@ -0,0 +1,12 @@
+//#Safe
+// Author: schuessf@informatik.uni-freiburg.de
+// Date: 2023-09-18
+
+//@ ghost int i;
+
+int main() {
+  int x = 0;
+  //@ ghost i = x;
```
maul-esel, line 9: In the PR comment you write

> `//@ ghost x = expr;` (local only)

But here it seems you use it with a global ghost variable?

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/acsl/parser/GlobalLocalParser.cup
```diff
@@ -1141,6 +1142,18 @@ code_annotation::=
   {: 
   	 RESULT = setNodeLocationFromSymbols(new CodeAnnotStmt(setNodeLocationFromSymbols(new CodeInvariant(expr), i$, expr$)), i$, s$);
 	 parser.setAcslNode(RESULT); :}
+| GHOST:g IDENTIFIER:i EQUAL lexpr:expr SEMICOLON:s 
+  {: 
+  	 RESULT = setNodeLocationFromSymbols(new CodeAnnotStmt(setNodeLocationFromSymbols(new GhostUpdate(i, expr), i$, expr$)), g$, s$);
+	 parser.setAcslNode(RESULT); :}
+| GHOST:g type_spec:t IDENTIFIER:i SEMICOLON:s
```
maul-esel, line 1149: I haven't fully read the grammar, so this may be a stupid question: Is there a specific reason why this case and the next one, which look like variable declarations to me (since they specify the type) are handled here and not with the declarations (`decl`) below?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -212,32 +222,80 @@ public Result visit(final IDispatcher main, final ACSLNode node) {
...
 			final ExpressionResultBuilder resultBuilder = new ExpressionResultBuilder();
-			resultBuilder.addAllExceptLrValue(formula);
-
-			final AssertStatement assertStmt = new AssertStatement(loc, formula.getLrValue().getValue());
-			// TODO: Handle havoc statements
-			for (final Overapprox overapprItem : resultBuilder.getOverappr()) {
-				overapprItem.annotate(assertStmt);
+			final CodeStatement codeStmt = ((CodeAnnotStmt) node).getCodeStmt();
+			if (codeStmt instanceof Assertion) {
+				ExpressionResult formula =
```
maul-esel: Suggestion: Move the handling of the three cases (`Assertion`, `GhostUpdate` and `GhostDeclaration`) into separate methods

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -212,32 +222,80 @@ public Result visit(final IDispatcher main, final ACSLNode node) {
...
+						(ExpressionResult) main.dispatch(update.getExpr(), main.getAcslHook());
+				resultBuilder.addAllExceptLrValue(exprResult);
+				final SymbolTableValue stv = mSymboltable.findCSymbol(main.getAcslHook(), update.getIdentifier());
+				if (stv == null) {
+					throw new IncorrectSyntaxException(loc,
+							"Undeclared variable in ACSL expression: " + update.getIdentifier());
+				}
+				if (!stv.getBoogieName().startsWith(SFO.GHOST)) {
+					throw new IncorrectSyntaxException(loc,
+							"C variable " + update.getIdentifier() + " cannot be assigned in ghost statement.");
```
maul-esel: Does this mean that ghost variables cannot shadow program variables with the same name?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -212,32 +222,80 @@ public Result visit(final IDispatcher main, final ACSLNode node) {
...
+				} else {
+					check = new Check(Spec.ASSERT);
+				}
+				check.annotate(assertStmt);
+			}
+			if (codeStmt instanceof GhostUpdate) {
+				final GhostUpdate update = (GhostUpdate) codeStmt;
+				final ExpressionResult exprResult =
+						(ExpressionResult) main.dispatch(update.getExpr(), main.getAcslHook());
+				resultBuilder.addAllExceptLrValue(exprResult);
```
maul-esel: Naive question: Is the call to `transformSwitchRexIntToBool` (see the case for `Assertion`) not needed in this case?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -212,32 +222,80 @@ public Result visit(final IDispatcher main, final ACSLNode node) {
...
+					check = new Check(Spec.WITNESS_INVARIANT);
+				} else {
+					check = new Check(Spec.ASSERT);
+				}
+				check.annotate(assertStmt);
+			}
+			if (codeStmt instanceof GhostUpdate) {
+				final GhostUpdate update = (GhostUpdate) codeStmt;
+				final ExpressionResult exprResult =
+						(ExpressionResult) main.dispatch(update.getExpr(), main.getAcslHook());
```
maul-esel: What is the situation regarding (possibly implicit) assumptions or side effects resulting from the evaluation of the expression? For example:

- Wouldn't `// ghost x = 1/y`, where `y` is non-ghost, introduce some assumption that `y != 0`? Similarly for `*y` etc.
- What would happen for `// ghost x = y++` ?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -212,32 +222,80 @@ public Result visit(final IDispatcher main, final ACSLNode node) {
...
+				final String boogieName = SFO.GHOST + decl.getIdentifier();
+				final CPrimitive cType = AcslTypeUtils.translateAcslTypeToCType(decl.getType());
+				final ASTType astType = mTypeHandler.cType2AstType(loc, cType);
+				final Declaration boogieDecl = new VariableDeclaration(loc, new Attribute[0],
+						new VarList[] { new VarList(loc, new String[] { boogieName }, astType) });
+				final CDeclaration cDecl = new CDeclaration(cType, decl.getIdentifier());
+				final IASTFunctionDefinition scope = CdtASTUtils.findScope(main.getAcslHook());
+				DeclarationInformation declInfo;
+				if (scope == null) {
+					declInfo = DeclarationInformation.DECLARATIONINFO_GLOBAL;
```
maul-esel: Can this happen even if a declaration is not parsed as `GlobalGhostDeclaration`?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
maul-esel: Follow-up question: If yes, is this checked at the ghost declarations as well?

trunk/examples/programs/regression/c/ACSL-ghostUpdateGlobal.c
schuessf: I just meant that the ghost-assignment itself can only occur in a local context, but can still assign global variables.

trunk/examples/programs/regression/c/ACSL-ghostInitGlobal.c
schuessf: It does of course influence the program, but only if you consider the ACSL-code. The `//@ assert` itself already influences the program, we could possibly add a settings to consider or ignore ACSL.

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/acsl/parser/GlobalLocalParser.cup
schuessf: `decl` refers to global declarations, but we can also have the declaration of local ghost variables, so they should be handled here.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: On the right-hand side of a ghost assignment only a ACSL expression (and not a C expression) is allowed, so `y++` would already fail during parsing. And since ACSL-expressions should have no side-effects, the assignment itself should have no side-effects on program-variables (only the one assigned ghost variable). For `1/y` we do not any other constraints during our translation, so it does not have any side-effects either (which should be fine, since `1/0` can have any value in Boogie, if I remember correctly).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: For the declaration there should not be any shadowing. If we declare x as a program variable and a ghost variable, the first one is just translated to `x` (with some prefix), and the latter one to `#ghost~x`.
However, there could we shadowing if we have `x` as a program and ghost variable, and now `x` occurs in another ACSL statement (e.g. assert or ghost assignment on the right hand side), because wha `x` do we choose then? I just checked our implementation, we choose the program variable then. But I am not sure either, what the expected semantics are in ACSL for that case.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: No, this should not happen. However, it might happen if we instrument witnesses as ghost-variables, because everything there is handled by `ACSLHandler`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: In Boogie `assert` takes a boolean argument, but this is not true for assignments.

bahnwaerter (approved):

trunk/examples/programs/regression/c/ACSL-ghostInitGlobal.c
maul-esel: We should have a look at the ACSL spec for this. Of course `//@ assert` influences the program, but in my intuition (from a witness perspective), ghost code should not. But it may well be that in this specific sense, ACSL allows such influence.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
maul-esel: We should also check the ACSL spec here. Intuitively, throwing an error at the declaration would be my preference -- but the precise handling might be hard in some cases, e.g. for global ACSL ghost variables and local C variables of the same name.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
maul-esel: Does `transformSwitchRexIntToBool` not perform other changes as well ( I may misremember) ? And what about assignments to boolean ghost variables?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: I checked it (see 2.12 Ghost variables and statements), but I could not find anything related. I might however add a check for the declaration (check if the return of `FlatSymbolTable::findCSymbol` is not null) and throw an exception for this case as well.

trunk/examples/programs/regression/c/ACSL-ghostInitGlobal.c
schuessf: But I mean here it is just the assert that influences the program and not the ghost statement, right? You are right, the ghost code should not influence any program variable or the control flow, but I cannot see, how this is possible.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: No, it does not have any other changes, I think for integer variables this conversion would even be wrong. But I need to check it for boolean variables.

bahnwaerter (approved):

maul-esel (approved):

PR #653, opened 2023-10-25
Support witnesses with ghost variables

schuessf:
This PR adds support for the validation of witnesses that contain ghost variables. Namely the entry-types (that are not official yet) in YAML-witnesses `ghost_variable` and `ghost_update` can now be handled in the validation. The validation of those entry-types works in a similar fashion to the existing entry-type invariant: We translate them to ACSL, s.t. the entries instrumented are in our Boogie-Code. This instrumentation uses ghost code in ACSL that is now supported (see #649).


danieldietsch (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedCorrectnessWitness.java
```diff
@@ -0,0 +1,30 @@
+package de.uni_freiburg.informatik.ultimate.plugins.generator.cacsl2boogietranslator.witness;
```
bahnwaerter: Add license header to source file!

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedGhostUpdate.java
```diff
@@ -0,0 +1,89 @@
+package de.uni_freiburg.informatik.ultimate.plugins.generator.cacsl2boogietranslator.witness;
```
bahnwaerter, line 28: Add license header to source file!

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedGhostVariable.java
```diff
@@ -0,0 +1,56 @@
+package de.uni_freiburg.informatik.ultimate.plugins.generator.cacsl2boogietranslator.witness;
```
bahnwaerter, line 28: Add license header to source file!

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/IExtractedWitnessDeclaration.java
```diff
@@ -0,0 +1,11 @@
+package de.uni_freiburg.informatik.ultimate.plugins.generator.cacsl2boogietranslator.witness;
```
bahnwaerter, line 28: Add license header to source file!

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/GhostUpdate.java
```diff
@@ -0,0 +1,38 @@
+package de.uni_freiburg.informatik.ultimate.witnessparser.yaml;
```
bahnwaerter, line 28: Add license header to source file!

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/GhostVariable.java
```diff
@@ -0,0 +1,44 @@
+package de.uni_freiburg.informatik.ultimate.witnessparser.yaml;
```
bahnwaerter, line 28: Add license header to source file!

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/ILocationProvider.java
```diff
@@ -0,0 +1,5 @@
+package de.uni_freiburg.informatik.ultimate.witnessparser.yaml;
```
bahnwaerter: Add license header to source file!

bahnwaerter (changes requested):
Thanks @schuessf for your great work. Your changes look good to me. However, make sure that new files always contain a license header.

bahnwaerter (approved):

schuessf:
Needs to be revised (after SV-COMP), because the format for YAML witnesses has changed and we also use a different parser.

trunk/examples/witness-checking/regression/AutomizerC.xml
```diff
@@ -3,6 +3,7 @@
 	<toolchain>
 		
 	<plugin id="de.uni_freiburg.informatik.ultimate.plugins.generator.cacsl2boogietranslator"/>
+	<plugin id="de.uni_freiburg.informatik.ultimate.boogie.procedureinliner"/>
```
maul-esel: Please undo this change (such a toolchain already exists).

trunk/examples/witness-checking/regression/correctness/case_distinction.c-witness.yml
```diff
@@ -0,0 +1,59 @@
+- entry_type: invariant_set
```
maul-esel: is it intentional that we have this file with 2 different file extensions?

trunk/examples/witness-checking/regression/correctness/mutex.c-witness.yml
```diff
@@ -0,0 +1,86 @@
+- entry_type: invariant_set
```
maul-esel: This file is also duplicated, and `nondet_inc.c-witness.{yaml,yml}` below as well.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -288,8 +287,12 @@ private Result handleGhostDeclaration(final IDispatcher main, final ILocation lo
 		final String boogieName = SFO.GHOST + decl.getIdentifier();
 		final CType cType = AcslTypeUtils.translateAcslTypeToCType(decl.getType());
 		final ASTType astType = mTypeHandler.cType2AstType(loc, cType);
-		final Declaration boogieDecl = new VariableDeclaration(loc, new Attribute[0],
+		final VariableDeclaration boogieDecl = new VariableDeclaration(loc, new Attribute[0],
 				new VarList[] { new VarList(loc, new String[] { boogieName }, astType) });
+		if (mWitnessInvariantMode) {
+			// TODO: Workaround: In witness mode we need to declare the declaration manually afterwards
```
maul-esel: I don't understand this TODO. What is the "afterwards" referring to?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainDispatcher.java
```diff
@@ -413,14 +414,20 @@ public Result transformWithWitness(final IASTNode node, final Result result) {
...
 		if (!(result instanceof ExpressionResult)) {
-			if (!matchedWitnessInvariants.isEmpty()) {
-				mLogger.warn("Unable to annotate " + node.getRawSignature() + " with a witness entry");
-			}
+			mLogger.warn("Unable to annotate " + node.getRawSignature() + " with a witness entry");
 			return result;
 		}
 		ExpressionResult rtr = (ExpressionResult) result;
 		final ILocation loc = mLocationFactory.createCLocation(node);
+		// Ensure that invariants are evaluated before the ghost variables are updated
```
maul-esel: Why does the code here look like it's doing things in the opposite order?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlCorrectnessWitnessExtractor.java
```diff
@@ -206,6 +222,8 @@ protected int genericVisit(final IASTNode node) {
 	private static final class YamlExtractedCorrectnessWitness implements IExtractedCorrectnessWitness {
 		private final HashRelation<IASTNode, ExtractedWitnessInvariant> mInvariants = new HashRelation<>();
 		private final HashRelation<IASTNode, ExtractedFunctionContract> mFunctionContracts = new HashRelation<>();
+		private final HashRelation<IASTNode, ExtractedGhostUpdate> mGhostUpdates = new HashRelation<>();
```
maul-esel: This seems like it does not preserve the order of ghost updates at a given location. But that could be important, right? Maybe a `Map<IASTNode, List<ExtractedGhostUpdate>>` would be better here.

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
```diff
@@ -92,6 +108,19 @@ private static Stream<WitnessEntry> parseWitnessEntry(final Map<String, Object>
 		}
 	}
 
+	private static GhostVariable parseGhostVariable(final Map<String, Object> entry) {
+		return new GhostVariable((String) entry.get("name"), (String) entry.get("initial"), (String) entry.get("scope"),
+				(String) entry.get("type"));
```
maul-esel: This seems similar but slightly different from the `case GhostVariable.NAME` code above. What is the reason to have both? Does it have to do with different versions of the format?

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
```diff
@@ -92,6 +108,19 @@ private static Stream<WitnessEntry> parseWitnessEntry(final Map<String, Object>
...
+	}
+
+	@SuppressWarnings("unchecked")
+	private static GhostUpdate parseGhostUpdate(final Map<String, Object> ghostUpdate) {
+		final Location location = parseLocation((Map<String, Object>) ghostUpdate.get("location"));
+		final Map<String, String> variables2Values = ((List<Map<String, String>>) ghostUpdate.get("updates")).stream()
+				.collect(Collectors.toMap(x -> x.get("ghost_variable"), x -> x.get("expression")));
+		return new GhostUpdate(variables2Values, location);
```
maul-esel: (similar as for ghost variables, see above)

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/GhostUpdate.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package de.uni_freiburg.informatik.ultimate.witnessparser.yaml;
+
+import java.util.Map;
+
+/**
+ *
+ * @author Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ *
```
maul-esel, line 35: Can you add a short description? (also for `GhostVariable`)

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/yaml/YamlWitnessWriterV2.java
```diff
@@ -47,24 +50,67 @@
  */
 public class YamlWitnessWriterV2 extends YamlWitnessWriter {
 	private final MetadataProvider mMetadataProvider;
-	private final boolean mWriteFunctionContracts;
+	private final boolean mWriteFunctionContractsAndGhostVariables;
```
maul-esel: shouldn't this be two separate settings?

maul-esel: Or perhaps rather, does it even make sense to just filter out ghost variables here? The invariants may still refer to them, making the generated witness invalid. If we want to sometimes avoid ghost variables, we may have to do a more complex transformation earlier on (e.g. by quantifying the variables in invariants).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedGhostUpdate.java
```diff
@@ -0,0 +1,221 @@
...
+		}
+		if (!isAnnotated) {
+			throw new UnsupportedOperationException("No statement found to annotate with the expected ghost update");
+		}
+		return result;
+	}
+
+	@Override
+	public ExpressionResult transform(final ILocation loc, final IDispatcher dispatcher,
+			final ExpressionResult expressionResult) {
```
maul-esel, line 161: This method defines the points where we allow ghost updates. We may want to extend or completely change these points in the future (e.g., allowing updates at sequence points). Should we do anything to ease such changes later (maybe add a setting, or at least a boolean flag, ...) ?

maul-esel (changes requested):
Thanks for adding this support. I've looked through the changes and marked some points that could be improved or clarified.

trunk/examples/witness-checking/regression/AutomizerC.xml
schuessf: This change is necessary for the regression tests, since we need to use inlining for concurrent programs.

trunk/examples/witness-checking/regression/correctness/case_distinction.c-witness.yml
schuessf: Yes, the file ending with `.yaml`  contains a witness in version 0.1 (the ghost-witness format as originally planned on VEWIT), while the file ending with `.yml` contains a witness in the current format 2.1. Since we support both formats, we should also have tests for both.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainDispatcher.java
schuessf: `rtr = entry.transform(loc, this, rtr)` basically adds `entry` before `rtr`, therefore this order ensures that we evaluate the invariants first.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedGhostUpdate.java
schuessf: You are right, the behavior of ghost updates might can. But what should happen, if the setting/flag is set to true/false?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlCorrectnessWitnessExtractor.java
schuessf: Wasn't this our idea that the ghost updates don't have a particular order?

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
schuessf: Yes, exactly, this function is used for the format version 2.1 (therefore this function is called in the `case "ghost_instrumentation"`), while the one above is used for version 0.1. The slight difference is `variable` vs. `name` as field name for the name of the ghost variable.

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/yaml/YamlWitnessWriterV2.java
schuessf: The intention here was to generate ghost variables and function contracts only for version 2.1, since the format in 2.0 does not allow ghost variables. But you are right, we can not simply ommit ghost variables in the witness, but keep them in invariants. So I think, we should just crash, if we want to print a witness with ghost in version 2.0.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: Without this, we "forget" the actual declaration of the ghost variable in Boogie (somehow caused by our handling of ACSL...). But I think, there is a nicer way to actually add the declaration to the `ExpressionResult`, I will just check.

trunk/examples/witness-checking/regression/AutomizerC.xml
maul-esel: :see_no_evil: my mistake, sorry; I didn't look at the full path

trunk/examples/witness-checking/regression/correctness/case_distinction.c-witness.yml
maul-esel: - Is it worth the effort to keep supporting both?
- If we do, I'm strongly against distinguishing them by omitting one letter in the file extension. It makes it really hard for anyone to understand what is going.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlCorrectnessWitnessExtractor.java
maul-esel: But one ghost update may use the value of another ghost  variable assigned at the same location. So what would that mean exactly in terms of the semantics of witnesses? 

- One possibility: _The witness must be valid for all possible orders of ghost updates._

  This is still a bit unclear (e.g. may the order change if the ghost updates are executed multiple times?), and seems unnecessarily complicated.

- Another possibility: _For validity of witnesses, we require that all ghost updates at a certain location commute._

  Or more syntactically: _We require that no ghost variable is assigned twice at the same location, and no ghost variable that is assigned at this location is used in another ghost assignment at the same location._

  Then we should actually perform a check for such a condition.

I would prefer fixing an order of ghost updates to both possibilities above, as it seems the most intuitive and least restrictive solution. The natural order to choose would be the order of the entries in the witness.

---

Independent of this question (regarding witness semantics and validation) is the fact that when we generate a witness, for all algorithms we are _currently_ considering, the order happens not to matter.

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
maul-esel: See above: It seems like a lot of effort just to keep supporting a draft format version proposed in a workshop talk. If we update our artifact repo for the workshop to point to the ULTIMATE version where this format is supported, I think we can safely abandon the old format going forward.

trunk/examples/witness-checking/regression/correctness/case_distinction.c-witness.yml
schuessf: * We already have the support and since our internal datastructures for witnesses are now independant from the format version, supporting another format version requires only a couple of lines in `YamlWitnessParser` and `YamlWitnessPrinter`. Also: the witnesses from Goblint are currently still in the old format 🙂.
* I agree, the file endings are somehow just a "coincide". It it much better to include the version in the file name (in some way). I just need to check the test suite for witness validation first, since it depends on the file name.

schuessf: Okay, I just renamed the files to indicate ther format version, s.t. the regression tests still consider the witnesses.

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/yaml/YamlWitnessWriterV2.java
```diff
@@ -70,16 +70,17 @@ public String toString(final Witness witness) {
...
-			if (!ghostVariables.isEmpty()) {
-				final Map<String, Object> ghostInstrumentation = new LinkedHashMap<>();
-				ghostInstrumentation.put("entry_type", "ghost_instrumentation");
-				ghostInstrumentation.put("metadata", mMetadataProvider.getFreshMetadata());
-				ghostInstrumentation.put("ghost_variables", ghostVariables);
-				ghostInstrumentation.put("ghost_updates", extractGhostUpdates(witness));
-				resultEntries.add(ghostInstrumentation);
+		final List<Map<String, String>> ghostVariables = extractGhostVariables(witness);
+		if (!ghostVariables.isEmpty()) {
+			if (!mWriteFunctionContractsAndGhostVariables) {
```
maul-esel: Oh, is the idea that this flag indicates the format version 2.0 vs 2.1?

schuessf: 👍

maul-esel: Could we either rename it, or add a comment to document that fact?

schuessf: Currently this is the case, yes. But I don't think the flag should directly indicate this, but rather the caller in `YamlWitnessWriter::construct` should ensure this.
Because if the witness format might be extended in the future with some additional feature in version 2.2, it seems ugly to have two parameters `isVersion21` and `isVersion22` (or sth. similar).
So, in my opinion, the best option is to seperate the parameters for function contracts and ghost variables.

bahnwaerter (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedGhostUpdate.java
maul-esel: Hm, I can't think of something meaningful here either. Disregard my comment.

bahnwaerter (approved):

bahnwaerter:
An important note: The current implementation does not follow the intended [witness schema for ghost variables](https://github.com/ultimate-pa/concurrency-witnesses/blob/main/correctness-witness-schema.yml). The following things are different and should be changed in my opinion (in order to remain backwards compatible with the original style of the schema):
- `ghost_variables` and `ghost_updates` are not put under `content` (like the `invariant_set` entry type does).
- Expressions of `ghost_variables` and `ghost_updates` are not represented correctly. They are composed of a `value` and `format` field instead of directly assigning the value to the `expression` field.

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
schuessf: Since the format was now adapted and Goblint also supports the new format, I removed ghost variables in version 0.1.

bahnwaerter (approved):
Looks great! Now the implementation follows our final [correctness witness schema with ghosts](https://github.com/ultimate-pa/concurrency-witnesses/blob/main/correctness-witness-schema.yml).

trunk/examples/witness-checking/regression/concurrent/AutomizerC.xml
```diff
@@ -0,0 +1,12 @@
...
+	<toolchain>
+		
+	<plugin id="de.uni_freiburg.informatik.ultimate.plugins.generator.cacsl2boogietranslator"/>
+	<plugin id="de.uni_freiburg.informatik.ultimate.boogie.procedureinliner"/>
+	<plugin id="de.uni_freiburg.informatik.ultimate.boogie.preprocessor"/>
+	<plugin id="de.uni_freiburg.informatik.ultimate.plugins.generator.rcfgbuilder"/>
+	<plugin id="de.uni_freiburg.informatik.ultimate.plugins.generator.traceabstraction"/>
+
+	</toolchain>
+</rundefinition>
```
maul-esel: To avoid future confusion, name this file `AutomizerCInline.xml`?

trunk/examples/witness-checking/regression/sequential/AutomizerC.epf
```diff
@@ -0,0 +1,77 @@
+#Mon Nov 27 17:23:30 CET 2017
+\!/instance/de.uni_freiburg.informatik.ultimate.core=
```
maul-esel, line 2: Do we need to duplicate this file? It should be enough to have one copy in `witness-checking/regression`, right?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainTranslator.java
```diff
@@ -237,6 +239,18 @@ private CHandlerTranslationResult performMainRun(final TranslationSettings trans
 		final MainDispatcher mainDispatcher = new MainDispatcher(mLogger, witness, locationFactory, typeHandler,
 				mainCHandler, ppHandler, acslHandler);
 
+		if (witness != null) {
+			// Add the declaration of global ghost variables
+			for (final var ghost : witness.getGlobalDeclarations()) {
+				final ExpressionResult res = ghost.getInitializationResult(mainDispatcher);
+				staticObjectsHandler.addStatementsForUltimateInit(res.getStatements());
+				for (final var d : res.getDeclarations()) {
+					staticObjectsHandler.addGlobalVarDeclarationWithoutCDeclaration((VariableDeclaration) d);
```
maul-esel: question: what are these declarations? how are they related to the declaration added after the loop? should all of them be global?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedGhostUpdate.java
```diff
@@ -0,0 +1,214 @@
...
+		case "pthread_rwlock_unlock":
+			return new ExpressionResultBuilder(expressionResult).addAllExceptLrValueAndStatements(witness)
+					.resetStatements(annotateAtomicCall(loc, expressionResult.getStatements(), witness.getStatements(),
+							MemoryModelDeclarations.ULTIMATE_PTHREADS_RWLOCK_UNLOCK.getName()))
+					.build();
+		case "pthread_create":
+			// Make the ghost update itself atomic and insert it just before the fork.
+			// TODO: Maybe we should do this atomically, but the CFG builder crashes for that case
+			// We are not sure, if this does have any different semantics.
+			return new ExpressionResultBuilder(expressionResult).addAllExceptLrValueAndStatements(witness)
```
maul-esel, line 206: We should look into this and see if CfgBuilder can be modified to support this; perhaps next week.

It's not a blocker for the merge though.

maul-esel (approved):
Looks good! I only left some minor comments.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainTranslator.java
schuessf: Previously `res` itself contained the declarations. This was however changed, this loop can be removed.

trunk/examples/witness-checking/regression/sequential/AutomizerC.epf
schuessf: No, this does not work at the moment. `AbstractRegressionTestSuite::getRunConfiguration` only searches for settings in the subfolders of the toolchains, but not vice versa. We could however adapt this method.

maul-esel: ah ok, my mistake

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainTranslator.java
maul-esel: Can the initialization require declarations of other variables? e.g. auxvars? (though they should probably not be global)

schuessf: Currently this is not possible, since we (luckily) restricted the ghost updates, s.t. they can only have a primitive type and their initial value can only be a constant expression. If we wanted to be less restrictive, we probably need to adapt this anway (as it seems, we cannot add aux-vars or local declarations to the init-procedure here).

maul-esel: I think we said in the schema that the expression may be any expression over global variables, right @bahnwaerter?

maul-esel: (side-effect-free and without undefined behaviour, of course)

schuessf: > The only variables allowed in this expression are the global variables of the program.
> The expression is evaluated after global variables declared in the program are initialized.

Okay, I see, then there might be aux-vars involved, e.g. if a ghost variable is initialized by a read of a global array.
But in that case the second sentence is problematic. Currently we initialize the ghost variables (in ULTIMATE.init) before the program. To match the specification, we would need to initialize the ghost variables at the very end, i.e., between dispatching the actual program, but before calling `PostProcessor` (where the ULTIMATE.init procedure is actually build). 
I tried to implement this in f6123d9, but I am not sure, if this is a nice solution...

schuessf:
This PR has canged the behaiour for location invariants, they are now added inside an atomic-block. Therefore some witness-tests (e.g. [this](https://jenkins.sopranium.de/job/Ultimate/job/Ultimate%20Nightly/view/change-requests/job/PR-653/24/testReport/de.uni_freiburg.informatik.ultimate.regressiontest.generic/WitnessRegressionTestSuite/I__witness_checking_regression_sequential_correctness_implicitunsignedconversion_2_c_witness_checking_regression_sequential_correctness_implicitunsignedconversion_2_c_witness_v0_1_yaml__S_witness_checking_regression_sequential_AutomizerC_epf_T__1240501109/)) fail now with an AssertionError: Atomic section must begin with ICFG location.

@maul-esel Since you worked recently on atomic blocks, why is this the case?

maul-esel:
The problem seems to be an atomic block in dead code, due to a wrong instrumentation of invariants pointing to labels. The instrumented Boogie program contains the following code:

```bpl
    ...
    #res := 0;
    return;
    atomic {
        assert false;
    }
  ERROR:
    assert { :reach "ERROR_FUNCTION","reach_error" } false;
    ...
```

The witness specifies the invariant `false` at the location for the label. I don't know if we have definitively fixed the semantics for location invariants at labels, but I would expect them to hold whenever one jumps to the label. I've committed a fix for these cases, have a look if you agree.

maul-esel:
Incidentally, to add regression tests for these kinds of bugs (the instrumentation was wrong, even if the atomic wouldn't have caused it to crash) we would need to support tests where we expect a witness to be rejected. We should add that in the future (maybe we can base it on our planned experiments comparing with other validators).

PR #654, opened 2023-10-31
Use SnakeYAML instead of eo-yaml for witnesses

schuessf:
Previously `eo-yaml` was used to parse and generate YAML-witnesses. However, there seems to be a performance issue. For slightly larger witnesses (~3MB) it took multiple hours for parsing! Therefore I exchanged `eo-yaml` with [SnakeYAML](https://bitbucket.org/snakeyaml/snakeyaml/) in this PR. SnakeYAML is able to parse even larger files within seconds. Additionally I regenerated the witnesses due to a slightly different output format (and to incorporate changes in the backtranslation).


danieldietsch (approved):
Looks like it even has a better API!

bahnwaerter (changes requested):
Thanks @schuessf for fixing the performance bug of the YAML parser.

@schuessf, @danieldietsch: Please add the new SnakeYAML packages and dependencies to the `p2-repository-ultimate-dependencies` first bevor the merging takes place. Otherwise we will observe build errors.

After that, feel free to merge your changes.

schuessf:
> Please add the new SnakeYAML packages and dependencies to the `p2-repository-ultimate-dependencies` first bevor the merging takes place. Otherwise we will observe build errors.

I already did that, otherwise the Jenkins-build would not work either. You can see SnakeYAML in the p2 plugin repo: https://struebli.informatik.uni-freiburg.de/p2-repository-ultimate-dependencies/plugins/

bahnwaerter (approved):
> I already did that, otherwise the Jenkins-build would not work either.

Perfect. Now I can see it in my Eclipse target definition as well. Everything should be fine now.

trunk/source/BA_SiteRepository/Ultimate_E4.17_Java11_Linux.target
```diff
@@ -37,6 +37,7 @@
 		</location>
 		<location includeAllPlatforms="false" includeConfigurePhase="true" includeMode="planner" includeSource="true" type="InstallableUnit">
 			<repository location="https://struebli.informatik.uni-freiburg.de/p2-repository-ultimate-dependencies/"/>
+			<unit id="org.yaml.snakeyaml" version="2.2.0"/>
 			<unit id="eo-yaml" version="7.0.8"/>
```
maul-esel, line 41: Shouldn't we remove all references to eo-yaml now?

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
```diff
@@ -65,115 +62,61 @@
...
-		final List<WitnessEntry> entries = new ArrayList<>();
-
-		for (final YamlNode witnessEntry : witnessEntries) {
-			final WitnessEntry newEntry = parseWitnessEntry(witnessEntry);
-			entries.add(newEntry);
-		}
-
-		return new Witness(entries);
+		final LoaderOptions loaderOptions = new LoaderOptions();
+		loaderOptions.setCodePointLimit(30 * 1024 * 1024);
```
maul-esel: Is there a specific reason for this number? (And can we make it a `final static private` constant?)

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/IYamlProvider.java
```diff
@@ -27,13 +27,13 @@
...
 
 /**
- * Interface to transform to eo-yaml, i.e. provide a {@link YamlNode}.
+ * Interface to transform data to a map for translation to yaml.
  *
  * @author Frank Schüssele (schuessf@informatik.uni-freiburg.de)
  */
 public interface IYamlProvider {
-	YamlNode toYaml();
+	Map<String, Object> toMap();
```
maul-esel: Should this really just return a map? And if so, we might want to rename the interface (though I don't like `IMapProvider` either...)

maul-esel (commented):
Thanks for taking care of this. I left some small comments.

trunk/source/BA_SiteRepository/Ultimate_E4.17_Java11_Linux.target
schuessf: If I remember correctly, @Heizmann currently uses eo-yaml for parsing SV-COMP YAMLs. So I would keep it for now.

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
schuessf: Sure, I thought I already did that.

trunk/source/BA_SiteRepository/Ultimate_E4.17_Java11_Linux.target
maul-esel: :+1: Ok. I propose we also switch that to the same library soon-ish (to keep things simple), but I guess it doesn't have to be in this PR.

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/IYamlProvider.java
schuessf: This is currently needed for serialization, because out-of-the-box SnakeYAML can only serialize "standard" types (string, int, List, Map,...). So this method just transforms the class to a map. There is also some way for SnakeYAML to do this directly, but I have not looked to that in detail.
But yes, I don't know a good name either. I have already thought about moving this interface into `Witness`, because it is specifically for this purpose.

schuessf: I just renamed it to `IMapSerializable `, but I am unsure, if this is better...

bahnwaerter (approved):

maul-esel (approved):

PR #655, opened 2023-11-07
Fix CArray expression evaluation/conversion to boolean value

bahnwaerter:
This patch fixes the `CArray` expression evaluation/conversion to a boolean value to allow a `CArray` evaluation in a boolean expression:
```
assert(a != b && "This is a CArray error message")
```

In this example, the `CArray` as part of the conjunction will be always evaluated with an address check (memory address of `CArray` != `NULL`) to the boolean value `true` since the address of the error message string literal can never be `NULL`.

### DIFF
--- a/trunk/examples/programs/regression/c/builtin/assert/assert_string-literal_safe_01.c
+++ b/trunk/examples/programs/regression/c/builtin/assert/assert_string-literal_safe_01.c
@@ -1,0 +1,14 @@
+//#Safe
+/*
+ * Valid assertion expression composed of a string literal.
+ * 
+ * Author: Manuel Bentele
+ *   Date: 2023-11-10
+ */
+
+int main()
+{
+    assert("Hello World!");
+
+    return 0;
+}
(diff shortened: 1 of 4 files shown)


schuessf:
The example does not quite show, what you want to change or do I missunderstand something here? The problematic example here would be `assert(a)`, where `a` is an array, right?

bahnwaerter:
> The example does not quite show, what you want to change or do I missunderstand something here? The problematic example here would be `assert(a)`, where `a` is an array, right?

Without the patch, Ultimate cannot handle `CArray`s in boolean expressions. And yes, your problematic example will be covered as well. In this case, the assertion should always hold (what it indeed does).

schuessf:
So, we also crash on `assert(a != b)` right now, or what is the behavior?

bahnwaerter:
> So, we also crash on `assert(a != b)` right now, or what is the behavior?

No, assume that `a` and `b` are valid terms (boolean expressions), then we only have problems if the string literal comes into play.

schuessf (approved):

danieldietsch (approved):

maul-esel (approved):

PR #657, opened 2023-12-04
Support atomic functions

schuessf:
This PR adds support for atomic functions (mostly read/write atomically without data-races) as defined in [stdatomic.h](https://en.cppreference.com/w/c/atomic). Since we currently only support preprocessed files from GCC, we need to support the builtin atomic functions from [here](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html).

Every atomic-function has an additional memory-order-argument (see [here](https://en.cppreference.com/w/c/atomic/memory_order)). For sequential consistency (which is just the constant `5` in GCC 🙂), we simply wrap every statement produced by the translation into an atomic-block. We currently do not support other memory orders (and will not in the near future), so I just try to overapproximate the other cases. Therefore the general translation (as handled in `StandardFunctionHandler::buildWrtMemoryOrder`) looks like this:
```
if (memory_order == 5) {
  atomic { <modelled reads/writes> }
} else {
  assert false { :overapproximation "memory order" };
}
```
For all the modelled function, we also introduce data-race-checks, because atomic read/writes can still lead to data races, if the read/writes of another thread is not atomic.

Note regarding SV-COMP: Currently theses functions are currently only used in AWS-benchmarks, which are not even concurrent. It does however not hurt, if we support them properly, they might occur in concurrent benchmarks in the future. However, benchmarks that model atomic reads/writes using `__VERIFIER_atomic_begin` and `__VERIFIER_atomic_end` with normal assignments are quite a bit easier for us to verify than atomic functions in C (because every function introduces additional memory accesses).


trunk/examples/concurrent/pthreads/races/regression/atomic_fetch.c
```diff
@@ -0,0 +1,25 @@
+//#Safe
+
+/*
+ * Author: Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ * Date: 2023-08-07
+ */
+
+typedef unsigned long pthread_t;
+
+int x, y;
```
bahnwaerter: Variable `y` is unused and can be omitted:

```suggestion
int x;
```

trunk/examples/concurrent/pthreads/races/regression/atomic_fetch_n.c
```diff
@@ -0,0 +1,24 @@
+//#Safe
+
+/*
+ * Author: Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ * Date: 2023-12-04
+ */
+
+typedef unsigned long pthread_t;
+
+int x, y;
```
bahnwaerter: Variable `y` is unused and can be omitted:

```suggestion
int x;
```

trunk/examples/concurrent/pthreads/races/regression/atomic_fetch_unsafe.c
```diff
@@ -0,0 +1,24 @@
+//#Unsafe
+
+/*
+ * Author: Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ * Date: 2023-08-07
+ */
+
+typedef unsigned long pthread_t;
+
+int x, y;
```
bahnwaerter: Variable `y` is unused and can be omitted:

```suggestion
int x;
```

trunk/examples/concurrent/pthreads/races/regression/atomic_store.c
```diff
@@ -0,0 +1,26 @@
+//#Safe
+
+/*
+ * Author: Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ * Date: 2023-12-04
+ */
+
+typedef unsigned long pthread_t;
+
+int x, y;
```
bahnwaerter: Variable `y` is unused and can be omitted:

```suggestion
int x;
```

trunk/examples/concurrent/pthreads/races/regression/atomic_store_n.c
```diff
@@ -0,0 +1,24 @@
+//#Safe
+
+/*
+ * Author: Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ * Date: 2023-12-04
+ */
+
+typedef unsigned long pthread_t;
+
+int x, y;
```
bahnwaerter: Variable `y` is unused and can be omitted:

```suggestion
int x;
```

trunk/examples/concurrent/pthreads/races/regression/atomic_store_unsafe.c
```diff
@@ -0,0 +1,25 @@
+//#Unsafe
+
+/*
+ * Author: Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ * Date: 2023-12-04
+ */
+
+typedef unsigned long pthread_t;
+
+int x, y;
```
bahnwaerter: Variable `y` is unused and can be omitted:

```suggestion
int x;
```

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -938,6 +976,250 @@ private List<Statement> makeVarargAssignment(final ILocation loc, final LRValue
...
+		}
+		return buildWrtMemoryOrder(loc, builder, memoryOrder);
+	}
+
+	private ExpressionResult buildWrtMemoryOrder(final ILocation loc, final ExpressionResultBuilder atomicBlockBuilder,
+			final ExpressionResult memoryOrder) {
+		// Check the memory order
+		// - "5" means sequential consistency, if this is the case make an atomic block of those statements
+		// - Otherwise we cannot say anything, since we only support sequential consistency (so we just overapproximate)
+		// TODO: Is this always 5?
```
bahnwaerter: Currently, the `memory_order_seq_cst` enum specifier always has the fixed value `5` according to the `memory_order` enum from `stdatomic.h` since C11.

Nevertheless, it would be nicer if we didn't hard code the value, but let it be resolved using our enum specifier utilities. Then atomic support would still work even if the `memory_order` enum were changed in future C standards. However, this requires that we can include header files and parse them correctly. Since this is not possible yet, we have to leave it with your version for now. But we should solve the TODO as soon as the included support is available.

bahnwaerter (changes requested):
Thanks @schuessf for the great atomic support.

As usual, I commented on a few little things that could still be improved. Otherwise your changes look good to me and can be merged.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: I agree, it would be much better! Currently we can only work with preprocessed files, so  I also implemented atomic-support for preprocessed files.

bahnwaerter (approved):

bahnwaerter (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/PreRunner.java
```diff
@@ -172,6 +172,17 @@ public int visit(final IASTParameterDeclaration declaration) {
...
 
+	private static boolean isAtomicFunctionCall(final IASTNode node) {
+		if (!(node instanceof IASTFunctionCallExpression)) {
+			return false;
+		}
+		final IASTExpression functionName = ((IASTFunctionCallExpression) node).getFunctionNameExpression();
+		if (!(functionName instanceof IASTIdExpression)) {
+			return false;
+		}
+		return ((IASTIdExpression) functionName).getName().toString().startsWith("__atomic");
```
maul-esel: Wouldn't it be safer to have an explicit list of these function names?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/PreRunner.java
```diff
@@ -172,6 +172,17 @@ public int visit(final IASTParameterDeclaration declaration) {
 		return super.visit(declaration);
 	}
 
+	private static boolean isAtomicFunctionCall(final IASTNode node) {
```
maul-esel: We might want to do the same thing also for other builtin functions (`pthread_create`, `pthread_mutex_*`, ...). It's not really an issue of "atomics", it's a question of whether we can model the function internally without referring to the memory array.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/PreRunner.java
```diff
@@ -181,7 +192,10 @@ public int visit(final IASTExpression expression) {
 			// final ILocation loc = LocationFactory.createCLocation(expression);
 			final ILocation loc = LocationFactory.createIgnoreCLocation(expression);
 			final IASTUnaryExpression ue = (IASTUnaryExpression) expression;
-			if (ue.getOperator() == IASTUnaryExpression.op_amper) {
+			// TODO 2024-01-09 Frank: We want to avoid the detection of & operators in atomic-functions, so we ignore
+			// this case here. However, the restriction to these special functions seems arbitrary, there are possibly
+			// more cases. What is a suitable check?
```
maul-esel: See comment above: Maybe `isAtomicFunctionCall` should be replaced with a more general method that determines if the function is a builtin function which we can model without accessing the memory array at the given argument? (in this case, we should however include the parameter index; for some functions the answer may depend on it)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -409,6 +412,41 @@ private Map<String, IFunctionModelHandler> getFunctionModels() {
...
+		fill(map, "__atomic_fetch_and", this::handleAtomicFetchAnd);
+		fill(map, "__atomic_fetch_or", this::handleAtomicFetchOr);
+		fill(map, "__atomic_fetch_xor", this::handleAtomicFetchXor);
+
+		fill(map, "__atomic_test_and_set", this::handleAtomicTestAndSet);
+		fill(map, "__atomic_clear", this::handleAtomicClear);
+
+		fill(map, "__atomic_compare_exchange",
+				(main, node, loc, name) -> handleUnsupportedFunctionByOverapproximation(main, loc, name,
+						new CPrimitive(CPrimitives.BOOL)));
```
maul-esel, line 458: Would it be difficult to implement this?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
maul-esel: Maybe in the meantime we can at least use a `static final int MEMORY_ORDER_SEQ_CST = 5;` ?

schuessf:
I just added an optimization to improve the performance of our verification on atomic functions.
Most of the atomic function that are supported with this PR require pointers. So if you have a variable `x` with a primitive type, you always need its address `&x` in the calls of atomic functions like `__atomic_store_n(&x, val, 5)`. Since the programs takes the address of `x`, we add this variable to the heap and thus it is translated to a pointer in Boogie. So the translation of this C-statement is similar to `write~...(x, val)` (which is basically a write to the memory-array).
However, if the address is only used for atomic functions (there are possible other cases where this applies as well), we do not really need to model x using a pointer, we can then simply translate it to an assignment `x := val`. Therefore we do not introduce any operations on the memory array, if only primitive types are used in atomic operations.
I had to add an exception for `&` to ignore it if it occurs in an atomic function in 4dfe1f3e10f5abd8d98b1e371f9e78b9d4aa3126. While it works as expected in our case, this is probably more specific than necessary and we should try to generalize it.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -938,6 +976,249 @@ private List<Statement> makeVarargAssignment(final ILocation loc, final LRValue
 		}
 	}
 
+	private Result handleAtomicClear(final IDispatcher main, final IASTFunctionCallExpression node, final ILocation loc,
```
maul-esel, line 1037: (I haven't checked the code of each of these functions in detail yet.)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: You need to understand the [specification](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#index-_005f_005fatomic_005fcompare_005fexchange_005fn) first (which is rather complicated) 😄:
>This built-in function implements an atomic compare and exchange operation. This compares the contents of *ptr with the contents of *expected. If equal, the operation is a read-modify-write operation that writes desired into *ptr. If they are not equal, the operation is a read and the current contents of *ptr are written into *expected. weak is true for weak compare_exchange, which may fail spuriously, and false for the strong variation, which never fails spuriously. Many targets only offer the strong variation and ignore the parameter. When in doubt, use the strong variation.
> 
>If desired is written into *ptr then true is returned and memory is affected according to the memory order specified by success_memorder. There are no restrictions on what memory order can be used here.
>
>Otherwise, false is returned and memory is affected according to failure_memorder. This memory order cannot be __ATOMIC_RELEASE nor __ATOMIC_ACQ_REL. It also cannot be a stronger order than that specified by success_memorder.

The implementation itself should not be too difficult, but I did not want to spend too much time to understand the specificationand it does not seems as important as load/store/... to me.

maul-esel (approved):
Looks good to me. Thanks for putting in all this effort!

PR #662, opened 2024-03-20
Backtranslate Boogie to ACSL, fix backtranslation issues

schuessf:
Currently our backtranslation from Boogie to C (`CACSL2BoogieBacktranslator`) returns an `IASTExpression`. However, we return our own implementation `FakeExpression` in most cases, which is bascially a string. Therefore we do not create an actual AST, but do some string concatenation instead.

Therefore this PR aims to translate Boogie to an actual AST in ACSL. To output these ACSL-expressions in witnesses or on the console the already existing class `ACSLPrettyPrinter` is used, which might be extended/improved (e.g. avoid unnecessary parentheses based on the operator precedence). Additionally I extracted a class `Boogie2ACSL` for this to extract functionality from `CACSL2BoogieBacktranslator` and translate the Boogie expressions mostly independent from the underlying C-Location.

Also I fixed some issues of the backtranslation:
* Introduce casts in backtranslation (d75e13eb592b736a4271bfad05f393467e82d689) to capture the translation of Boogie's unbounded integers to C properly and avoid unwanted overflows and wraparounds (e.g. for addition/multiplication/...). To avoid too many unnecessary casts, we try to track the possible value of every backtranslated (integer) expression and only introduce casts if necessary (i.e. the backtranslated expression does not fit in the underlying type)
* Fix the backtranslation of large integer literals (a0e59a304ebf9e6f4ffcde9116f142b0d2e30ab6): For some literals we need to add an unsigned suffix, 128bit literals even need to be split using bitwise operations (see https://stackoverflow.com/questions/16088282/is-there-a-128-bit-integer-in-gcc)

Before I rebased this branched, I checked if the new backtranslation via ACSL still produced the same String as the current backtranslation, see the results on [Jenkins](https://jenkins.sopranium.de/job/Ultimate/job/Ultimate%20Nightly/job/wip%2Ffs%2Fbacktranslation-casts/14/)


trunk/examples/Backtranslation/regression/c/standard/test06_false-unreach-call_GenericAutomizerC_Generic.errorpath
```diff
@@ -9,21 +9,21 @@
 [L34]  EXPR        \read(a)
        VAL         [a={6:0}, p1={5:0}, p2={2:0}]
 [L34]  CALL, EXPR  f(a, &p1, &p2)
-       VAL         [\old(a)=4, p1={5:0}, p2={2:0}]
+       VAL         [\old(a)=4, \old(p1)={5:0}, \old(p2)={2:0}]
```
danieldietsch, line 12: Do you know why we still sometimes have old and "current" values, and sometimes both (e.g., see below)

schuessf: It looks like there was a bug in the old backtranslation s.t. we did not translate old-pointers properly. See also in line 14 of the same file `[\old(a)=4, p1={5:0}, p1={5:0}, p2={2:0}, p2={2:0}]`, where there are two entries for `p1` and `p2` each, I presume one of those entries should be old variables.

The variables that are present in the states depend highly on the settings that are used (Kojak/Automizer, with/without blockencoding, ...).

danieldietsch: Yes, but these tests should use single statement

maul-esel: What would we expect for single-statement?

- For a global variable, is there any reason why we wouldn't always have both `old` and current value?
- For a parameter, I wouldn't be surprised if for technical reasons, we might in the beginning only have `old` -- the local Boogie variable for "current" is only declared and initialized in the body.
  + After the initialization, I would have assumed we always get both.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/AcslTypeUtils.java
```diff
@@ -101,4 +101,54 @@ private static CPrimitives translatePrimitiveAcslTypeToCType(final ACSLType t) {
 			throw new UnsupportedOperationException("Unhandled ACSL type " + t);
 		}
 	}
+
+	public static ACSLType translateCTypeToAcslType(final CType type) {
+		return new ACSLType(cTypeToString(type));
+	}
+
+	private static String cTypeToString(final CType type) {
```
maul-esel: Could we move this functionality to `CType` and/or `CPrimitive`? E.g. we can even declare the strings in the enum à la

```java
enum CPrimitive {
  CHAR(CPrimitiveCategory.INTTYPE, "char"),
  SCHAR(CPrimitiveCategory.INTTYPE, "signed char"),
  ...
}
```

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -0,0 +1,657 @@
...
+		mReporter.accept(
+				"Expression type not yet supported in backtranslation: " + expression.getClass().getSimpleName());
+		return null;
+	}
+
+	private BacktranslatedExpression translateIdentifierExpression(
+			final de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression expr, final ILocation context) {
+		final String boogieId = expr.getIdentifier();
+		if (boogieId.equals(SFO.RES)) {
+			// TODO: Can we somehow get the return type here?
```
maul-esel: Does the lack of a type cause any known problems? Maybe we can get it from the `context`?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -0,0 +1,657 @@
...
+		}
+	}
+
+	private static String bitvecToString(final BitvecLiteral lit) {
+		final String binStr = new BigInteger(lit.getValue()).toString(2);
+		assert binStr.length() <= lit.getLength() : "Binary string cannot be longer than bitvector literal length";
+		final int missingZeros = lit.getLength() - binStr.length();
+		if (missingZeros > 0) {
+			final String formatStr = "%" + lit.getLength() + "s";
+			return String.format(formatStr, binStr).replace(' ', '0');
```
maul-esel, line 199: You can rewrite the format string to do 0-padding directly

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -0,0 +1,657 @@
...
+		}
+		return binStr;
+	}
+
+	// TODO: Currently we don't care about types here, since function applications should only occur in bitvector
+	// setting, where we should not need any casts.
```
maul-esel, line 205: Why do we not need casts in the bitvector setting?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -0,0 +1,657 @@
...
+		if (reversed == null) {
+			mReporter.accept("Cannot identify Boogie2SMT function " + function);
+			return null;
+		}
+		final Integer bitSize = reversed.getSecond();
+
+		switch (reversed.getFirst()) {
+		case "sign_extend":
+		case "zero_extend":
+			// TODO: This might be problematic for signed types!
```
maul-esel, line 227: Is it?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -0,0 +1,657 @@
...
+					new IntegerLiteral(String.valueOf(1L << bitSize))));
+		case "bvmul":
+			return new BacktranslatedExpression(new BinaryExpression(Operator.ARITHMUL,
+					translatedArguments[0].getExpression(), translatedArguments[1].getExpression()));
+		case "bvsub":
+			final Expression sub = new BinaryExpression(Operator.ARITHMINUS, translatedArguments[0].getExpression(),
+					translatedArguments[1].getExpression());
+			return new BacktranslatedExpression(
+					new BinaryExpression(Operator.ARITHMOD, sub, new IntegerLiteral(String.valueOf(1L << bitSize))));
+		case "bvand":
```
maul-esel, line 252: (It looks like the code below could be simplified a bit.)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -0,0 +1,657 @@
...
+			resultType = new CPrimitive(CPrimitives.INT);
+			break;
+		case COMPNEQ:
+			minValue = BigInteger.ZERO;
+			maxValue = BigInteger.ONE;
+			operator = Operator.COMPNEQ;
+			resultType = new CPrimitive(CPrimitives.INT);
+			break;
+		case LOGICAND:
+			// TODO: This is only an overapproximation, if the expression is in NNF, can we assume this?
```
maul-esel: Seems risky. Can we check during the backtranslation whether we have an even or odd number of negations before this expression?

maul-esel (commented):
Thanks for taking this on! It looks good, I added some smaller comments.

Two questions that came to my mind:
- Is the wrapper `BacktranslatedExpression` useful outside of `Boogie2ACSL`?
- I see you removed `containsProcedureCall`, but I didn't quite grasp how the functionality works now. Could you explain?

danieldietsch (approved):

trunk/examples/Backtranslation/regression/c/standard/test06_false-unreach-call_GenericAutomizerC_Generic.errorpath
schuessf: This PR should not change anything fundamental, it only fixes pointer-old-vars. The maybe unexpected values come probably from Matthias' Live CFG.

schuessf:
> Is the wrapper `BacktranslatedExpression` useful outside of `Boogie2ACSL`?

Currently not, it is mainly used for two different reasons:
*  `BacktranslatedExpression` has a `CType` that might be used
* If ACSL-`Expression` would be used as a return, we need to prefix the full package name in `CACSL2BoogieBacktranslator`, since it already uses Boogie-`Expression`

> I see you removed `containsProcedureCall`, but I didn't quite grasp how the functionality works now. Could you explain?

We do not create any function-calls in our ACSL-backtranslation (I am not entirely sure, if our ACSL currently allows this). So this is no more needed.

Previously we used the `CLocation` of the Boogie-expression in the backtranslation. We only translated `IdentifierExpression`, if it contained an `IASTDeclaration` or `IASTExpression`. This way we wanted to translate variable, if they did not represent a C-variable, but another expression (e.g. `f(x)`, `x++`). I already removed this backtranslation of non-identifiers in 5f661d159df0abe32d0264e1ae9f08a3380d6099, because it was not handled correctly anyway and we should not have C-expressions with side-effects in invariants (`x++ == f(x)` is a bad invariant 😄). Now we backtranslate variables only to C-variables if possible and independant of the location.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/AcslTypeUtils.java
schuessf: Good idea 👍

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: If we ommit a type, we might not add casts, so it could be problematic. You are right, we could get it from the context here 👍

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: Thanks, I will change this, I just copied it from `CACSL2BoogieBacktranslator`

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: I did not check this in detail, but for functions like `bvadd` we do not require any casts for soundness, since the result of the function application is restricted by the number of bits, so we do a modulo instead. However, using ranges we might avoid unnecessary modulos.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: This is just copied from `CACSL2BoogieBacktranslator` again. I need to check the extension for bitvectors.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: Yes, we could just set the operator here. I am not quite sure, if this is much better. Maybe some of the bitvector-operators might require some special handling (like modulo), so this might also change in the future. I just adapted the code from `CACSL2BoogieBacktranslator`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: Yes, I could add a flag for this, that is probably the simplest solution. This was already present before -- as a hotfix for SV-COMP 😉

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: Unfortunately this does not work, 0 padding is only supported for numbers, but we want to pad the binary string here.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: I just checked it, it should be fine.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: I am not too familiar with our bitvector translation, so I might discuss this with backtranslation with Matthias first.

bahnwaerter (approved):

maul-esel (approved):

PR #668, opened 2024-05-27
Simplify implementation of YAML witnesses

schuessf:
This PR attempts to simplify the implementation for YAML witnesses. The format has completely changed just before SV-COMP 2024 and thus the previous implementation for YAML witnesses was just an ad-hoc adaptation to the changed format.

* Firstly the format for function contracts was changed, such that function contracts are also contained in the `invariant_set` (as in version 2.0), therefore the format for function contracts is downwards compatible and thus the format version was changed from 3.0 to 2.1
* Previously each `WitnessEntry` had the methods `toMap` and `toSetEntry` for the string-conversion to format version 0.1 and 2.0 respectively. Instead of these methods, we have now two different writer-classes that decide which entries should be considered (function contracts should only be exported in version >= 2.1) and how the entries should be formatted in YAML.
* The witness entry classes were initially made to represent the AST. However, since there are now multiple format versions, we use the classes just as container objects to store the necessary informations (e.g. no metadata). Therefore several classes were removed and we avoid parsing the unnecessary informations in `YamlWitnessParser`. Metadata are only needed for witness export, which are now produced using the introduced class `MetadataProvider`
* Previously each witness entry had a set of labels as an unique identifier, to ensure in `MainDispatcher` that each entry is only instrumented once. This is however only specific to GraphML-witnesses. Therefore I added an interface `IExtractedCorrectnessWitness` as an abstraction for the extracted witness entries from GraphML and YAML witnesses and only the GraphML-implementation uses these labels internally. For YAML witnesses we can already ensure that each entry is only instrumented once during extraction.

After this PR is merged, #653 needs to be reworked.


trunk/examples/witness-checking/regression/correctness/contract.c-witness.yml
```diff
@@ -15,13 +15,13 @@
...
-      column: 1
-      function: f
-    ensures:
-    - global > \old(global) && \result == \old(global)
-    format: acsl
\ No newline at end of file
+  - function_contract:
+      location:
+        file_name: contract.c
+        file_hash: UNUSED
```
danieldietsch, line 22: 😅

trunk/examples/witness-checking/regression/correctness/contract.c-witness.yml
```diff
@@ -1,8 +1,8 @@
-- entry_type: entry_set
+- entry_type: invariant_set
   metadata:
-    format_version: '3.0'
-    uuid: 45befebe-7a0b-4bac-9d4f-25b33879955f
-    creation_time: '2024-02-06T11:15:01+01:00'
+    format_version: '2.1'
```
danieldietsch, line 3: Why is the format version downgraded?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/GraphMLExtractedCorrectnessWitness.java
```diff
@@ -11,11 +38,17 @@
...
+ * instrumented once, store the labels of the GraphML nodes alongside.
+ *
+ * @author Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ */
 public class GraphMLExtractedCorrectnessWitness implements IExtractedCorrectnessWitness {
 	private final Map<IASTNode, Pair<ImmutableSet<String>, ExtractedWitnessInvariant>> mInvariants;
 	private final Set<ImmutableSet<String>> mNodeLabelsOfAddedWitnesses = new HashSet<>();
 
-	public GraphMLExtractedCorrectnessWitness(
+	GraphMLExtractedCorrectnessWitness(
```
bahnwaerter: Do you omit the access modifier to want the constructor and methods to be package-private?
If you don't want to use them outside of the package in the future, then that's a good decision.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlExtractedCorrectnessWitness.java
```diff
@@ -47,15 +47,15 @@ public class YamlExtractedCorrectnessWitness implements IExtractedCorrectnessWit
...
 		mInvariants.addPair(node, entry);
 	}
 
-	public void addInvariants(final Map<IASTNode, ? extends ExtractedWitnessInvariant> invariantMap) {
+	void addInvariants(final Map<IASTNode, ? extends ExtractedWitnessInvariant> invariantMap) {
 		invariantMap.forEach(this::addInvariant);
 	}
 
-	public void addFunctionContract(final IASTNode function, final ExtractedFunctionContract contract) {
+	void addFunctionContract(final IASTNode function, final ExtractedFunctionContract contract) {
```
bahnwaerter: Do you omit the access modifier to want the methods to be package-private?
If you don't want to use them outside of the package in the future, then that's a good decision.

bahnwaerter (commented):
Your changes look great. Especially, the documentation comments increase code quality.
If omitting the access modifier is intentional, then I have nothing more to complain about.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/GraphMLCorrectnessWitnessExtractor.java
```diff
@@ -178,17 +180,21 @@ private static ExtractedWitnessInvariant mergeInvariants(final ExtractedWitnessI
 			newInvariant = builder.toString();
```
danieldietsch, line 174: why not just return ewi1?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/GraphMLExtractedCorrectnessWitness.java
schuessf: Yes, the constructor should be only accessible during construction by `GraphMLCorrectnessWitnessExtractor`, only the getters should be public. So maybe it would be even better to make this implementation a nested class in `GraphMLCorrectnessWitnessExtractor` (as it is also rather small and simple).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlExtractedCorrectnessWitness.java
schuessf: Yes, these methods should be only accessible during construction by `YamlCorrectnessWitnessExtractor`, only the getters should be public. So maybe it would be even better to make this implementation a nested class in `YamlCorrectnessWitnessExtractor` (as it is also rather small and simple).

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
```diff
@@ -80,112 +71,51 @@ public static Witness parseWitness(final File yamlInput) throws IOException {
...
 		case LocationInvariant.NAME: {
-			if (formatVersion.getMajor() != 0) {
-				throw new UnsupportedOperationException(
-						LocationInvariant.NAME + " is only allowed in format version 0.x");
-			}
 			final Location location = parseLocation((Map<String, Object>) entry.get("location"));
-			final Invariant locationInvariant = parseInvariant((Map<String, String>) entry.get(LocationInvariant.NAME));
-			return Stream.of(new LocationInvariant(metadata, location, locationInvariant));
+			final Map<String, String> invariant = (Map<String, String>) entry.get(LocationInvariant.NAME);
+			return Stream.of(new LocationInvariant(location, invariant.get("string"), invariant.get("format")));
```
danieldietsch, line 78: We got rid of the ugly cast, but now we have hardcoded field names in here?

danieldietsch: Ah, the cast is there as well ...

trunk/source/WitnessPrinter/META-INF/MANIFEST.MF
```diff
@@ -12,7 +12,8 @@ Require-Bundle: de.uni_freiburg.informatik.ultimate.lib.core,
  de.uni_freiburg.informatik.ultimate.core,
  edu.uci.ics.jung;visibility:=reexport,
  de.uni_freiburg.informatik.ultimate.plugins.generator.rcfgbuilder,
- de.uni_freiburg.informatik.ultimate.witnessparser
+ de.uni_freiburg.informatik.ultimate.witnessparser,
+ org.yaml.snakeyaml;bundle-version="2.2.0"
```
danieldietsch: Why this version constraint?

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/preferences/PreferenceInitializer.java
```diff
@@ -105,7 +105,7 @@ public enum WitnessVerifierType {
 			"Prevent the generation of invariants in GraphML which require ACSL syntax";
 
 	public static final String LABEL_YAML_FORMAT_VERSION = "YAML Format version";
-	private static final String VALUE_YAML_FORMAT_VERSION = "3.0";
+	private static final String VALUE_YAML_FORMAT_VERSION = "2.0";
```
danieldietsch, line 108: This version downgrade really confuses me. 
Furthermore, it looks like we dont have any way to support different versions, right? Do we even check whether the format version is the one we actually support?

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/yaml/YamlCorrectnessWitnessGenerator.java
```diff
@@ -40,72 +34,62 @@ public class YamlCorrectnessWitnessGenerator {
...
-			entries.addAll(extractFunctionContracts(allProgramPoints, metadataSupplier, hash, formatVersion));
-		}
-		final Witness witness = new Witness(entries);
-		if (formatVersion.getMajor() < 2) {
-			return witness;
-		}
-		return new Witness(List.of(witness.toInvariantSet(metadataSupplier)));
+		final var invariants = extractInvariants(allProgramPoints);
+		final var functionContracts = extractFunctionContracts(allProgramPoints);
+		return new Witness(DataStructureUtils.concat(invariants, functionContracts));
```
danieldietsch, line 67: Is the sorting taken care of now? Or does this not apply anymore?

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/yaml/YamlWitnessWriter.java
```diff
@@ -0,0 +1,34 @@
...
+public abstract class YamlWitnessWriter {
+	public static YamlWitnessWriter construct(final FormatVersion formatVersion,
+			final MetadataProvider metadataProvider) {
+		if (formatVersion.getMajor() == 0) {
+			return new YamlWitnessWriterV0(metadataProvider);
+		}
+		if (formatVersion.getMajor() == 2) {
+			return new YamlWitnnesWriterV2(metadataProvider, formatVersion.getMinor() >= 1);
+		}
+		throw new UnsupportedOperationException("Unknown format version " + formatVersion);
```
danieldietsch, line 57: I am getting even more confused about versions now. Here we seem to talk about 0 and 2, but throw an exception for 3?

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/yaml/YamlWitnessWriterV0.java
```diff
@@ -0,0 +1,55 @@
...
+public class YamlWitnessWriterV0 extends YamlWitnessWriter {
+	private final MetadataProvider mMetadataProvider;
+
+	public YamlWitnessWriterV0(final MetadataProvider metadataProvider) {
+		mMetadataProvider = metadataProvider;
+	}
+
+	@Override
+	public String toString(final Witness witness) {
+		return formatYaml(
```
danieldietsch, line 55: Why are all other entry types thrown away?

trunk/examples/witness-checking/regression/correctness/contract.c-witness.yml
schuessf: There was never an official version 3.0. We wanted to extend version 2.0 to also support function contracts. Our initial idea was to completely change the format, therefore we also wanted to go from version 2.0 to 3.0. Recently we decided to change the format such that it is downwards compatible (you can see it also in this witness file), therefore we also changed the format version to 2.1.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/GraphMLCorrectnessWitnessExtractor.java
schuessf: Here we create the disjunction of the invariants from ewi1 and ewi2, also with the union of their node labels.

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
schuessf: This PR did not really change the behavior, the cast and the hardcoded field names already have been there before. With this PR we only extract some necessary fields and without the usage of another method. But I agree that both of the issues are not nice, for the cast I cannot see another solution...

trunk/source/WitnessPrinter/META-INF/MANIFEST.MF
schuessf: I am not sure, I simply copied this line from the manifest of `WitnessParser`. Is it better to ommit this version constraint?

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/preferences/PreferenceInitializer.java
schuessf: For the SV-COMP version 2.0 should be still the standard. Therefore I decided to change the default from the experimental version (previously 3.0, now 2.1) to the "official" version 2.0.

The format version is checked in `YamlWitnessWriter::construct`, where we try to get the writer that matches the format version. If we do not support the format version, we throw an exception.

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/yaml/YamlCorrectnessWitnessGenerator.java
schuessf: This TODO got lost during refactoring 😅
If we do not sort, we rely on the order of `mIcfg.getProgramPoints()`, which seems to be consistent for `BasicIcfg` the very least, so maybe this is not a critical issue -- it might however not hurt to keep the TODO.

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/yaml/YamlWitnessWriter.java
schuessf: There was the format version 0.1 before the SV-COMP24, in SV-COMP itself version 2.0 was used and is still the current version (version 1.x was simply skipped 😅). We plan to extend this version to be downwards compatible (this already is implemented for function contracts), therefore we use version 2.1 for these extensions. There is currently no version 3.0 (not even "unofficial").

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/yaml/YamlWitnessWriterV0.java
schuessf: We now always create the same `Witness`-object, independant from the format version. Therefore this object may also contain e.g. function contracts that are not supported in version 0.1. The writer now decides which entries should be exported to YAML.

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
danieldietsch: Yeah, that would require even more work

trunk/source/WitnessPrinter/META-INF/MANIFEST.MF
danieldietsch: It is definitely better. In particular if there is only one snakeyml.

trunk/source/WitnessPrinter/src/de/uni_freiburg/informatik/ultimate/witnessprinter/preferences/PreferenceInitializer.java
danieldietsch: So basically, having 3.0 was the actual mistake.

schuessf: Yes, if we want to have a "stable" version as a default 😉

bahnwaerter (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/GraphMLExtractedCorrectnessWitness.java
bahnwaerter: Sure, you could implement it as a nested class, but I think you can leave it as it is.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlExtractedCorrectnessWitness.java
bahnwaerter: Sure, you could do it as a nested class, but I think you can leave it as it is.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedFunctionContract.java
```diff
@@ -77,9 +76,15 @@ private ACSLNode parseString(final ILocation loc, final String string) {
 		}
 	}
 
-	public Stream<ACSLNode> transformToAcsl() {
+	public List<ACSLNode> getAcslContracts() {
```
maul-esel: _minor:_ should this be `getAcslContractClauses` or similar? I was confused by the idea that there would be multiple contracts.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/GraphMLCorrectnessWitnessExtractor.java
```diff
@@ -105,35 +106,34 @@ protected ExtractedCorrectnessWitness extractWitness() {
...
-		rtr.addWitnessStatements(map);
 		mLogger.info("Processed " + closed.size() + " nodes");
-		return rtr;
+		return new GraphMLExtractedCorrectnessWitness(map);
 	}
 
-	protected Map<IASTNode, ExtractedWitnessInvariant> mergeMatchesIfNecessary(
-			final Map<IASTNode, ExtractedWitnessInvariant> mapA, final Map<IASTNode, ExtractedWitnessInvariant> mapB) {
+	protected Map<IASTNode, Pair<ImmutableSet<String>, ExtractedWitnessInvariant>> mergeMatchesIfNecessary(
+			final Map<IASTNode, Pair<ImmutableSet<String>, ExtractedWitnessInvariant>> mapA,
```
maul-esel: Would it make sense to introduce a dedicated data type for `Pair<ImmutableSet<String>, ExtractedWitnessInvariant>` ?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/YamlCorrectnessWitnessExtractor.java
```diff
@@ -197,11 +195,46 @@ protected int genericVisit(final IASTNode node) {
 			// should be matched the first node of the line in that case) or it also matches the AST node
 			if (mLocation.getLine() == loc.getStartLine()
 					&& (mLocation.getColumn() == null || mLocation.getColumn() == loc.getStartColumn())) {
-				mMatchedNodes.add(node);
-				// skip the subtree if a match occurred, but continue with siblings.
-				return PROCESS_SKIP;
+				mMatchedNode = node;
+				// Stop the subtree, since we want the witness entry to be matched at most once.
```
maul-esel: "skip the subtree" :stuck_out_tongue:

maul-esel (approved):
This looks good to me. Thanks for all the work on improving this stuff :)

PR #670, opened 2024-08-09
Wip/lf/pea complement only

henkele:
Implements PEA complement and redundancy analysis

PR #671, opened 2024-08-09
Proof Refactoring

maul-esel:
This PR refactors how ULTIMATE works with proof artifacts (for now, primarily Floyd-Hoare annotations). The point is:
1. to make it easier to work with proofs, by making them first-class concepts (e.g. `IFloydHoareAnnotation`) rather than annotations in an Icfg. This allows e.g. considering the proofs of multiple CEGAR loops separately (a more elegant solution than the workaround used in #670).
2. to make the handling of proofs more flexible, in preparation for supporting other kinds of proofs; most importantly, _thread-modular proofs at many levels_ and _Owicki-Gries proofs_. These other kinds of proofs are e.g. needed for concurrency correctness witnesses.

I consider the PR more or less ready for review, if the currently running nightly tests don't show any more issues. The only change I still have planned from my side is to go through the `.epf` files and adapt them to the following change in the settings for `TraceAbstraction`:
> The setting `Compute Hoare Annotation of negated interpolant automaton, abstraction and CFG` has been removed. The existing setting `Positions where we compute the Hoare Annotation` has a possible value `NONE` (the default value) to disable Hoare computation completely.

However, I wanted to first ask if anyone objects to this or would propose a different settings structure, before I start going through all the settings files.


trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/IProof.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package de.uni_freiburg.informatik.ultimate.lib.proofs;
+
+/**
+ * A marker interface for different kinds of correctness proofs (wrt. different specifications) for a program.
+ *
+ * @author Dominik Klumpp (klumpp@informatik.uni-freiburg.de)
+ */
+public interface IProof {
+	// no methods here
```
bahnwaerter: Will proofs be able to implement interface functions in the future so that multiple implementations of a single proof is possible? If that is never the case, then an abstract class might be a more suitable generalization.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/floydhoare/HoareAnnotationComposer.java
```diff
@@ -284,8 +306,81 @@ public HoareAnnotationStatisticsGenerator getHoareAnnotationStatisticsGenerator(
...
+		Term renamedFormula = Substitution.apply(mgdScript, substitutionMapping, ps.getFormula());
+		renamedFormula = SmtUtils.simplify(mgdScript, renamedFormula, services, simplificationTechnique);
+		final IPredicate result = predicateFactory.newPredicate(renamedFormula);
+		return result;
+	}
+
+	/**
+	 * Construct Term which represents the same set of states as ps, but where all globalVars are renamed to
+	 * oldGlobalVars.
+	 *
```
bahnwaerter: Remove empty comment line!

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/floydhoare/HoareAnnotationPositions.java
```diff
@@ -0,0 +1,70 @@
...
+			hoareAnnotationLocs.addAll(IcfgUtils.getReturnPredecessorPoints(root));
+			hoareAnnotationLocs.addAll(IcfgUtils.getPotentialCycleProgramPoints(root));
+			break;
+		case LoopHeads:
+			hoareAnnotationLocs.addAll(root.getLoopLocations());
+			hoareAnnotationLocs.addAll(IcfgUtils.getCallerAndCalleePoints(root));
+			hoareAnnotationLocs.addAll(IcfgUtils.getReturnPredecessorPoints(root));
+			break;
+		case None:
+			return Set.of();
```
bahnwaerter, line 64: You can also return an empty set using `Collections.emptySet()` if it should be immutable.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
```diff
@@ -512,6 +488,10 @@ private void refineAbstractionInternal(final AutomatonType automatonType)
 		}
 	}
 
+	protected void performAbstractionSanityCheck() {
+		// Empty implementation. Subclasses may override this method.
+	}
```
bahnwaerter, line 493: Maybe declare it as an abstract method and add implementation in the subclasses?

trunk/source/TraceAbstractionConcurrent/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstractionconcurrent/TraceAbstractionConcurrentObserver.java
```diff
@@ -142,15 +142,15 @@ public boolean process(final IElement root) {
 		clrReporter.reportCegarLoopResult(result);
 		clrReporter.reportAllSafeResultIfNecessary(result, errNodesOfAllProc.size());
 
-		mLogger.info("Statistics - iterations: " + abstractCegarLoop.getIteration());
+		// mLogger.info("Statistics - iterations: " + abstractCegarLoop.getIteration());
```
bahnwaerter: Shouldn't be that printed anymore? If so, you can remove this line.

trunk/source/TraceAbstractionConcurrent/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstractionconcurrent/TraceAbstractionConcurrentObserver.java
```diff
@@ -142,15 +142,15 @@ public boolean process(final IElement root) {
...
 		// s_Logger.info("Statistics - biggest abstraction: " +
 		// abstractCegarLoop.mBiggestAbstractionSize + " states");
 		// s_Logger.info("Statistics - biggest abstraction in iteration: " +
 		// abstractCegarLoop.mBiggestAbstractionIteration);
 
 		String stat = "";
 		stat += "Statistics:  ";
-		stat += " Iterations " + abstractCegarLoop.getIteration() + ".";
+		// stat += " Iterations " + abstractCegarLoop.getIteration() + ".";
```
bahnwaerter: Shouldn't be that printed anymore? If so, you can remove this line.

bahnwaerter (approved):
The introduction of the new library for proofs looks great, at least for my knowledge on the subject. The new proof library has been created correctly. We just need to upgrade the Java version number in the `Manifest.MF` file later when we finally migrate to Java 21. In addition, @danieldietsch has to adjust the sonar coverage configuration in the Maven `pom.xml` under `trunk/source/BA_MavenParentUltimate` so that the code coverage for the library can be reported properly.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
maul-esel: That would force every subclass to provide an implementation. Most of them currently do not have anything meaningful to do here.

trunk/source/TraceAbstractionConcurrent/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstractionconcurrent/TraceAbstractionConcurrentObserver.java
maul-esel: Sure, removed. We should probably get around to deleting the entire project soon, as it is obsolete and unused, but complicates implementation of other projects.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/floydhoare/HoareAnnotationPositions.java
maul-esel: I've wondered about this in other places before -- do you know if either `Set.of()`  or `Collections.emptySet()` has an advantage over the other?

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/IProof.java
maul-esel: Currently, there are no common functions planned for the immediate future. But in principle, some could be added -- e.g. a representation of the specification that is proved by the instance.

I do not see the point of an abstract class here. That would restrict implementations (they cannot inherit from other base classes), but what benefit would we get?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
bahnwaerter: Okay, I see. Your implemented version would be the better choice.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/floydhoare/HoareAnnotationPositions.java
bahnwaerter: To my knowledge, there is a slight difference in the implementation:

````
Collections.emptySet().contains(null); // false
Set.of().contains(null); // NullPointerException
````

Also the serialized form might be different.

File: trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/IProof.java (hunk 1 above)
Comment (bahnwaerter) [reply trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/IProof.java]: Yeah your're right. I've seen later in the code review that my inital remark does not make sense anymore. Your model looks quite good to represent all possible objects related to proofs.

**Review by schuessf — APPROVED** [review]

**Comment by maul-esel** [comment]
Thanks for the reviews. I'll wait for the nightly test run; if no further issues occur I will merge this tomorrow.

**Comment by maul-esel** [comment]
Fixing the tests required some changes:
- The precondition for our Floyd-Hoare annotations is not necessarily `true`; instead, it is a conjunction of `(= g |old(g)|)` for all modifiable global variables `g` of the initial location's procedure. This requires supporting different preconditions per initial location, in the case of multiple initial locations.
- A bug in `BlockEncodingV2` caused contract computation to fail, as some procedures did not have an exit location. I've fixed this, and added some checks to prevent similar bugs in the future.

While doing this, I also added a feature required e.g. in the context of requirements analysis: Each proof now also has an explicit description of the specification that it proves. This information is also added to `InvariantResult` and `ProcedureContractResult` (in the form of `Check` objects).

File: releaseScripts/default/adds/Ultimate.py [hunk 7]
```diff
@@ -508,9 +508,9 @@ def create_cli_settings(prop, validate_witness, architecture, c_file):
         # we need to disable hoare triple generation as workaround for an internal bug
         # but only for reachability witness validation
         ret.append(
-            "--traceabstraction.compute.hoare.annotation.of.negated.interpolant.automaton,.abstraction.and.cfg"
+            "--traceabstraction.positions.where.we.compute.the.hoare.annotation"
```
danieldietsch, line 511: To be honest, I dont like options that huamnize Ultimate, i.e., "where we compute".

trunk/examples/LTL/regression/c/LTLAutomizerC.epf
```diff
@@ -205,7 +205,7 @@ file_export_version=3.0
 /instance/de.uni_freiburg.informatik.ultimate.plugins.blockencoding/Create\ interprocedural\ compositions=false
 /instance/de.uni_freiburg.informatik.ultimate.plugins.blockencoding/Iterate\ optimizations\ for\ n\ times\ (<\=0\ means\ until\ nothing\ changes)=0
 #Fri Dec 09 13:35:39 CET 2016
-/instance/de.uni_freiburg.informatik.ultimate.plugins.generator.traceabstraction/Positions\ where\ we\ compute\ the\ Hoare\ Annotation=All
```
danieldietsch, line 208: Why remove the explicit setting here?

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/cegar/BuchiCegarLoopFactory.java
```diff
@@ -116,8 +117,8 @@ public BuchiCegarLoopFactory(final IUltimateServiceProvider services, final TAPr
 				.getEnum(BuchiAutomizerPreferenceInitializer.LABEL_AUTOMATON_TYPE, AutomatonTypeConcurrent.class);
 		switch (automatonTypeConcurrent) {
 		case BUCHI_AUTOMATON:
-			final var automatonProvider = new Petri2FiniteAutomatonAbstractionProvider.Lazy<>(petriNetProvider,
-					stateFactoryForRefinement, new AutomataLibraryServices(mServices));
+			final var automatonProvider = new Petri2FiniteAutomatonAbstractionProvider.Lazy<>(mServices,
+					petriNetProvider, stateFactoryForRefinement);
```
danieldietsch, line 121: 👍

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/floydhoare/FloydHoareMapping.java
```diff
@@ -0,0 +1,79 @@
+/*
+ * Copyright (C) 2023 Dominik Klumpp (klumpp@informatik.uni-freiburg.de)
```
danieldietsch, line 2: Sitting there for a while :)

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/PrePostConditionSpecification.java
```diff
@@ -0,0 +1,117 @@
...
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.cfg.structure.IIcfg;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.cfg.structure.IcfgLocation;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.IPredicate;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.IPredicateUnifier;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.PredicateUtils;
+import de.uni_freiburg.informatik.ultimate.lib.smtlibutils.SmtUtils;
+
+/**
+ * A specification stating that any execution starting in an initial location and with a variable assignment satisfying
+ * a given precondition must, when (and if) it reaches a final location, have a current variable assignment satisfying a
```
danieldietsch: What is it, when or if?

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/PrePostConditionSpecification.java
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @author Dominik Klumpp (klumpp@informatik.uni-freiburg.de)
+ *
+ * @param <S>
+ *            the type of states (or "locations")
+ */
+public class PrePostConditionSpecification<S> implements ISpecification {
+	private final Map<S, IPredicate> mInitialStates;
+	private final Predicate<S> mIsFinalState;
+	private final IPredicate mPostcondition;
```
danieldietsch, line 79: * The initial locations are explicitly specified, but the final locations use a predicate. Is there a reason for that? Might be worth a comment :). 
* Are there any constraints on the number of final locations? Are these final locations in the automata-sense, or are these exit locations of the program?
* Am I understanding the idea here correctly: for each initial state and its precondition, all reachable final states satsify the post condition? So, if I wanted to split a program (i.e., construct a residual program), I would construct two of these specifications, one for the postcondition `false` and all initial states that guarantee that, and one for `true`. And I would then think that "final states" are not the exits of the program (which would be somewhat natural when talking about pre/postconditions), but the accepting states of the automaton, i.e., our error locations.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/floydhoare/FloydHoareMapping.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @param <S>
+ *            the type of states which are annotated
+ */
+public class FloydHoareMapping<S> implements IFloydHoareAnnotation<S> {
+	private final PrePostConditionSpecification<S> mSpecification;
+	private final Map<S, IPredicate> mAnnotation;
+	private final IPredicate mDefaultPredicate;
+
+	public FloydHoareMapping(final PrePostConditionSpecification<S> specification,
```
danieldietsch, line 59: Is there a reason to support both, null and a default predicate? Wouldnt it be easier to force a default predicate?

danieldietsch (commented):
Review so far, I would read on, but feel free to merge anyways,

releaseScripts/default/adds/Ultimate.py
maul-esel: :+1: I fully agree. This is an existing option, so for this PR I would keep it. When we do the settings restructuring that we discussed, we should change it.

trunk/examples/LTL/regression/c/LTLAutomizerC.epf
maul-esel: The setting is not removed, the value is changed to `None`. The reason is that the total effect of this setting with the removed line `/instance/de.uni_freiburg.informatik.ultimate.plugins.generator.traceabstraction/Compute\ Hoare\ Annotation\ of\ negated\ interpolant\ automaton,\ abstraction\ and\ CFG=false` below was to disable Hoare computation completely, which is now simply achieved with the setting value `None`.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/PrePostConditionSpecification.java
maul-esel: I've added some comments to hopefully clarify the first two questions.

I don't fully understand the 3rd question: Are you saying you would create a specification with _postcondition_ `true` ? That specification would be trivial, so it's perhaps not what you intended.
- I could split the specification by splitting the set of initial states, but keeping the same postcondition.
- Alternatively, I could split it by splitting the set of final states, again with the same postcondition.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/floydhoare/FloydHoareMapping.java
maul-esel: We use `null` to indicate that a location does not have any annotation. Since we have settings to only compute the annotation at certain locations (e.g. loop heads), this can be the case for ICFGs; and there is no non-null default predicate we could use in this case.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/PrePostConditionSpecification.java
maul-esel: I've reworded the text to be hopefully more clear.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/floydhoare/FloydHoareMapping.java
danieldietsch: Its just that having no annotation would mean I could always use true in that case, right?

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/PrePostConditionSpecification.java
```diff
@@ -42,23 +42,40 @@
...
  * There may be multiple initial locations, and the precondition that holds initially may differ depending on the
  * initial condition. This is for example needed in interprocedural programs, where (in "library mode") each procedure's
  * entry node is an initial location, and the corresponding precondition states that g = old(g) for all modifiable
  * global variables of the procedure.
  *
+ * The intuition of the "final" locations can differ depending on the context. For instance, they could consist of the
+ * error locations of a program, or the accepting states of an interpolant automaton, together with the postcondition
+ * {@code false}. In this case, the specification expresses unreachability of these states (see
+ * {@link #isUnreachabilitySpecification()}). Alternatively, one might use the exit location of a program (or procedure)
+ * with a postcondition specifying the expected variable values when the program terminates.
```
danieldietsch, line 58: 👍

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/PrePostConditionSpecification.java
```diff
@@ -42,23 +42,40 @@
...
+	// As there are typically relatively few initial states (or just one), listing them explicity should not be
+	// problematic, and it allows to check validity of a proof for this specification by simply performing a DFS
+	// starting from these states.
 	private final Map<S, IPredicate> mInitialStates;
+
+	// "Final" states, i.e., states where the postcondition must hold.
+	//
+	// These are not listed explicitly but only identified by a boolean function, because in some cases (e.g. concurrent
+	// programs) the set of final states can be very large and costly to compute up-front.
 	private final Predicate<S> mIsFinalState;
```
danieldietsch, line 77: :+1:

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/floydhoare/FloydHoareMapping.java
maul-esel: No, that would be different. Imagine an edge `src --(x:=0)--> tgt`, where `tgt` has an annotation like `x==y`. If we check the validity of this annotation (as in `FloydHoareValidityChecker`), we have to differentiate:
- If `src` has the annotation `true` (perhaps as default predicate), the Hoare triple `{true} x:=0 {x==y}` is not valid, and so the entire annotation is invalid.
- If `src` has the annotation `null`, it is clear that we do not have the full Hoare triple to check. But we don't consider the annotation invalid in this case.

danieldietsch: Ah, perhaps this should be a comment for this class :)

maul-esel: :+1: I've added this explanation to `IFloydHoareAnnotation` and linked to it in the documentation for the two different constructors here.

PR #673, opened 2024-08-21
Delete obsolete plugin TraceAbstractionConcurrent

maul-esel:
This PR deletes the plugin `TraceAbstractionConcurrent` which is no longer in use, but has often required hacks in other code to maintain the support.

Some code of this plugin which was used in `TraceAbstractionWithAFAs` has been moved there.


schuessf (approved):
Thanks for your work! These changes look good!
Just to make sure: Your only "additions" are that you moved two classes (`CFG2Automaton` and `Cfg2Nwa`) from the plugin `TraceAbstractionConcurrent` to `TraceAbstractionWithAFAs` and added some code to `TAwAFAsCegarLoop`?
* Are all the methods in the moved classes actually used? Or can some of them be also removed (e.g. because they were only used by `TraceAbstractionConcurrent`)?
* Where does the additional code from `TAwAFAsCegarLoop` stem from?

maul-esel:
- The code in `TAwFAsCegarLoop` is taken from `CegarLoopConcurrentAutomata`. `TAwFAsCegarLoop` used to inherit from this class, which in turn inherits from `NwaCegarLoop`. I have
  + made `TAwFAsCegarLoop` inherit directly from `NwaCegarLoop`,
  + copied the relevant code (i.e., overridden methods and constructor) from `CegarLoopConcurrentAutomata` to `TAwFAsCegarLoop`,
  + and deleted `CegarLoopConcurrentAutomata`.
 - The two moved classes are called, but I had not checked if some of their methods can be removed. But from a quick look, it seems all methods are necessary.

maul-esel:
@Heizmann also agreed, so I am merging (squashing).

PR #680, opened 2024-09-26
Support YAML violation witnesses

schuessf:
This PR adds support for violation witnesses in the [YAML format](https://sosy-lab.gitlab.io/benchmarking/sv-witnesses/yaml/violation-witnesses.html). The main work (thanks to @HelenAnnaMeyer) was the following:

* Add AST objects for the new witness format
* Allow Parsing violation witnesses from YAML (extend `YamlWitnessParser`).
* Generate AST object to represent violation witnesses from an `IProgramExecution` (see `YamlViolationWitnessGenerator`) 
* Validate violation witnesses (see `YamlWitnessProductAutomaton`): Create a product automaton of a program automaton and a witness to use this product as as the initial abstraction in the verification (`TraceAbstraction` / `BüchiAutomizer`).

In order to work properly, this required the following additional changes:

* Add `WitnessTransformer` as an abstraction for violation witnesses in YAML and GraphML
* Slightly adapt our C-translation (see `CHandler`) for loops to extract the correct branching condition for the witnesses
* Add methods to get the required infos (line, column, function) for the witness to `IBacktranslationValueProvider`

There are still some issues / open questions:

* Some of the added regression tests currently fail, because of bugs in the `RCFGBuilder`. They should however work properly, if the `IcfgBuilder` is used instead (i.e., once [this branch](https://github.com/ultimate-pa/ultimate/tree/wip/nk/icfgbuilder) is merged).
* Currently we don't generate any assumptions from our counterexamples. We could get them from our program states (there is already the class `ProgramStatePrinter` for this), but we currently only get states at locations that are not legal for assumptions according to the format. This might also change with the usage of `IcfgBuilder`.
* We currently cannot check the assumptions during validation (which happens in the trace abstraction), as this would require to translate them from C to SMT (which has to be done in the translation and not only in the trace abstraction). However, we currently try to validate at least the location of the assumptions. This might however be problematic if we use large block encoding (as no matching edge might exist for the C statement), therefore there is a boolean flag `CHECK_ASSUMPTION_LOCATIONS ` in `YamlWitnessProductAutomaton`. We need to figure out the best "mode" for validation (combination of block encoding and this flag), especially for the SV-COMP.
* Witness generation and validation is only properly tested for `ReachSafety`. However, I just tested generation and validation for simple examples for `NoOverflow` and `MemSafety`, and it seems to work in principle. This might require some further testing for SV-COMP.

For SV-COMP, we still need to add a [benchdef](https://gitlab.com/sosy-lab/sv-comp/bench-defs/-/tree/main/benchmark-defs) for violation witnesses 2.0, once this PR is merged.


trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/Waypoint.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+
+package de.uni_freiburg.informatik.ultimate.witnessparser.yaml;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Waypoints are the basic building blocks of entry-based violation witnesses. Technically a waypoint is a mapping with
+ * our keys: type, location, constraint and action.
```
maul-esel: "four keys"?

trunk/examples/witness-checking/regression/sequential/violation/assumption.c
```diff
@@ -0,0 +1,4 @@
+int main(){
+  int x = __VERIFIER_nondet_int();
+  if (x >= 0) reach_error();
+}
```
maul-esel: - Could you add an expected verdict (i.e. `//#False`) ?
- I guess it works for us even without declarations of these functions. Would be nicer though if they were present.

trunk/examples/witness-checking/regression/sequential/violation/assumption.c-witness.yml
```diff
@@ -0,0 +1,52 @@
...
+        constraint:
+          value: 'x >= 0'
+          format: c_expression
+        location:
+          file_name: C:\ultimate\trunk\examples\witness-checking\regression\violation\assumption.c
+          file_hash: UNUSED
+          line: 3
+          column: 3
+          function: main
+        action: follow
```
maul-esel, line 30: Ah interesting, so we do generate assumption waypoints sometimes? If this only the case if it is followed by a branching waypoint with the same condition?

trunk/examples/witness-checking/regression/sequential/violation/long_function_call.c
```diff
@@ -0,0 +1,14 @@
+int this_is_a_function_with_lots_of_arguments(int a,int b, int c, int d, int e,
```
maul-esel, line 1: :+1: :joy:

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
```diff
@@ -118,30 +127,37 @@ public boolean process(final IElement root) throws IOException {
...
 				throw new UnsupportedOperationException("two witness models");
 			}
-			mWitnessNode = (WitnessNode) root;
+			mLogger.warn(
+					"Found a witness automaton. I will only consider traces that are accepted by the witness automaton");
+			final INestedWordAutomaton<WitnessEdge, WitnessNode> witnessAutomaton =
+					new WitnessModelToAutomatonTransformer((WitnessNode) root, mServices).getResult();
+			mWitnessTransformer =
+					(automaton, predicateFactory) -> WitnessUtils.constructGraphMLWitnessProduct(mServices, automaton,
+							extendWitnessAutomaton(witnessAutomaton), predicateFactory, mLogger, Property.TERMINATION);
```
maul-esel, line 140: Does any witness format already support violation witnesses for termination? I.e. non-termination witnesses, which would require infinite paths? Or why do we have to handle witnesses in BüchiAutomizer?

maul-esel: And since the code above suggests GraphML witnesses, are these changes related to YAML witnesses, or are they separate fixes?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/CLocation.java
```diff
@@ -212,4 +212,12 @@ public String getFunction() {
 		}
 		return scope.getDeclarator().getName().toString();
 	}
+
+	/**
+	 * Returns a location for the parent node of this.
+	 */
+	public CLocation getParent() {
+		return new CLocation(mNode.getParent(), ignoreDuringBacktranslation(), mLineDirectiveMapping,
```
maul-esel, line 220: Is it sound to say the parent should be ignored iff this location should be ignored?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -3598,14 +3597,14 @@ private List<Statement> handleLoopCondition(final ILocation loc, final IDispatch
 				mExprResultTransformer.transformSwitchRexIntToBool(condResult, loc, cond);
 		final List<Statement> result = new ArrayList<>(condTransformed.getStatements());
 		resultBuilder.addDeclarations(condTransformed.getDeclarations());
-		// Insert an if-statement: if (!cond) break;
+		// Insert an if-statement: if (cond) {} else break;
+		// Note: we could invert the condition and ommit the then branch, but we want to keep the negation consistent in
```
maul-esel: `omit`

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -3598,14 +3597,14 @@ private List<Statement> handleLoopCondition(final ILocation loc, final IDispatch
 				mExprResultTransformer.transformSwitchRexIntToBool(condResult, loc, cond);
 		final List<Statement> result = new ArrayList<>(condTransformed.getStatements());
 		resultBuilder.addDeclarations(condTransformed.getDeclarations());
-		// Insert an if-statement: if (!cond) break;
+		// Insert an if-statement: if (cond) {} else break;
```
maul-esel, line 3637: Why is this change needed?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/CACSL2BoogieBacktranslator.java
```diff
@@ -316,8 +316,9 @@ private boolean checkTempHavoc(final AtomicTraceElement<BoogieASTNode> ate) {
 
 	private AtomicTraceElement<CACSLLocation> handleConditional(final AtomicTraceElement<BoogieASTNode> ate,
```
maul-esel: Should we maybe rename this to `handleLoopConditional` ? I was just surprised when I saw the method name and the unfinished backtranslation message below.

trunk/source/Library-BoogieAST/src/de/uni_freiburg/informatik/ultimate/boogie/BoogieBacktranslationValueProvider.java
```diff
@@ -31,6 +34,21 @@ public int getEndLineNumberFromStep(final BoogieASTNode step) {
 		return step.getLocation().getEndLine();
 	}
 
+	@Override
+	public int getLineNumberFromStep(final BoogieASTNode step, final EnumSet<AtomicTraceElement.StepInfo> stepInfo) {
+		// TODO: Do we want some different lines for Boogie?
```
maul-esel: Do we?

trunk/source/Library-BoogieAST/src/de/uni_freiburg/informatik/ultimate/boogie/BoogieBacktranslationValueProvider.java
```diff
@@ -31,6 +34,21 @@ public int getEndLineNumberFromStep(final BoogieASTNode step) {
...
+		return getStartLineNumberFromStep(step);
+	}
+
+	@Override
+	public int getColumnNumberFromStep(final BoogieASTNode step, final EnumSet<AtomicTraceElement.StepInfo> stepInfo) {
+		if (step.getLocation() == null) {
+			return -1;
+		}
+		// TODO: Do we want some different columns for Boogie?
```
maul-esel: Do we?

trunk/source/Library-BoogieAST/src/de/uni_freiburg/informatik/ultimate/boogie/BoogieBacktranslationValueProvider.java
```diff
@@ -31,6 +34,21 @@ public int getEndLineNumberFromStep(final BoogieASTNode step) {
...
+	@Override
+	public int getLineNumberFromStep(final BoogieASTNode step, final EnumSet<AtomicTraceElement.StepInfo> stepInfo) {
+		// TODO: Do we want some different lines for Boogie?
+		return getStartLineNumberFromStep(step);
+	}
+
+	@Override
+	public int getColumnNumberFromStep(final BoogieASTNode step, final EnumSet<AtomicTraceElement.StepInfo> stepInfo) {
+		if (step.getLocation() == null) {
+			return -1;
```
maul-esel, line 45: Does column `-1` have a special meaning? What is it?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/CegarLoopFactory.java
```diff
@@ -204,8 +199,8 @@ private NwaCegarLoop<L> createFiniteAutomataCegarLoop(final IUltimateServiceProv
 			return new CegarLoopSWBnonRecursive<>(name, abstraction, root, csToolkit, predicateFactory, mPrefs,
 					errorLocs, proofProducer, services, mTransitionClazz, stateFactoryForRefinement);
 		}
-		if ((FORCE_FINITE_AUTOMATA_FOR_SEQUENTIAL_PROGRAMS && !IcfgUtils.isConcurrent(root)) || witnessAutomaton != null
-				|| mPrefs.getAutomataTypeConcurrency() == Concurrency.FINITE_AUTOMATA) {
+		if ((FORCE_FINITE_AUTOMATA_FOR_SEQUENTIAL_PROGRAMS && !IcfgUtils.isConcurrent(root))
+				|| witnessTransformer != null || mPrefs.getAutomataTypeConcurrency() == Concurrency.FINITE_AUTOMATA) {
```
maul-esel: This means we ignore some settings when a witness is given, right? Should we perhaps output a warning?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/WitnessTransformer.java
```diff
@@ -0,0 +1,45 @@
...
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.IPredicate;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.PredicateFactory;
+
+/**
+ * Interface to transform a program automaton using a witness, commonly used for witnesses in GraphML and YAML.
+ *
+ * @author Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ */
+@FunctionalInterface
+public interface WitnessTransformer<L> {
```
maul-esel: By our convention, this should be `IWitnessTransformer`

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/witnesschecking/YamlWitnessProductAutomaton.java
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	@Override
+	public int size() {
+		return mAllProductStates.size();
+	}
+
+	@Override
+	public String sizeInformation() {
+		return mAbstraction.sizeInformation();
```
maul-esel: This is inconsistent with the implementation of `size()`

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/witnesschecking/YamlWitnessProductAutomaton.java
```diff
@@ -0,0 +1,281 @@
...
+		final ProductPredicate result = new ProductPredicate(predicate, waypointCounter, violationSequenceCounter);
+		mAllProductStates.add(result);
+		return result;
+	}
+
+	/**
+	 * IPredicate annotated with counters for the Violation Sequence and Segments of an entry-based violation witness
+	 */
+	private static final class ProductPredicate extends AnnotatedPredicate<IPredicate, Pair<Integer, Integer>>
+			implements ISLPredicate {
```
maul-esel: This would not work for concurrent programs, right?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/witnesschecking/YamlWitnessProductAutomaton.java
```diff
@@ -0,0 +1,281 @@
...
+		public ProductPredicate(final IPredicate underlying, final int segmentCounter,
+				final int violationSequenceCounter) {
+			super(underlying, new Pair<>(segmentCounter, violationSequenceCounter));
```
maul-esel, line 270: Would it be more convenient to store the violation sequence directly rather than this `violationSequenceCounter`?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/witnesschecking/YamlWitnessProductAutomaton.java
```diff
@@ -0,0 +1,281 @@
...
+		mAllProductStates.add(result);
+		return result;
+	}
+
+	/**
+	 * IPredicate annotated with counters for the Violation Sequence and Segments of an entry-based violation witness
+	 */
+	private static final class ProductPredicate extends AnnotatedPredicate<IPredicate, Pair<Integer, Integer>>
+			implements ISLPredicate {
+		public ProductPredicate(final IPredicate underlying, final int segmentCounter,
```
maul-esel: Given the cast in `getProgramPoint()`, the constructor should require an `ISLPredicate`.

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/yaml/Witness.java
```diff
@@ -53,7 +53,12 @@ public String toString() {
 	}
 
 	public boolean isCorrectnessWitness() {
-		// TODO: Check this, when we also support violation witnesses
+		// TODO: Consistency check with invariants
```
maul-esel: This means "fail if both invariants and violation sequences are present", right?

maul-esel (commented):
Looks good, thanks @HelenAnnaMeyer and @schuessf for adding this support!

I've left some smaller comments. I also remember in our discussion after the presentation, it was mentioned that some aspects of the semantics of violation witnesses were not entirely clear (though I forgot what it was). Should we take this up with the authors / at SV-COMP before merging this, or do you think it's better to work with the current implementation and, if necessary, make changes later on?

trunk/examples/witness-checking/regression/sequential/violation/assumption.c
schuessf: > Could you add an expected verdict (i.e. `//#False`) ?

What do you mean with this verdict? Currently all witness regression tests have an implicit verdict based on their folder, i.e., we assume that all witnesses are valid.

> I guess it works for us even without declarations of these functions. Would be nicer though if they were present.

Yes, I can change that.

trunk/examples/witness-checking/regression/sequential/violation/assumption.c-witness.yml
schuessf: Currently no (see`YamlViolationWitnessGenerator`), this is simply a manually wirtten witness to test the validation.

trunk/source/BuchiAutomizer/src/de/uni_freiburg/informatik/ultimate/plugins/generator/buchiautomizer/BuchiAutomizerObserver.java
schuessf: It seems like we support GraphML witnesses for non-termination, this code above is for the validation of such. I think using GraphML witnesses we can express infinite paths (using loops). However, YAML witnesses currently don't work for non-termination, as infinite paths cannot be expressed and there is no meaningful target.

This code was just slightly refacted to get the witness transformer already on a higher level (and not just in the CEGAR loop).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: Previosuly, we translated some loop `while (c) st` to `while (true) if (!c) break; st`. This led to inverted `ConditionAnnotation` and thus the validation of branching waypoints did not work as expected.
In the backtranslation`CACSL2BoogieBacktranslator` made already some attempt to invert the step infos again. However, this does not help us for the validation.

trunk/source/Library-BoogieAST/src/de/uni_freiburg/informatik/ultimate/boogie/BoogieBacktranslationValueProvider.java
schuessf: Probably not 😉

trunk/source/Library-BoogieAST/src/de/uni_freiburg/informatik/ultimate/boogie/BoogieBacktranslationValueProvider.java
schuessf: Probably not 😉

trunk/source/Library-BoogieAST/src/de/uni_freiburg/informatik/ultimate/boogie/BoogieBacktranslationValueProvider.java
schuessf: You are right, the implemented interface is missing documentation. I assume `-1` means that there is no line / column number. I just kept the same behavior as for the existing methods of this class.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/CACSL2BoogieBacktranslator.java
schuessf: I can do this on `dev`, since it is not connected to this PR.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/CegarLoopFactory.java
schuessf: I can do this on `dev`, since it is not connected to this PR.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/witnesschecking/YamlWitnessProductAutomaton.java
schuessf: Maybe slightly, but the full violation sequences are not necessary outside of this class and they might need much more memory than just storing some counters.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/witnesschecking/YamlWitnessProductAutomaton.java
schuessf: Currently no, because of the usage of `ISLPredicate`. Maybe this could be slightly adapted though.

trunk/examples/witness-checking/regression/sequential/violation/assumption.c
maul-esel: I meant a verdict for program correctness, which, in this folder, would always be `False`. Just in case these programs ever occur in any other context.

Though you raise an interesting point that I also meant to raise in our discussions of correctness witnesses: We should also have tests for invalid witnesses to see they are indeed rejected. Though I wouldn't write this verdict in the program (there may be multiple witnesses for a program, some valid and some invalid).

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/witnesschecking/YamlWitnessProductAutomaton.java
maul-esel: But you're just storing the pointer to the object (which must still exist, otherwise successor computation fails), so it's not saving any significant amount of memory. But it's not really important either way :shrug:

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: However, I just checked the [backtranslation tests](https://jenkins.sopranium.de/job/Ultimate/job/Ultimate%20Nightly/job/PR-680/3/testReport/de.uni_freiburg.informatik.ultimate.regressiontest.backtranslation/CBacktranslationRegressionTestSuite/) and some of them fail now because of this change. We might get some more program states in the counterexamples (which can be actually helpful to extract assumptions 😉). Therefore we should check, if this deviation is acceptable (maybe our block encoding is now slightly "worse"?) and adapt the tests in that case.

trunk/examples/witness-checking/regression/sequential/violation/assumption.c
schuessf: I agree, expected verdicts would be good and I suppose the best location for the verdict is in the witness. Maybe we can even have some test suite for witness generation and validation in one run (to actually test witness generation). Both ideas have to be integrated into `WitnessRegressionTestSuite`, but I guess that is nothing to be changed in this PR.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/CLocation.java
schuessf: I guess it depends on the usage: For witness generation, we are only interested in the line and column of the parent and there the location is not ignored (as witnesses are generated after the backtranslation; so ignored locations should be already filtered out). Alternatively, I could replace this method with two separate methods to get just the line or column of the parent location.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/witnesschecking/WitnessUtils.java
```diff
@@ -34,50 +36,63 @@
...
+	public static <LETTER extends IIcfgTransition<?>> IDoubleDeckerAutomaton<LETTER, IPredicate>
+			constructYamlWitnessProduct(final IUltimateServiceProvider services,
+					final INwaOutgoingLetterAndTransitionProvider<LETTER, IPredicate> abstraction,
+					final Witness witness, final PredicateFactory predicateFactory, final ILogger logger,
+					final Property property) throws AutomataOperationCanceledException {
+		final var result = new YamlWitnessProductAutomaton<>(abstraction, witness, predicateFactory);
+		logger.info(
+				"Constructing product of automaton with %d states and violation witness of the following lengths: %s",
+				abstraction.size(), witness.getEntries().stream().map(x -> ((ViolationSequence) x).getSegments().size())
+						.collect(Collectors.toList()));
```
bahnwaerter, line 95: There is some indentation issue. Is this the result of an unfavorable formatter setting?

trunk/source/WitnessParser/src/de/uni_freiburg/informatik/ultimate/witnessparser/YamlWitnessParser.java
```diff
@@ -96,11 +105,53 @@ private static Stream<WitnessEntry> parseWitnessEntry(final Map<String, Object>
 					(List<Map<String, Map<String, Object>>>) entry.get("content");
 			return content.stream().map(x -> parseInvariantSetEntry(x));
 		}
+		case "violation_sequence":
+			return Stream
+					.of(parseViolationSequence((List<Map<String, List<Map<String, Object>>>>) entry.get("content")));
```
bahnwaerter, line 110: Another example of an unfavorable formatter setting?

bahnwaerter (approved):
I quickly skimmed through the changes and they look good so far. I have also noted a few minor things. You could also consider removing personal data from the example witnesses. This affects in particular the encoding of the user name in the file path of the specified witness files in the metadata section of each witness file.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/witnesschecking/WitnessUtils.java
schuessf: Yes, but this might have been fixed now, this was before the Java 21 migration, where we also slightly adapted the formatter. I will just keep this for now, we will run the formatter anyway on the `dev` branch.

PR #683, opened 2024-10-08
Support more standard functions

schuessf:
This PR adds support for new standard functions that occur in SV-COMP benchmarks.
Most of these functions are simply overapproximated (if possible just the return value), as we cannot handle them properly (e.g. files, time, ...).


maul-esel (approved):
Looks great, thanks :+1:

PR #684, opened 2024-10-08
Fix translation of __builtin_*_overflow functions

schuessf:
In GNU C there are functions to check for overflows of arithmetic operations ([`__builtin_*_overflow`](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html)). These functions perform an arithmetic operation on the first two arguments (without producing any overflows), write the result to the third argument and return whether an overflow would occur using the normal arithmetic operation.
Currently, we simply overapproximate the return value of these functions. This is in general unsound as we do not consider that the third argument changes at all. Also this is quite imprecise, which led to 55 `unknown` results in SV-COMP 24.

Therefore this PR handles these functions properly. This is done in the following way:
* First we calculate the result of the arithmetic operation (without any overflows). For the integer translation this is the "usual" operations, while for the bitvector translation we choose a large enough bitvector to store the result (e.g. 64bit for the multiplication of two 32bit integers).
* Then we have to shrink this result to fit in the result type again, s.t. we can write it to the third argument (Note: This also uses the optimization to avoid "unnecessary" heap variables, if the third argument is of the form `&x`. Therefore this depends on #685).
* Then we return 1 iff the original result did fit in the result type.

For each of the three cases I added a new method to `ExpressionTranslation` accordingly, as these cases differ between the integer translation and the bitvector translation.


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
```diff
@@ -575,4 +575,23 @@ public abstract Pair<Expression, Expression> constructOverflowCheckForArithmetic
...
+
+	/**
+	 * Returns an expression to check whether the given expression with infinite precision fits in the resultType.
+	 */
+	public abstract Expression checkInRangeInfinitePrecision(ILocation loc, Expression expr, ASTType inputType,
+			CPrimitive resultType);
+
+	/**
+	 * Converts the given expression with infinite precision to the given type. This conversion should extract the
+	 * lowest bits that fit in type.
```
maul-esel, line 594: I didn't quite get (from https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html) what is supposed to happen here for signed types... is the sign bit always lost?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/IntegerTranslation.java
```diff
@@ -394,6 +401,25 @@ private Expression convertToIntegerType(final ILocation loc, final Expression op
...
+		// For unsigned types, we apply a wraparound (in the corresponding unsigned type) and shift the result
+		// accordingly.
+		final CPrimitive correspondingUnsignedType = mTypeSizes.getCorrespondingUnsignedType(type);
+		final Expression wrapped = applyWraparound(loc, correspondingUnsignedType, exp);
+		final Expression maxValue = constructLiteralForIntegerType(loc, type,
+				mTypeSizes.getMaxValueOfPrimitiveType(correspondingUnsignedType).add(BigInteger.ONE));
+		final Expression shiftedToType =
+				ExpressionFactory.newBinaryExpression(loc, Operator.ARITHMINUS, wrapped, maxValue);
+		return ExpressionFactory.constructIfThenElseExpression(loc,
+				checkInRangeInfinitePrecision(loc, wrapped, null, type), wrapped, shiftedToType);
```
maul-esel: I don't quite get what is happening here, but that may just be me :)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -408,27 +409,45 @@ private Map<String, IFunctionModelHandler> getFunctionModels() {
...
+				IASTBinaryExpression.op_multiply, new CPrimitive(CPrimitives.LONGLONG)));
+		fill(map, "__builtin_umul_overflow", (main, node, loc, name) -> handleBuiltinOverflow(main, node, loc, name,
+				IASTBinaryExpression.op_multiply, new CPrimitive(CPrimitives.UINT)));
+		fill(map, "__builtin_umull_overflow", (main, node, loc, name) -> handleBuiltinOverflow(main, node, loc, name,
+				IASTBinaryExpression.op_multiply, new CPrimitive(CPrimitives.ULONG)));
+		fill(map, "__builtin_umulll_overflow", (main, node, loc, name) -> handleBuiltinOverflow(main, node, loc, name,
+				IASTBinaryExpression.op_multiply, new CPrimitive(CPrimitives.ULONGLONG)));
 		fill(map, "__builtin_add_overflow_p", overapproximateGccOverflowCheck);
 		fill(map, "__builtin_sub_overflow_p", overapproximateGccOverflowCheck);
 		fill(map, "__builtin_mul_overflow_p", overapproximateGccOverflowCheck);
```
maul-esel, line 499: I guess these overapproximations are not unsound (no side-effect happens), but could we also handle them precisely? Or would that be more complicated?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -2759,6 +2778,42 @@ private List<ExpressionResult> handleFloatArguments(final IDispatcher main, fina
...
+		final ExpressionResult right = mExprResultTransformer.convertIfNecessary(loc,
+				mExprResultTransformer.transformDispatchDecaySwitchRexBoolToInt(main, loc, arguments[1]), resultType);
+		builder.addAllExceptLrValue(left, right);
+		// Apply the operator to the first two parameters with infinite precision (i.e. ignoring any wraparound or
+		// overflows), convert the result to the given type and write it to the third argument.
+		final Pair<Expression, ASTType> infinitePrecisionResult =
+				mExpressionTranslation.constructInfinitePrecisionOperation(loc, operator, left.getLrValue().getValue(),
+						right.getLrValue().getValue(), resultType);
+		final ASTType infinitePrecisionType = infinitePrecisionResult.getSecond();
+		final AuxVarInfo auxVar = mAuxVarInfoBuilder.constructAuxVarInfo(loc, infinitePrecisionType, AUXVAR.RETURNED);
```
maul-esel: Why `AUXVAR.RETURNED`? This is not the return value, right?

maul-esel: In fact, why is an auxvar needed here at all?

Wouldn't this prevent any automatic simplification of `inRange` e.g. when the arguments are constants? (admittedly, not a major concern, as it probably doesn't happen often)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -2759,6 +2778,42 @@ private List<ExpressionResult> handleFloatArguments(final IDispatcher main, fina
...
+		// Write the (converted) result of the operation to the third argument
+		builder.addAllExceptLrValue(mExprResultTransformer.dispatchPointerWrite(main, loc, arguments[2],
+				mExpressionTranslation.convertInfinitePrecisionExpression(loc, auxVar.getExp(), resultType)));
+		// If the infinite precision result fits in the given type, return 0 otherwise 1.
+		final Expression inRange = mExpressionTranslation.checkInRangeInfinitePrecision(loc, auxVar.getExp(),
+				infinitePrecisionType, resultType);
+		final CPrimitive boolType = new CPrimitive(CPrimitives.BOOL);
+		final Expression zero = mExpressionTranslation.constructLiteralForIntegerType(loc, boolType, BigInteger.ZERO);
+		final Expression one = mExpressionTranslation.constructLiteralForIntegerType(loc, boolType, BigInteger.ONE);
+		final Expression resultExpr = ExpressionFactory.constructIfThenElseExpression(loc, inRange, zero, one);
```
maul-esel, line 3047: Don't we have any existing support to convert from a Boogie `bool` to whatever we are using for the C type? Maybe this could be simplified.

maul-esel (approved):
Thanks for noticing and fixing this! I left some minor comments and questions, but overall :+1:

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: In principle they could be handled similarly. However, these functions accept any types. Therefore we would need to determine the type of the third first (and be careful about any promotions etc.). And `handleBuiltinOverflow` would need to be slightly refactored, as no value is written. Since these functions are currently not used in SV-COMP, I chose to keep the overapproximation for now.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
schuessf: Yes, unfortunately there is no documentation about the behavior. But I just ran GCC and it seems that the sign bit is indeed lost. Maybe this is even implementation defined like the cast into a smaller type (6.3.1.3.3) 😉

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/IntegerTranslation.java
schuessf: I also need to understand this again, but I will try to add some comments 😉

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
schuessf: Now I share the code between conversion of "normal" expressions and infinite precision expression to signed, as it behaves the same (at least in the implementation we choose).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/IntegerTranslation.java
schuessf: The added function `convertToSmallerSignedType` now contains the previous comment of `convertToIntegerType`, as well as an example for the conversion. I hope this helps to understand the conversion better.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
maul-esel: :+1: great, that should ensure consistency

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/IntegerTranslation.java
maul-esel: :+1: thanks for the explanation and example

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: I used `ExpressionResultTransformer::rexBoolToInt` in f2fe42e2177fdb23bc3c8b0384aead46615f19f7. However, this led to type problems and failing tests (see [here](https://jenkins.sopranium.de/job/Ultimate/job/Ultimate%20Nightly/job/PR-684/1/testReport/de.uni_freiburg.informatik.ultimate.regressiontest.generic/RegressionTestSuite/I_programs_regression_c_builtin_sadd_overflow_c_S_programs_regression_c_AutomizerC_BitvectorTranslation_epf_T_programs_regression_c_AutomizerC_xml/)). The problem is that the literals by that [function](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java#L555) always have type `int`, whereas we want `_Bool` literals here. Therefore I will revert this change again.

PR #685, opened 2024-10-08
Fix translation of atomic functions

schuessf:
PR #657 added support for atomic functions. This had however two problems (in the current version):
* To improve our performance, this PR tried to avoid unecessary heap variables, i.e., we tried to translate `__atomic_store_n(&x, 1, 5)` to `x := 1` inside an atomic block rather than `write~(x, 1)` (see d08b25f and in the [PR](https://github.com/ultimate-pa/ultimate/pull/657#issuecomment-1883310446)). This way the arguments were dispatched possibly multiple times, which is problematic if they have side-effects (see [atomic_load_n_sideeffects.c](https://github.com/ultimate-pa/ultimate/compare/dev...wip/fs/atomic-fix#diff-d25c8f52ac3262e385e626766bb1204bcdc757a393dfc39c4f9ae3903c818526)).
* We put everything inside an atomic-block. However, not everything is guaranteed to be executed atomically, e.g. for `__atomic_store_n(&x, expr1, expr2)` only the actual write on `x` is atomic, not the evaluation of `expr1` or `expr2`. This could have led to missed data-races (see [atomic_memory_order.c](https://github.com/ultimate-pa/ultimate/compare/dev...wip/fs/atomic-fix#diff-6ea6a44548a6065be56965bfb6edc8167a6833a1a439bc060a9694b709734be7))
* We considered the read and write to be atomic for `__atomic_load`, but only the read should be atomic (fixed in f2e474d)

~~To fix these issues, the PR adds an interface `IPointerReadWrite` (and two implementations) that allows to get the dispatched argument, a read and a write separately (s.t. we can "control" the atomicity) and that guarantees that the dispatch is only done once. This interface is the used in `StandardFunctionHandler` for the translation of atomic functions.~~

To fix these issues, the PR adds three separate method to `ExpressionResultTransformer`:
* `dispatchPointerLValue` to dispatch the pointer expression, but tries to produce `LocalLValue`
* `makePointerAssignment` writes to the given address, produces simple assignment for `LocalLValue` and write-call for `HeapLValue`
* `readPointerValue` reads from the given address, always returns an aux-var as `RValue`, assigns the `LocalLValue` to the aux-var or reads to the aux-var from the memory for `HeapLValue`

### DIFF
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java
@@ -34,7 +34,6 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
-import java.util.function.Function;
 
 import org.eclipse.cdt.core.dom.ast.IASTBinaryExpression;
 import org.eclipse.cdt.core.dom.ast.IASTInitializerClause;
@@ -47,7 +46,6 @@ import de.uni_freiburg.informatik.ultimate.boogie.ast.ArrayLHS;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.BinaryExpression;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.Declaration;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.Expression;
-import de.uni_freiburg.informatik.ultimate.boogie.ast.LeftHandSide;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.Statement;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.StructConstructor;
 import de.uni_freiburg.informatik.ultimate.boogie.type.BoogieType;
@@ -1041,8 +1039,9 @@ public class ExpressionResultTransformer {
 	}
 
 	/**
-	 * Method to dispatch the read/write of {@code *pointer}. Uses an optimization that if pointer is of the form
-	 * {@code &x}, we try to avoid putting x on the heap and dispatch {@code x} instead.
+	 * Dispatches a pointer and ensures that the result contains either a {@code LocalLValue} or a {@code HeapLValue}.
+	 * If possible (i.e., pointer is of the form {@code &x} where {@code x} is not already on the heap), a
+	 * {@code LocalLValue} is returned as an optimization.
 	 *
 	 * @param main
 	 *            Dispatcher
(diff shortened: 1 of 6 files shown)


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java
```diff
@@ -1147,4 +1076,104 @@ ExpressionResult apply(final ExpressionResultTransformer ert, final ExpressionRe
 				final CType targetCType, final ILocation loc, final IASTNode hook);
 	}
 
+	public interface IPointerReadWrite {
```
maul-esel: Please document this interface and its methods. It is not clear to me what each of them is supposed to do.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java
```diff
@@ -1147,4 +1076,104 @@ ExpressionResult apply(final ExpressionResultTransformer ert, final ExpressionRe
...
+		ExpressionResult getRead();
+
+		ExpressionResult getWrite(final Function<CType, ExpressionResult> valueProvider);
+
+		default ExpressionResult getWrite(final Expression value) {
+			return getWrite(type -> new ExpressionResult(new RValue(value, type)));
+		}
+	}
+
+	private class NormalPointerReadWrite implements IPointerReadWrite {
```
maul-esel: Can we rename this? I think it is not clear what "normal" means here (at least, I don't know). Also, even after renaming, please document the class.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java
```diff
@@ -1147,4 +1076,104 @@ ExpressionResult apply(final ExpressionResultTransformer ert, final ExpressionRe
...
+			builder.addStatements(
+					mMemoryHandler.getWriteCall(mLoc, heapValue, value.getLrValue().getValue(), valueType, false));
+			if (mDataRaceChecker != null) {
+				mDataRaceChecker.checkOnWrite(builder, mLoc, heapValue);
+			}
+			return builder.build();
+		}
+	}
+
+	private class AddressofReadWrite implements IPointerReadWrite {
```
maul-esel: (nitpick) `AddressOfReadWrite`

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java
maul-esel: Also, just for my understanding: Could you explain how the two classes implementing this interface differ from `HeapLValue` vs `LocalLValue` ? Do we need both distinctions?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java
maul-esel: Also, please document the class.

maul-esel (approved):
Thanks for making sure these functions work correctly!

I took a look, and it seems a good change. I was not able to check every atomic function in precise detail though. I left some comments on the general mechanism, mostly asking for documentation / explanation.

schuessf:
> Also, just for my understanding: Could you explain how the two classes implementing this interface differ from `HeapLValue` vs `LocalLValue` ? Do we need both distinctions?

@maul-esel This interface was not really necessary,  `HeapLValue` and `LocalLValue`  are sufficient for distinction.
Therefore I replaced this interface with separate methods and also added some documentation.
So please review this PR again.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java
```diff
@@ -1041,110 +1039,112 @@ public ExpressionResult convertNullPointerConstantToPointer(final ExpressionResu
...
-					disp.getCType(), null);
-			mDataRaceChecker.checkOnRead(builder, loc, heapValue);
+		// Otherwise simply dispatch the expression, but make sure that the result contains a HeapLValue.
+		final ExpressionResult result = decayArrayToPointer((ExpressionResult) main.dispatch(pointer), loc, pointer);
+		if (result.getLrValue() instanceof HeapLValue) {
+			return result;
 		}
+		final ExpressionResultBuilder builder = new ExpressionResultBuilder(result);
+		builder.resetLrValue(LRValueFactory.constructHeapLValue(mTypeHandler, result.getLrValue().getValue(),
+				result.getCType(), null));
```
maul-esel, line 1075: Out of interest: In which cases could it happen that we do not have a `HeapLValue` here, and what do we have instead?

maul-esel (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ExpressionResultTransformer.java
schuessf: This happens for example if the expression is of the form `&x`, where `x` is already on the heap (we get an `RValue` in that case) or if the expression is some variable `p` of pointer type (we get a `LocalLValue` in that case). Actually, I did not find a case, where we get a `HeapLValue` here, but I will keep that case anyway (`result.getLrValue().getValue()` would throw an exception, if this happens).

PR #687, opened 2024-10-16
Backtranslation of contracts with handling of "modifies" clause

maul-esel:
This PR changes the backtranslation of procedure contracts.
- When backtranslating contracts from Boogie (which has `modifies` clauses to encode frame conditions) to C/ACSL (where our contracts currently do not have `modifies` or other features for frame conditions), we add conjuncts of the form `x == \old(x)` for variables that are not modifiable by the procedure.
- To support this, we add a specialized method for contract backtranslation to the `ITranslator` interface.

### DIFF
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
@@ -32,7 +32,10 @@ import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
 import java.util.function.Consumer;
+import java.util.stream.Collectors;
 
 import org.eclipse.cdt.core.dom.ast.IASTFunctionDefinition;
 
@@ -88,6 +91,53 @@ public final class Boogie2ACSL {
 	public BacktranslatedExpression translateExpression(
 			final de.uni_freiburg.informatik.ultimate.boogie.ast.Expression expression, final ILocation context) {
 		return translateExpression(expression, context, false);
+	}
+
+	/**
+	 * Provides a specialized translation for the 'ensures' clause in a procedure contract. Specifically, it adds
+	 * conjuncts of the form 'x == \old(x)' for all global variables that are not modifiable by the procedure, and that
+	 * are off-heap (because we assume conservatively that the procedure may modify the heap).
+	 *
+	 * @param expression
+	 *            The original 'ensures' clause to translate
+	 * @param context
+	 *            The context, i.e., the location of the C function definition
+	 * @param modifiableGlobals
+	 *            The set of variables listed in the Boogie procedure "modifies" clause
+	 * @return A backtranslated ensures clause with the additional conjuncts
+	 */
+	public BacktranslatedExpression translateEnsuresExpression(
+			final de.uni_freiburg.informatik.ultimate.boogie.ast.Expression expression, final ILocation context,
+			final Set<de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression> modifiableGlobals) {
+		final var intType = new CPrimitive(CPrimitives.INT);
+		final var boolRange = BigInterval.booleanRange();
+
+		if (expression == null) {
+			return computeOldVarEqualities(modifiableGlobals)
+					.map(e -> new BacktranslatedExpression(e, intType, boolRange)).orElse(null);
+		}
+
+		final var mainExpression = translateExpression(expression, context);
+		final var oldVarEqualities = computeOldVarEqualities(modifiableGlobals);
+		if (oldVarEqualities.isPresent()) {
+			return new BacktranslatedExpression(
+					new BinaryExpression(Operator.LOGICAND, mainExpression.getExpression(), oldVarEqualities.get()),
+					new CPrimitive(CPrimitives.INT), BigInterval.booleanRange());
+		}
+
+		return mainExpression;
+	}
+
+	private Optional<Expression> computeOldVarEqualities(
+			final Set<de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression> modifiableGlobals) {
+		final var modifiableNames = modifiableGlobals.stream().map(e -> e.getIdentifier()).collect(Collectors.toSet());
+		return mSymbolTable.getGlobalScope().entrySet().stream()
+				.filter(e -> !modifiableNames.contains(e.getValue().getBoogieName()))
+				.filter(e -> !mMapping.getVar(e.getValue().getBoogieName(), e.getValue().getDeclarationInformation())
+						.getThird())
+				.map(e -> new IdentifierExpression(e.getKey()))
+				.<Expression> map(v -> new BinaryExpression(Operator.COMPEQ, v, new OldValueExpression(v)))
+				.reduce((eq1, eq2) -> new BinaryExpression(Operator.LOGICAND, eq1, eq2));
 	}
 
 	private BacktranslatedExpression translateExpression(
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
@@ -111,15 +111,18 @@ public final class Boogie2ACSL {
 	public BacktranslatedExpression translateEnsuresExpression(
 			final de.uni_freiburg.informatik.ultimate.boogie.ast.Expression expression, final ILocation context,
 			final Set<de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression> modifiableGlobals) {
+		final String cFun =
+				((IASTFunctionDefinition) ((CLocation) context).getNode()).getDeclarator().getName().toString();
+
 		final var mainExpression = expression == null ? null : translateExpression(expression, context);
 		if (mainExpression == null) {
 			final var intType = new CPrimitive(CPrimitives.INT);
 			final var boolRange = BigInterval.booleanRange();
-			return computeOldVarEqualities(modifiableGlobals)
+			return computeOldVarEqualities(cFun, modifiableGlobals)
 					.map(e -> new BacktranslatedExpression(e, intType, boolRange)).orElse(null);
 		}
 
-		final var oldVarEqualities = computeOldVarEqualities(modifiableGlobals);
+		final var oldVarEqualities = computeOldVarEqualities(cFun, modifiableGlobals);
 		if (oldVarEqualities.isPresent()) {
 			return new BacktranslatedExpression(
 					new BinaryExpression(Operator.LOGICAND, mainExpression.getExpression(), oldVarEqualities.get()),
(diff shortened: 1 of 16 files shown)


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -89,6 +95,85 @@ public BacktranslatedExpression translateExpression(
...
+
+		final var boogieName = stv.getBoogieName();
+		if (modifiableNames.contains(boogieName)) {
+			// No equalities for modifiable variables.
+			return Optional.empty();
+		}
+
+		if (!mMapping.hasVar(boogieName, stv.getDeclarationInformation())) {
+			// We omit the conjunct if the variable cannot be backtranslated and output a warning (sound but imprecise).
+			mReporter.accept("Unknown variable: " + boogieName);
```
schuessf, line 161: Can and should we check here, if this variable is just some auxiliary variable (`#length`, `#valid`, `#race`) and don't report a warning in that case?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -89,6 +95,85 @@ public BacktranslatedExpression translateExpression(
...
+		if (expression == null) {
+			return computeOldVarEqualities(modifiableGlobals)
+					.map(e -> new BacktranslatedExpression(e, intType, boolRange)).orElse(null);
+		}
+
+		final var mainExpression = translateExpression(expression, context);
+		final var oldVarEqualities = computeOldVarEqualities(modifiableGlobals);
+		if (oldVarEqualities.isPresent()) {
+			return new BacktranslatedExpression(
+					new BinaryExpression(Operator.LOGICAND, mainExpression.getExpression(), oldVarEqualities.get()),
```
schuessf, line 128: Check first if `mainExpression` is `null`, i.e., could not be backtranslated, otherwise this can lead to a NPE (I just experienced one).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -89,6 +95,85 @@ public BacktranslatedExpression translateExpression(
...
+
+		final var info = mMapping.getVar(boogieName, stv.getDeclarationInformation());
+		final boolean isOnHeap = info.getThird();
+		if (isOnHeap) {
+			// We cannot say whether an on-heap variable is modified, so we omit the conjunct (sound but imprecise).
+			//
+			// TODO We could check if the procedure can modify the heap. To do so, we would need the Boogie names of the
+			// heap data arrays.
+			mReporter.accept("Cannot encode non-modifiability of on-heap variable " + cId);
+			return Optional.empty();
```
schuessf, line 170: Can you explain in which case this branch is entered?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -139,7 +224,9 @@ private BacktranslatedExpression translateIdentifierExpression(
 			final var range = getRangeForCType(type);
 			return new BacktranslatedExpression(new ACSLResultExpression(), type, range);
 		} else if (mMapping.hasVar(boogieId, expr.getDeclarationInformation())) {
-			final Pair<String, CType> pair = mMapping.getVar(boogieId, expr.getDeclarationInformation());
+			final Triple<String, CType, Boolean> pair = mMapping.getVar(boogieId, expr.getDeclarationInformation());
+			// TODO Should we do something different if the variable is on-heap?
```
schuessf: Should we?

schuessf (commented):
Looks good, just a few questions and comments

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
maul-esel: We don't have such a check for the normal backtranslation of expressions, right?

I expect that this check is mostly triggered by such variables.  The point of the warning is to indicate potential problems, e.g. for inductiveness of the contract. Unless we can soundly rule out that the contract becomes non-inductive due to some specific variable, I think the warning is justified.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
maul-esel: :+1: good catch, done in efe74121b

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
maul-esel: This branch is entered if some global variable `z` is on the heap. It is then in general unsound to claim that `z == old(z)`, because the procedure may modify the value of `z` through some pointer. Since, at the same time, we cannot express the effect of the procedure on the heap (yet), we omit any information on `z` and output a warning (possibly non-inductive contract).

In heap-manipulating programs, this warning will occur rather often. I don't see how it can be generally avoided without a complex alias analysis (MemorySlicer & co). But a possible improvement would be to see if the procedure is at all able to modify the heap. For that we would need access to the heap data arrays -- since you have worked in this backtranslation a lot, do you happen to know if we can easily get them here?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
maul-esel: Good question :grin: If some variable `z` is on-heap, the corresponding Boogie variable `~#z~0` actually represents the pointer to `z`, right? So perhaps we should backtranslate it to `&z` ? Or am I missing something?

schuessf: > If some variable `z` is on-heap, the corresponding Boogie variable `~#z~0` actually represents the pointer to `z`, right?

Yes, I think so.

> So perhaps we should backtranslate it to `&z` ? Or am I missing something?

We could do it like that, but this would also influence all of our backtranslation tests that use pointers.
I would rather propose to keep it for now and handle this in `translateArrayAccess` when we support pointer dereferences there.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
schuessf: Thank you, this makes sense.

>For that we would need access to the heap data arrays -- since you have worked in this backtranslation a lot, do you happen to know if we can easily get them here?

I think it is safe to assume that every heap array starts with `SFO.MEMORY`, so could simply check if any of the `modifiableNames` matches this name.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
maul-esel: Ok, to put it another way: While the current code is slightly incorrect, it does not currently lead to a bug because we can't backtranslate expressions with pointers anyway. Is that the gist of it?

Then I would leave the translation as-is, but update the TODO accordingly.

I briefly thought about adding `\old(&z) == &z` for on-heap global variables `z` where `~#z~0` is unmodifiable, but I guess that should be an ACSL tautology, so there's not really a point :wink:

schuessf: > Ok, to put it another way: While the current code is slightly incorrect, it does not currently lead to a bug because we can't backtranslate expressions with pointers anyway. Is that the gist of it?

Yes, kind of. Another point is that pointers are boogie structs, so the variable should not occur standalone in the expression, but only as a struct access.

> Then I would leave the translation as-is, but update the TODO accordingly.

👍 

> I briefly thought about adding `\old(&z) == &z` for on-heap global variables `z` where `~#z~0` is unmodifiable, but I guess that should be an ACSL tautology, so there's not really a point 😉

Not to be confused with `&\old(z) == &z`, but we probably don't want this in the contract either (for now) 😄

maul-esel: On a related note (but not something for this PR): It is sometimes confusing to read `z={3:0}` in counterexamples after an assignment `z = 10;`. Writing `&z={3:0}` might make things less confusing?

schuessf: Yes, this might be a good idea.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
maul-esel: Oh, neat. Implemented in 240548f.

schuessf (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
maul-esel: > `&\old(z)`

Oh god, that looks evil :smiling_imp: Is that even allowed -- taking the address of an old value expression?

maul-esel: > Another point is that pointers are boogie structs, so the variable should not occur standalone in the expression, but only as a struct access.

That is something to be handled in the backtranslator for `BoogiePreprocessor`. Currently we still have `IdentifierExpression`s with identifier `~#z~0!base` etc.

PR #688, opened 2024-10-16
Support atomic types

schuessf:
This PR adds support for [atomic types](https://en.cppreference.com/w/c/language/atomic) (`_Atomic`). To do so, I performed the following steps:
* First the keyword is replaced by some artificial GCC attribute (`__attribute__((atomic))`) for our parser not to crash.
* Whether we have an atomic type is then stored in the `CType`. This flag is only set explicitly in `CPrimitive` (based on the GCC attribute) and a `CNamed` is atomic iff the underlying type is atomic. All other types are currently not atomic. There can be actually an atomic pointer (see [here](https://stackoverflow.com/questions/42082219/declaring-atomic-pointers-vs-pointers-to-atomics)), but we cannot parse the attribute properly at this location and thus don't support atomic pointers.
* We don't check for data-races on atomic types (see `DataRaceChecker`).
* We have to ensure that all operations on atomic variables are actually performed atomically. Therefore the translation of reads (see `MemoryHandler::getReadCall`), writes (see `MemoryHandler::getWriteCall` and `CHandler::makeAssignment`) and combined read-writes (e.g. `++`, `--`, `*=`, ...) (see `CHandler::makeAtomicAssignmentIfNecessary`) was adapted.

### DIFF
--- a/trunk/examples/concurrent/pthreads/races/regression/atomic_type_array.c
+++ b/trunk/examples/concurrent/pthreads/races/regression/atomic_type_array.c
@@ -1,0 +1,24 @@
+//#Safe
+
+/*
+ * Author: Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ * Date: 2024-10-18
+ */
+
+typedef unsigned long pthread_t;
+
+_Atomic int x[10];
+
+void* thread1() {
+  int local = x[7];
+}
+
+void* thread2() {
+  x[7] = 42;
+}
+
+int main() {
+  pthread_t t1, t2;
+  pthread_create(&t1, 0, thread1, 0);
+  pthread_create(&t2, 0, thread2, 0);
+}
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
@@ -130,6 +130,7 @@ import de.uni_freiburg.informatik.ultimate.boogie.annotation.LTLPropertyCheck;
 import de.uni_freiburg.informatik.ultimate.boogie.annotation.LTLPropertyCheck.CheckableExpression;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.ASTType;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.AssignmentStatement;
+import de.uni_freiburg.informatik.ultimate.boogie.ast.AtomicStatement;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.Attribute;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.Axiom;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.BinaryExpression;
@@ -835,7 +836,11 @@ public class CHandler {
 			final ExpressionResult rr = mExprResultTransformer.transformSwitchRexBoolToInt(rightOperand, loc, node);
 			final ExpressionResult result =
 					mCExpressionTranslator.handleMultiplicativeOperation(loc, node.getOperator(), rl, rr);
-			return makeAssignment(loc, leftOperand.getLrValue(), Collections.emptyList(), result, node);
+			// Make sure that the evaluation of the operands is not inside an atomic block, but the read of the left
+			// operand (i.e., the potential of a heap variable) is.
+			final List<Statement> statementsBeforeRead =
+					DataStructureUtils.concat(leftOperand.getStatements(), rr.getStatements());
+			return handleAtomicReadWrite(loc, leftOperand.getLrValue(), statementsBeforeRead, result, node);
 		}
 		case IASTBinaryExpression.op_plus:
 		case IASTBinaryExpression.op_minus: {
@@ -859,7 +864,11 @@ public class CHandler {
 					mExprResultTransformer.transformDecaySwitchRexBoolToInt(rightOperand, loc, node);
 			final ExpressionResult result =
 					mCExpressionTranslator.handleAdditiveOperation(loc, node.getOperator(), rl, rr);
-			return makeAssignment(loc, leftOperand.getLrValue(), Collections.emptyList(), result, node);
+			// Make sure that the evaluation of the operands is not inside an atomic block, but the read of the left
+			// operand (i.e., the potential of a heap variable) is.
+			final List<Statement> statementsBeforeRead =
+					DataStructureUtils.concat(leftOperand.getStatements(), rr.getStatements());
+			return handleAtomicReadWrite(loc, leftOperand.getLrValue(), statementsBeforeRead, result, node);
 		}
 		case IASTBinaryExpression.op_binaryAnd:
 		case IASTBinaryExpression.op_binaryOr:
@@ -875,7 +884,11 @@ public class CHandler {
 			final ExpressionResult rr = mExprResultTransformer.transformSwitchRexBoolToInt(rightOperand, loc, node);
 			final ExpressionResult result =
 					mCExpressionTranslator.handleBitwiseArithmeticOperation(loc, node.getOperator(), rl, rr);
-			return makeAssignment(loc, leftOperand.getLrValue(), Collections.emptyList(), result, node);
+			// Make sure that the evaluation of the operands is not inside an atomic block, but the read of the left
+			// operand (i.e., the potential of a heap variable) is.
+			final List<Statement> statementsBeforeRead =
+					DataStructureUtils.concat(leftOperand.getStatements(), rr.getStatements());
+			return handleAtomicReadWrite(loc, leftOperand.getLrValue(), statementsBeforeRead, result, node);
 		}
 		case IASTBinaryExpression.op_shiftLeft:
 		case IASTBinaryExpression.op_shiftRight: {
@@ -890,7 +903,11 @@ public class CHandler {
 			final ExpressionResult rr = mExprResultTransformer.transformSwitchRexBoolToInt(rightOperand, loc, node);
 			final ExpressionResult result =
 					mCExpressionTranslator.handleBitshiftOperation(loc, node.getOperator(), rl, rr);
-			return makeAssignment(loc, leftOperand.getLrValue(), Collections.emptyList(), result, node);
+			// Make sure that the evaluation of the operands is not inside an atomic block, but the read of the left
+			// operand (i.e., the potential of a heap variable) is.
+			final List<Statement> statementsBeforeRead =
+					DataStructureUtils.concat(leftOperand.getStatements(), rr.getStatements());
+			return handleAtomicReadWrite(loc, leftOperand.getLrValue(), statementsBeforeRead, result, node);
 		}
 		default:
 			final String msg = "Unknown or unsupported unary operation";
@@ -2549,12 +2566,12 @@ public class CHandler {
 		case IASTUnaryExpression.op_postFixIncr:
 		case IASTUnaryExpression.op_postFixDecr: {
 			return mCExpressionTranslator.handlePostfixIncrementAndDecrement(loc, node.getOperator(), operand, node,
-					a -> makeAssignment(loc, operand.getLrValue(), Collections.emptyList(), a, node));
+					a -> handleAtomicReadWrite(loc, operand.getLrValue(), operand.getStatements(), a, node));
 		}
 		case IASTUnaryExpression.op_prefixDecr:
 		case IASTUnaryExpression.op_prefixIncr: {
 			return mCExpressionTranslator.handlePrefixIncrementAndDecrement(node.getOperator(), loc, operand, node,
-					a -> makeAssignment(loc, operand.getLrValue(), Collections.emptyList(), a, node));
+					a -> handleAtomicReadWrite(loc, operand.getLrValue(), operand.getStatements(), a, node));
 		}
 		case IASTUnaryExpression.op_bracketedPrimary:
 			return operand;
@@ -2713,6 +2730,28 @@ public class CHandler {
 		}
 	}
 
+	private ExpressionResult handleAtomicReadWrite(final ILocation loc, final LRValue leftHandSide,
+			final List<Statement> statementsBeforeRead, final ExpressionResult rhs, final IASTNode hook) {
+		final ExpressionResult assignment = makeAssignment(loc, leftHandSide, Set.of(), rhs, hook);
+		if (!leftHandSide.getCType().isAtomic()) {
+			// For non-atomic types return the normal assignment
+			return assignment;
+		}
+		final List<Statement> allStatements = assignment.getStatements();
+
+		// Check if statementsBeforeRead are a prefix of allStatements
+		if (!statementsBeforeRead.equals(allStatements.subList(0, statementsBeforeRead.size()))) {
+			throw new AssertionError(
+					"Unexpected result of makeAssignment: statements do not start with statements of rhs");
+		}
+		// For atomic types make sure that all statement except the ones before the read are in an atomic block
+		final List<Statement> atomicStatements =
+				allStatements.subList(statementsBeforeRead.size(), allStatements.size());
+		return new ExpressionResultBuilder().addStatements(statementsBeforeRead)
+				.addAllExceptLrValueAndStatements(assignment).setLrValue(assignment.getLrValue())
+				.addStatement(StatementFactory.constructAtomicStatement(loc, atomicStatements)).build();
+	}
+
 	/**
 	 *
 	 * @param loc
@@ -2800,6 +2839,7 @@ public class CHandler {
 				builder.addStatement(assignment);
 				resultRhs = auxVar.getExp();
 			}
+			// MemoryHandler::getWriteCall already handles atomic types properly, so there is nothing to do here.
 			builder.addStatements(mMemoryHandler.getWriteCall(loc, hlv, resultRhs,
 					rightHandSideValueWithConversionsApplied.getCType(), false));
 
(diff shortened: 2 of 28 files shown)


maul-esel:
> We have to ensure that all operations on atomic variables are actually performed atomically. Our translation seems to ensure this for read and write operations (I am not 100% sure though).

I think we should ensure this by adding atomic statements in Boogie.

This would also solve the current unsoundness on this branch when using the setting `--rcfgbuilder.only.consider.context.switches.at.boundaries.of.atomic.blocks true` aka the "assume no data race"-LBE.

schuessf:
> I think we should ensure this by adding atomic statements in Boogie.
> 
> This would also solve the current unsoundness on this branch when using the setting `--rcfgbuilder.only.consider.context.switches.at.boundaries.of.atomic.blocks true` aka the "assume no data race"-LBE.

Thanks, that's a good point. With this settings, it actually makes a difference, whether we translate to a single Boogie statement `st` or `atomic { st }`.
So I guess, the only C expressions where the translation neeeds to be adapted for atomic types, are assignments (also sth. like `+=`) and reads (probably only for variables on the heap).

trunk/examples/concurrent/pthreads/races/regression/atomic_type_array.c
```diff
@@ -0,0 +1,24 @@
+//#Safe
+
+/*
+ * Author: Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ * Date: 2024-10-18
+ */
+
+typedef unsigned long pthread_t;
+
+_Atomic int x[10];
```
maul-esel, line 10: Is this an atomic array of integers, or an array of atomic integers? And do we support this usage?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CExpressionTranslator.java
```diff
@@ -588,7 +589,14 @@ public Result handlePostfixIncrementAndDecrement(final ILocation loc, final int
 		builder.setOrResetLrValue(new RValue(valueXcremented, oType, false, false));
 		final ExpressionResult assign = funMakeAssignment.apply(builder.build());
 
-		return new ExpressionResultBuilder().addAllExceptLrValue(assign).setLrValue(tmpRValue).build();
+		final ExpressionResultBuilder resultBuilder = new ExpressionResultBuilder().setLrValue(tmpRValue);
+		if (oType.isAtomic()) {
+			final AtomicStatement atomic = new AtomicStatement(loc, assign.getStatements().toArray(Statement[]::new));
```
maul-esel: The atomic block contains all statements inside the original `exprRes`. Could this not be unsound? Only the assignment of `auxvar` and the assignment created by `funMakeAssignment` should be atomic.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CExpressionTranslator.java
```diff
@@ -635,7 +643,14 @@ public Result handlePrefixIncrementAndDecrement(final int prefixOp, final ILocat
 		builder.setLrValue(new RValue(valueXcremented, oType, false, false));
 		final ExpressionResult assign = funMakeAssignment.apply(builder.build());
 		final RValue tmpRValue = new RValue(auxvar.getExp(), oType);
-		return new ExpressionResultBuilder().addAllExceptLrValue(assign).setLrValue(tmpRValue).build();
+		final ExpressionResultBuilder resultBuilder = new ExpressionResultBuilder().setLrValue(tmpRValue);
+		if (oType.isAtomic()) {
+			final AtomicStatement atomic = new AtomicStatement(loc, assign.getStatements().toArray(Statement[]::new));
```
maul-esel: Similar to the method above: the atomic block might contain too many statements.

trunk/examples/concurrent/pthreads/races/regression/atomic_type_array.c
maul-esel: Hmm, sentence 3 in section 6.7.3 of the C standard (N1570) says

> The type modified by the _Atomic qualifier shall not be an array type or a function type.

Is this an issue here?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/DataRaceChecker.java
```diff
@@ -260,6 +260,10 @@ private static ImmutableList<String> getAccessPath(final LeftHandSide lhs) {
 	}
 
 	private static boolean isRaceImpossible(final LRValue lrVal) {
+		if (lrVal.getCType().isAtomic()) {
+			// Atomic types cannot lead to data races
+			return true;
+		}
```
maul-esel, line 266: Is it ever possible (and defined behaviour) to access the same memory location once atomically and once non-atomically? If so, this would probably constitute a data race and we should add checks here. If all accesses are atomic, then these checks hold trivially.

E.g. is something like the code below allowed?
```c
atomic_int x;

void* thread(void* ptr) {
  int* intptr = (int*) ptr;
  // I don't know if accessing x via intptr is allowed, e.g. because sizeof(atomic_int) and sizeof(int) may differ (6.2.4/27)
  *intptr = 7; // non-atomic (?) access to x
  return ptr;
}

int main() {
  pthread_t tid;
  pthread_create(&tid, NULL, thread, &x);
  int y = x; // atomic access to x, but RACE with non-atomic access
  return y;
}
```

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -2844,7 +2844,11 @@ public ExpressionResult makeAssignment(final ILocation loc, final LRValue leftHa
 		final AssignmentStatement assignStmt = StatementFactory.constructAssignmentStatement(loc,
 				new LeftHandSide[] { lValue.getLhs() }, new Expression[] { rhsWithBitfieldTreatment });
 
-		builder.addStatement(assignStmt);
+		if (leftHandSide.getCType().isAtomic()) {
+			builder.addStatement(new AtomicStatement(loc, new Statement[] { assignStmt }));
+		} else {
+			builder.addStatement(assignStmt);
+		}
```
maul-esel, line 2888: Note that compound assignment expressions such as `x += 3` also have to be treated atomically, as per paragraph 3 in Section 6.5.12.2 of the C standard (N1570):

> ... If E1 has an atomic type, compound assignment is a read-modify-write operation with memory_order_seq_cst memory order semantics.

maul-esel (commented):
:+1: Thanks for tackling this!

I've left some questions, and we should in particular also support other atomic operations such as `+=` etc.

trunk/examples/concurrent/pthreads/races/regression/atomic_type_array.c
schuessf: > Is this an atomic array of integers, or an array of atomic integers? And do we support this usage?

This should be an array of atomic integers (`(_Atomic int)[]`) and we support this, therefore I added this regression test.

> Hmm, sentence 3 in section 6.7.3 of the C standard (N1570) says
> 
> > The type modified by the _Atomic qualifier shall not be an array type or a function type.
> 
> Is this an issue here?

No, this means that `_Atomic (int[])` is disallowed and this coincides with the change to `CArray` and `CFunction` (the last parameter is always false, i.e., we never consider an atomic array type).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: Thanks for the hint. For non-heap variables this actually works as it is (as the read does not lead to additional statements), but it is currently handled unsound for heap variables (as the translation leads to two separate atomic blocks).

trunk/examples/concurrent/pthreads/races/regression/atomic_type_array.c
maul-esel: :+1: ok, I was just unsure because you had also added some regression test above where the comment says we would fail. So I wasn't sure if there should be a similar comment here.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
maul-esel: Does this (for off-heap variables) also work correctly (1) with the "no-data-race LBE" setting and (2) when data race checks are added (which would be additional statements for the read) ?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CExpressionTranslator.java
schuessf: > The atomic block contains all statements inside the original `exprRes`. Could this not be unsound?

Yes, you are right in principle. For example, if we have some variable `x` that is a pointer to `_Atomic int`, then the pointer increment in `*(x++)++` 😈 should not be executed atomically. 

> Only the assignment of `auxvar` and the assignment created by `funMakeAssignment` should be atomic.

Also the "actual" `switchToRValue` (i.e., the pointer read, without the original `ExpressionResult`) has to be executed atomically. But I currently don't how to handle this properly, without refactoring of every usage `switchToRValue` and `makeAssignment`...

schuessf: Should be fixed in [a54bfac](https://github.com/ultimate-pa/ultimate/pull/688/commits/a54bfacabb660ea591a5e7f8d0839fc5b3e1d060)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: Should be fixed in [a2a635d](https://github.com/ultimate-pa/ultimate/pull/688/commits/a2a635d79fe1f17aaf6bb23bb30828a10d4b7d5a)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -835,7 +836,8 @@ public Result visit(final IDispatcher main, final IASTBinaryExpression node) {
 			final ExpressionResult rr = mExprResultTransformer.transformSwitchRexBoolToInt(rightOperand, loc, node);
 			final ExpressionResult result =
 					mCExpressionTranslator.handleMultiplicativeOperation(loc, node.getOperator(), rl, rr);
-			return makeAssignment(loc, leftOperand.getLrValue(), Collections.emptyList(), result, node);
+			return makeAtomicAssignmentIfNecessary(loc, leftOperand.getLrValue(),
+					DataStructureUtils.concat(leftOperand.getStatements(), rr.getStatements()), result, node);
```
maul-esel: Why do you use `leftOperand.getStatements()` rather than `rl.getStatements()` here?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -2714,6 +2718,19 @@ public static CStorageClass scConstant2StorageClass(final int storageClass) {
...
 	}
 
+	private ExpressionResult makeAtomicAssignmentIfNecessary(final ILocation loc, final LRValue leftHandSide,
+			final List<Statement> nonAtomicStatements, final ExpressionResult rhs, final IASTNode hook) {
+		final ExpressionResult assignment = makeAssignment(loc, leftHandSide, Set.of(), rhs, hook);
+		if (!leftHandSide.getCType().isAtomic()) {
+			return assignment;
+		}
+		final List<Statement> atomicStatements = new ArrayList<>(assignment.getStatements());
+		atomicStatements.removeAll(nonAtomicStatements);
```
maul-esel: As I understand it, the expectation is that `nonAtomicStatements` is a prefix of the list `atomicStatements`. In that case, please add a check here that this indeed holds, and remove the first `nonAtomicStatements.size()` statements.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
maul-esel: Do we still need the handling in this method?

trunk/source/Library-BoogieAST/src/de/uni_freiburg/informatik/ultimate/boogie/StatementFactory.java
```diff
@@ -93,4 +96,21 @@ public static Statement constructIfStatement(final ILocation loc, final Expressi
...
+	public static AtomicStatement constructAtomicStatement(final ILocation loc, final List<Statement> statements) {
+		return constructAtomicStatement(loc, statements.stream());
+	}
+
+	public static AtomicStatement constructAtomicStatement(final ILocation loc, final Stream<Statement> statements) {
+		return new AtomicStatement(loc,
+				statements.flatMap(StatementFactory::getFlattenedStatements).toArray(Statement[]::new));
+	}
+
+	private static Stream<Statement> getFlattenedStatements(final Statement statement) {
```
maul-esel, line 109: :+1: good idea to avoid unnecessary nesting

maul-esel (commented):
Three small questions/requests left, otherwise it looks good. Thanks again for taking care of this feature!

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: Because the switch to RValue (i.e., read of the heap variable if on heap) should be executed atomically.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: Good idea, I will add this!

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: But I will add a comment for this.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: But what do you mean with `nonAtomicStatements.size()`?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: Yes, this method is to handle simple assignment. For clarification I will rename the method `makeAtomicAssignment` to `handleAtomicReadWrite`, since it is used for this purpose.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -2714,6 +2730,26 @@ public static CStorageClass scConstant2StorageClass(final int storageClass) {
...
+			return assignment;
+		}
+		// Check if statementsBeforeRead are a prefix of assignment.getStatements()
+		if (!statementsBeforeRead.equals(assignment.getStatements().subList(0, statementsBeforeRead.size()))) {
+			throw new AssertionError(
+					"Unexpected result of makeAssignment: statements do not start with statements of rhs");
+		}
+		// For atomic types make sure that all statement except the ones before the read are in an atomic block
+		final List<Statement> atomicStatements = new ArrayList<>(assignment.getStatements());
+		atomicStatements.removeAll(statementsBeforeRead);
```
maul-esel:
```suggestion
		final List<Statement> allStatements = assignment.getStatements();

		// Check if statementsBeforeRead are a prefix of allStatements
		if (!statementsBeforeRead.equals(allStatements.subList(0, statementsBeforeRead.size()))) {
			throw new AssertionError(
					"Unexpected result of makeAssignment: statements do not start with statements of rhs");
		}
		// For atomic types make sure that all statement except the ones before the read are in an atomic block
		final List<Statement> atomicStatements =
				allStatements.subList(statementsBeforeRead.size(), allStatements.size());
```

maul-esel (approved):

PR #689, opened 2024-10-22
Support `__atomic_compare_exchange`

maul-esel:
This PR adds support for the builtin atomic functions `__atomic_compare_exchange_n` and `__atomic_compare_exchange`.

### DIFF
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
@@ -1305,8 +1305,9 @@ public class StandardFunctionHandler {
 	// https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#index-_005f_005fatomic_005fcompare_005fexchange_005fn
 	// https://en.cppreference.com/w/c/atomic/atomic_compare_exchange
 	//
-	// The implementation below generates Boogie code that roughly follows this schema, where success and ptr_val are
-	// auxiliary variables:
+	// The implementation below generates Boogie code that roughly follows the schema below, where success and ptr_val
+	// are auxiliary variables. However, if the memoryOrder argument is not sequential consistency, we instead
+	// overapproximate the method with "assert false".
 	//
 	// @formatter:off
 	// (evaluate arguments)
@@ -1432,8 +1433,9 @@ public class StandardFunctionHandler {
 	// https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#index-_005f_005fatomic_005fcompare_005fexchange_005fn
 	// https://en.cppreference.com/w/c/atomic/atomic_compare_exchange
 	//
-	// The implementation below generates Boogie code that roughly follows this schema, where success and ptr_val are
-	// auxiliary variables:
+	// The implementation below generates Boogie code that roughly follows the schema below, where success and ptr_val
+	// are auxiliary variables. However, if the memoryOrder argument is not sequential consistency, we instead
+	// overapproximate the method with "assert false".
 	//
 	// @formatter:off
 	// (evaluate arguments)
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
@@ -125,6 +125,7 @@ import de.uni_freiburg.informatik.ultimate.core.model.models.ILocation;
 import de.uni_freiburg.informatik.ultimate.core.model.models.annotation.Spec;
 import de.uni_freiburg.informatik.ultimate.core.model.services.ILogger;
 import de.uni_freiburg.informatik.ultimate.plugins.generator.cacsl2boogietranslator.preferences.CACSLPreferenceInitializer.CheckMode;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.DataStructureUtils;
 import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.Pair;
 
 /**
@@ -502,8 +503,9 @@ public class StandardFunctionHandler {
 		// Preprocessing leads to: https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html
 
 		fill(map, "__atomic_load", this::handleAtomicLoad);
-		fill(map, "__atomic_exchange", this::handleAtomicExchange);
 		fill(map, "__atomic_store", this::handleAtomicStore);
+		fill(map, "__atomic_exchange", this::handleAtomicExchange);
+		fill(map, "__atomic_compare_exchange", this::handleAtomicCompareExchange);
 
 		fill(map, "__atomic_load_n", this::handleAtomicLoadN);
 		fill(map, "__atomic_store_n", this::handleAtomicStoreN);
@@ -524,9 +526,6 @@ public class StandardFunctionHandler {
 		fill(map, "__atomic_test_and_set", this::handleAtomicTestAndSet);
 		fill(map, "__atomic_clear", this::handleAtomicClear);
 
-		fill(map, "__atomic_compare_exchange",
-				(main, node, loc, name) -> handleUnsupportedFunctionByOverapproximation(main, loc, name,
-						new CPrimitive(CPrimitives.BOOL)));
 		fill(map, "__atomic_thread_fence", (main, node, loc, name) -> handleUnsupportedFunctionByOverapproximation(main,
 				loc, name, new CPrimitive(CPrimitives.VOID)));
 		fill(map, "__atomic_signal_fence", (main, node, loc, name) -> handleUnsupportedFunctionByOverapproximation(main,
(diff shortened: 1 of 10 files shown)


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -1395,30 +1550,35 @@ private Result handleAtomicFetch(final IDispatcher main, final IASTFunctionCallE
...
 	 *            The body that should be atomic based on the memory order
-	 * @param memoryOrder
-	 *            The memory order
+	 * @param memoryOrders
+	 *            The memory orders to apply
 	 * @return An ExpressionResult representing the translation respecting the memory order
 	 */
 	private ExpressionResult applyMemoryOrder(final ILocation loc, final ExpressionResult body,
-			final Expression memoryOrder) {
+			final Expression... memoryOrders) {
```
schuessf, line 1523: You updated the signature of the function, can you also update the documentation? It is not quite clear to me, how the multiple memory orders are handled here.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -1353,6 +1430,84 @@ private Result handleAtomicExchangeN(final IDispatcher main, final IASTFunctionC
...
 
+	// https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#index-_005f_005fatomic_005fcompare_005fexchange_005fn
+	// https://en.cppreference.com/w/c/atomic/atomic_compare_exchange
+	//
+	// The implementation below generates Boogie code that roughly follows this schema, where success and ptr_val are
+	// auxiliary variables:
+	//
+	// @formatter:off
+	// (evaluate arguments)
+	// atomic {
```
schuessf: Thanks for the illustration. This really helps to understand the specification.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -1353,6 +1430,84 @@ private Result handleAtomicExchangeN(final IDispatcher main, final IASTFunctionC
...
+		final var pointerWrite = mExprResultTransformer.makePointerAssignment(loc, pointer.getLrValue(),
+				desiredResult.getLrValue().getValue());
+
+		final var successBoolean =
+				mExpressionTranslation.constructBinaryEqualityExpression(loc, IASTBinaryExpression.op_notequals,
+						success.getExp(), boolType, mExpressionTranslation.constructZero(loc, boolType), boolType);
+		final var atomicBody = new ExpressionResultBuilder().addAuxVarWithDeclaration(success)
+				.addStatement(StatementFactory.constructIfStatement(loc, weakResult.getLrValue().getValue(),
+						new Statement[] { new HavocStatement(loc, new VariableLHS[] { success.getLhs() }) },
+						new Statement[] { StatementFactory.constructAssignmentStatement(loc,
```
schuessf: There is also the method `StatementFactory::constructSingleAssignmentStatement`, which improves readability (it already creates singleton arrays).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: Is this block really atomic or does it depend on the memory order?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -1353,6 +1430,84 @@ private Result handleAtomicExchangeN(final IDispatcher main, final IASTFunctionC
...
+		final var success = mAuxVarInfoBuilder.constructAuxVarInfo(loc, boolType, AUXVAR.RETURNED);
+		final var pointerRead = mExprResultTransformer.readPointerValue(loc, pointer.getLrValue());
+		final var expectedRead = mExprResultTransformer.readPointerValue(loc, expectedResult.getLrValue());
+		final var pointerWrite = mExprResultTransformer.makePointerAssignment(loc, pointer.getLrValue(),
+				desiredResult.getLrValue().getValue());
+
+		final var successBoolean =
+				mExpressionTranslation.constructBinaryEqualityExpression(loc, IASTBinaryExpression.op_notequals,
+						success.getExp(), boolType, mExpressionTranslation.constructZero(loc, boolType), boolType);
+		final var atomicBody = new ExpressionResultBuilder().addAuxVarWithDeclaration(success)
```
schuessf: This is a huge single statement. I think, it should improve readability if you refactor it in smaller statements.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -1305,6 +1302,86 @@ private Result handleAtomicExchange(final IDispatcher main, final IASTFunctionCa
...
+	//  if success && ptr_val == read(expected):
+	//    write(read(desired), ptr)
+	//  else
+	//    success := false
+	// }
+	// if !success
+	//   write(ptr_val, expected)
+	// return success
+	// @formatter:on
+	private Result handleAtomicCompareExchange(final IDispatcher main, final IASTFunctionCallExpression node,
```
schuessf, line 1360: If I understand the specification correctly, then the only difference between `__atomic_compare_exchange_n` and `__atomic_compare_exchange` is that the latter one has a pointer as `desired` that is simply dereferenced and written into `ptr`. Is it possible to refactor this method to call `handleAtomicCompareExchangeN`?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -1353,6 +1430,84 @@ private Result handleAtomicExchangeN(final IDispatcher main, final IASTFunctionC
...
+								new LeftHandSide[] { success.getLhs() }, new Expression[] { mExpressionTranslation
+										.constructLiteralForIntegerType(loc, boolType, BigInteger.ZERO) }) }))
+				.build();
+		final var atomic = applyMemoryOrder(loc, atomicBody, successMemoryOrder.getLrValue().getValue(),
+				failureMemoryOrder.getLrValue().getValue());
+
+		final var expectedWrite = mExprResultTransformer.makePointerAssignment(loc, expectedResult.getLrValue(),
+				pointerRead.getLrValue().getValue());
+		return resultBuilder.addAllExceptLrValue(atomic).addAllExceptLrValueAndStatements(expectedWrite)
+				.addStatement(StatementFactory.constructIfStatement(loc, successBoolean, new Statement[0],
```
schuessf: Minor note: Is there any reason why you create `if (success) {} else st` instead of `if (!success) st` here? It should not make a difference on the CFG, but the Boogie code of the latter looks slightly better?

schuessf (commented):
Thanks for the PR! I think it looks good, I have just some minor comments.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
maul-esel: :+1: done

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
maul-esel: I forgot about that. I didn't see how to apply negation at first, and only wanted to test out the method briefly :D good catch, fixed now.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
maul-esel: :+1: hopefully clearer now

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
maul-esel: I've added a comment explaining that we only follow this scheme if the memory order is sequential consistency, and otherwise `assert false`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
maul-esel: I found it sometimes difficult to read other translation code if it first assembles lots of little pieces of Boogie (without context what they are useful for) and then assembles them at the end in some order.

So I tried to stick as closely as possible to the structure of the generated Boogie code here... though I can't say the result is necessarily much better :face_with_diagonal_mouth:

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: Thanks!

> If every given memory order is equal to MEMORY_ORDER_SEQ_CST, we just make all statements atomic.

If understand correctly, you create nested ifs:

```
if (mem1 == SEQ_CST) {
  if (mem2 ==  SEQ_CST) {
    atomic { st };
  } else {
    assert false;
} else {
  assert false;
}
```

Wouldn't it be simpler to create simply the conjunction of all memory-order equalities?

```
if (mem1 == SEQ_CST && mem2 ==  SEQ_CST) {
    atomic { st };
} else {
  assert false;
}
```

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -1301,10 +1298,123 @@ private Result handleAtomicExchange(final IDispatcher main, final IASTFunctionCa
...
+						new Statement[] { StatementFactory.constructSingleAssignmentStatement(loc, success.getLhs(),
+								mExpressionTranslation.constructLiteralForIntegerType(loc, boolType,
+										BigInteger.ZERO)) }))
+				.build();
+		final var atomic = applyMemoryOrders(loc, atomicBody, successMemoryOrder.getLrValue().getValue(),
+				failureMemoryOrder.getLrValue().getValue());
+
+		final var expectedWrite = mExprResultTransformer.makePointerAssignment(loc, expectedResult.getLrValue(),
+				pointerRead.getLrValue().getValue());
+		return resultBuilder.addAllExceptLrValue(atomic).addAllExceptLrValueAndStatements(expectedWrite)
```
schuessf: Is it correct that the write is indeed not atomic? I don't fully understand this from the specification.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -1301,10 +1298,123 @@ private Result handleAtomicExchange(final IDispatcher main, final IASTFunctionCa
...
+				.addAllExceptLrValue(pointerRead, expectedRead).addAllExceptLrValueAndStatements(desiredRead)
+				.addAllExceptLrValueAndStatements(pointerWrite)
+				.addStatement(StatementFactory.constructIfStatement(loc,
+						new BinaryExpression(loc, Operator.LOGICAND, successBoolean,
+								new BinaryExpression(loc, Operator.COMPEQ, pointerRead.getLrValue().getValue(),
+										expectedRead.getLrValue().getValue())),
+						DataStructureUtils.concat(desiredRead.getStatements(), pointerWrite.getStatements())
+								.toArray(Statement[]::new),
+						new Statement[] { StatementFactory.constructSingleAssignmentStatement(loc, success.getLhs(),
+								mExpressionTranslation.constructLiteralForIntegerType(loc, boolType,
```
schuessf: You could use `boolZero` here

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: Okay, I see. I think it looks better now. Maybe can improve the readability further to move the constructor call (maybe with the aux var) and the `build()` to separate statements. I think the indentation could be reduced in that case.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
maul-esel: Hmm, I'm not sure. I was basing this on the description at https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html#index-_005f_005fatomic_005fcompare_005fexchange_005fn which says:

> If equal, the operation is a _read-modify-write_ operation that writes `desired` into `*ptr`. If they are not equal, the operation is a _read_ and the current contents of `*ptr` are written into `*expected.`

I understood that to say that the write is at least not atomic with the rest, otherwise the operation would be a _read-modify-write_. Though that still leaves open whether the write is atomic by itself.

The description of `atomic_compare_exchange` in Section 7.17.7.4 of the C standard sounds somewhat different: 
image

This seems more like the failure case is atomic with the rest. Of course, we're implementing the GNU variant, which "approximates" the standard... :shrug:

maul-esel: The C standard is also a bit more precise what "fail spuriously" looks like; it's different from the behaviour for different values. I will have to update the code for that as well.
image

maul-esel: I've reimplemented the method based on the C standard description.

schuessf (approved):
Looks good! I just assume your implementation follows the vague standard 😉

PR #692, opened 2024-11-08
Refactor: Use appropriate control configurations in trace checks

maul-esel:
As discussed internally, this is an attempt at fixing the detection of _perfect interpolant sequences_, and some other usages of the control locations (or in the case of concurrent programs, _control configurations_) visited along a counterexample trace.

## Motivation
At various points of the code, we either used the `IcfgLocation` returned by `IIcfgTransition::getSource` (usually via `TraceCheckUtils::getSequenceOfProgramPoints`) or the predicate in the `IRun` for the counterexample.
- The `IcfgLocation`s do not make sense in the context of concurrent programs, we want the whole control configuration, i.e., the locations of all threads.
- The run predicates are typically not what is intended either. If the run is a run of the difference automaton between program and interpolant automata, and is found by DFS or BFS, it will never contain any repeated state, and is thus useless for determining `isPerfect`.

## Changes
1. In places where the control configurations / states of the run are not needed at all, I have replaced `IRun` by `Word`.
  1.1 This includes usages in LassoChecker, where, according to @Heizmann, any emphasis on the usage of "perfect" proofs was unintentional.
2. In other places, I have replaced an `IRun` by a new class `Counterexample<L>`, which bundles the trace (a `Word<L>`) and an optional `List<Object> controlConfigurationSequences`.
  - 2.1 The type of configurations is `Object`, any type will do -- we only care about equality of instances (as defined by `Object::equals`).
  - 2.2 It is easier to pass some list rather than creating an `IRun` on the original program (e.g. a CEGAR loop's initial abstraction).
  
As a lot of these changes touch code I am unfamiliar with, I would appreciate it if someone with more knowledge could take a look, in particular for _accelerated interpolation_ and _assert order modulation_. @danieldietsch?

## Evaluation

The nightly tests look fine: https://jenkins.sopranium.de/job/Ultimate/job/Ultimate%20Nightly/job/wip%252Fdk%252Fperfect-tracechecks/4/testReport/

Once the current evaluations are finished, I will run a fast SV-COMP run on this branch to stress-test these changes, and also to see if the "correct" notion of perfect proofs for concurrent programs has any positive impact.

_Update: SV-COMP evaluation is currently running._


maul-esel:
The conflict is with the addition of the new Fox refinement strategy. I'll rebase after I have run the SV-COMP evaluation, because the current base is our reference point for the comparison.

trunk/source/Library-AcceleratedInterpolation/src/de/uni_freiburg/informatik/ultimate/lib/acceleratedinterpolation/AcceleratedInterpolation.java
```diff
@@ -124,14 +125,15 @@ public class AcceleratedInterpolation<L extends IIcfgTransition<?>> implements I
 	 */
 	public AcceleratedInterpolation(final IUltimateServiceProvider services, final ILogger logger,
 			final ITraceCheckPreferences prefs, final ManagedScript script, final IPredicateUnifier predicateUnifier,
-			final IRun<L, IPredicate> counterexample, final Class<L> transitionClazz,
+			final Word<L> counterexample, final List<?> controlLocations, final Class<L> transitionClazz,
```
danieldietsch: I had hoped we would have some datatype that keeps those two together. It seems not nice to pass them around independently. 

This happens now in multiple places: one passes these around without signifying that they actually belong together. Returning them is even worse, you now need a pair. 

Please introduce some specialized class that keeps these together and use it in all places where they are now duplicated.

trunk/source/Library-LassoRanker/src/de/uni_freiburg/informatik/ultimate/lassoranker/nontermination/FixpointCheck2.java
```diff
@@ -86,8 +85,8 @@ public FixpointCheck2(final IUltimateServiceProvider services, final ILogger log
 		final IPredicate postcondition = constructNegatedLoopPredicate(services, csToolkit.getManagedScript(), pf,
 				mLoop);
 		final SortedMap<Integer, IPredicate> pendingContexts = Collections.emptySortedMap();
-		final TraceCheck<L> tc = new TraceCheck<L>(precondition, postcondition, pendingContexts, mStem.getWord(),
-				services, csToolkit, AssertCodeBlockOrder.NOT_INCREMENTALLY, true, false);
+		final TraceCheck<L> tc = new TraceCheck<>(precondition, postcondition, pendingContexts, mStem, null, services,
+				csToolkit, AssertCodeBlockOrder.NOT_INCREMENTALLY, true, false);
```
danieldietsch: Why do you need to pass null here? I thought a central point of this refactoring was to get rid of the misleading constructor in TraceCheck that requires you to add a control location sequence although it never requires it (and fixing the AssertCodeBlockOrder to NOT_INCREMENTALLY and computeIcfgExecution to false for that constructor).

trunk/source/Library-PDR/src/de/uni_freiburg/informatik/ultimate/lib/pdr/Pdr.java
```diff
@@ -1215,7 +1215,7 @@ private final IPredicate[] computeInterpolants() {
 						mSymbolTable);
 
 		final OldVarsAssignmentCache oldVarsAssignmentCache = mCsToolkit.getOldVarsAssignmentCache();
-		final DefaultTransFormulas<L> rtf = new DefaultTransFormulas<>(nestedWord, mTruePred, mFalsePred,
+		final DefaultTransFormulas<L> rtf = new DefaultTransFormulas<>(nestedWord, null, mTruePred, mFalsePred,
```
danieldietsch: Why do you need to pass null here?

trunk/source/Library-AcceleratedInterpolation/src/de/uni_freiburg/informatik/ultimate/lib/acceleratedinterpolation/MetaTraceTransformer.java
```diff
@@ -226,8 +226,8 @@ private IPredicate[] getInductiveFirstEntryOnlyIterativePredTransformer(final IP
 			final Integer start, final Integer end, final IPredicate precondition, final IPredicate postcondition) {
 		final List<L> loopTrace = mCounterexample.subList(start, end);
 		final NestedWord<L> nestedWordLoopTrace = TraceCheckUtils.toNestedWord(loopTrace);
-		final DefaultTransFormulas<L> rtf = new DefaultTransFormulas<>(nestedWordLoopTrace, precondition, postcondition,
-				Collections.emptySortedMap(), mToolkit.getOldVarsAssignmentCache(), false);
+		final DefaultTransFormulas<L> rtf = new DefaultTransFormulas<>(nestedWordLoopTrace, null, precondition,
```
danieldietsch: Why do you need to pass null here?

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/AnnotateAndAsserter.java
```diff
@@ -51,6 +52,7 @@ public class AnnotateAndAsserter<L extends IAction> {
 
 	protected final ManagedScript mMgdScriptTc;
 	protected final NestedWord<L> mTrace;
+	protected final List<?> mControlConfigurationSequence;
```
danieldietsch: Why is it called mControlConfiugrationSequence instead of mControlLocationSequence?

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/AnnotateAndAsserter.java
```diff
@@ -67,6 +69,7 @@ public AnnotateAndAsserter(final ManagedScript mgdScriptTc, final NestedFormulas
 		mLogger = mServices.getLoggingService().getLogger(TraceCheckerUtils.PLUGIN_ID);
 		mMgdScriptTc = mgdScriptTc;
 		mTrace = nestedSSA.getTrace();
+		mControlConfigurationSequence = nestedSSA.getControlConfigurations();
```
danieldietsch: Why is it called mControlConfiugrationSequence instead of mControlLocationSequence?

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/AnnotateAndAsserterWithStmtOrderPrioritization.java
```diff
@@ -190,17 +189,18 @@ private <LOC> Map<Integer, Set<Integer>> partitionStatementsAccordingDepth(
 	@Override
 	public void buildAnnotatedSsaAndAssertTerms() {
 		assert mCheckSat == 0 : "You should not call this method twice";
-		final List<IcfgLocation> pps = TraceCheckUtils.getSequenceOfProgramPoints(mTrace);
```
danieldietsch, line 193: Ah, rewards :)

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/DefaultTransFormulas.java
```diff
@@ -43,10 +44,11 @@ public class DefaultTransFormulas<L extends IAction> extends NestedFormulas<L, U
 	private final OldVarsAssignmentCache mModifiableGlobalVariableManager;
 	private final boolean mWithBranchEncoders;
 
-	public DefaultTransFormulas(final NestedWord<L> nestedWord, final IPredicate precondition,
-			final IPredicate postcondition, final SortedMap<Integer, IPredicate> pendingContexts,
-			final OldVarsAssignmentCache oldVarsAssignmentCache, final boolean withBranchEncoders) {
-		super(nestedWord, pendingContexts);
+	public DefaultTransFormulas(final NestedWord<L> nestedWord, final List<?> controlConfigurationSequence,
```
danieldietsch: location vs configuration

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/InterpolatingTraceCheck.java
```diff
@@ -66,7 +66,7 @@ public abstract class InterpolatingTraceCheck<L extends IAction> extends TraceCh
 
 	protected IPredicate[] mInterpolants;
 
-	private final List<? extends Object> mControlLocationSequence;
+	protected final List<?> mControlConfigurationSequence;
```
danieldietsch, line 69: location vs configuration

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/InterpolatingTraceCheck.java
```diff
@@ -75,21 +75,21 @@ public abstract class InterpolatingTraceCheck<L extends IAction> extends TraceCh
...
+			final List<?> controlLocationSequence, final IUltimateServiceProvider services,
 			final CfgSmtToolkit csToolkit, final ManagedScript tcSmtManager, final PredicateFactory predicateFactory,
 			final IPredicateUnifier predicateUnifier, final AssertCodeBlockOrder assertCodeBlockOrder,
 			final boolean computeRcfgProgramExecution, final boolean collectInterpolatSequenceStatistics,
 			final SimplificationTechnique simplificationTechnique) {
 		super(precondition, postcondition, pendingContexts, trace,
 				TraceCheckUtils.decoupleArrayValues(csToolkit.getManagedScript(),
-						new DefaultTransFormulas<>(trace, precondition, postcondition, pendingContexts,
-								csToolkit.getOldVarsAssignmentCache(), false)),
+						new DefaultTransFormulas<>(trace, (List) controlLocationSequence, precondition, postcondition,
```
danieldietsch: Why this raw cast?

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/InterpolatingTraceCheck.java
```diff
@@ -75,21 +75,21 @@ public abstract class InterpolatingTraceCheck<L extends IAction> extends TraceCh
...
+								pendingContexts, csToolkit.getOldVarsAssignmentCache(), false)),
 				services, csToolkit, tcSmtManager, assertCodeBlockOrder, computeRcfgProgramExecution,
 				collectInterpolatSequenceStatistics, false);
 		mPredicateUnifier = predicateUnifier;
 		mPredicateFactory = predicateFactory;
 		mSimplificationTechnique = simplificationTechnique;
-		mControlLocationSequence = controlLocationSequence;
+		mControlConfigurationSequence = controlLocationSequence;
```
danieldietsch: location vs configuration

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/ModifiableNestedFormulas.java
```diff
@@ -66,8 +67,9 @@ public class ModifiableNestedFormulas<L extends IAction, TF, SF> extends NestedF
 	private final Map<Integer, TF> mGlobalOldVarAssignmentAtCall = new HashMap<>();
 
 	@SuppressWarnings("unchecked")
-	public ModifiableNestedFormulas(final NestedWord<L> nestedWord, final SortedMap<Integer, SF> pendingContexts) {
-		super(nestedWord, pendingContexts);
+	public ModifiableNestedFormulas(final NestedWord<L> nestedWord, final SortedMap<Integer, SF> pendingContexts,
+			final List<?> controlConfigurationSequence) {
```
danieldietsch: location vs configuration

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/NestedFormulas.java
```diff
@@ -53,20 +54,27 @@
...
 
 	private final NestedWord<L> mNestedWord;
+	private final List<?> mControlConfigurationSequence;
 	private SF mPrecondition;
 	private SF mPostcondition;
 	private final SortedMap<Integer, SF> mPendingContexts;
 
-	public NestedFormulas(final NestedWord<L> nestedWord, final SortedMap<Integer, SF> pendingContexts) {
+	public NestedFormulas(final NestedWord<L> nestedWord, final SortedMap<Integer, SF> pendingContexts,
+			final List<?> controlConfigurationSequence) {
```
danieldietsch: location vs configuration

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/NestedFormulas.java
```diff
@@ -53,20 +54,27 @@
...
 		assert pendingContexts != null;
 		mPendingContexts = pendingContexts;
+		mControlConfigurationSequence = controlConfigurationSequence;
 	}
 
 	public final NestedWord<L> getTrace() {
 		return mNestedWord;
 	}
 
+	public final List<?> getControlConfigurations() {
```
danieldietsch: location vs configuration

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/RelevantTransFormulas.java
```diff
@@ -89,12 +90,12 @@ public class RelevantTransFormulas<L extends IAction> extends NestedFormulas<L,
 	 */
 	private int mSumSizeFormulasInUnsatCore = 0;
 
-	public RelevantTransFormulas(final NestedWord<L> nestedTrace, final IPredicate precondition,
-			final IPredicate postcondition, final SortedMap<Integer, IPredicate> pendingContexts,
-			final Set<L> unsatCore, final OldVarsAssignmentCache oldVarsAssignmentCache,
-			final boolean[] localVarAssignmentsAtCallInUnsatCore, final boolean[] oldVarAssignmentAtCallInUnsatCore,
-			final ManagedScript script) {
-		super(nestedTrace, pendingContexts);
+	public RelevantTransFormulas(final NestedWord<L> nestedTrace, final List<?> controlConfigurationSequence,
```
danieldietsch: location vs configuration

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/TraceCheck.java
```diff
@@ -263,11 +263,11 @@ protected TraceCheck(final IPredicate precondition, final IPredicate postconditi
 	 */
 	public static TraceCheck<IAction> createTraceCheck(final IUltimateServiceProvider services,
```
danieldietsch: I think we should check where we use this method. In particular, as it uses `computeRcfgProgramExecution`, it will take much more time than necessary and still require control locations. 

Tracecheck should have something similar, but with computeRcfgProgramExecution=false and no control locations.

trunk/source/PEAtoBoogie/src/de/uni_freiburg/informatik/ultimate/pea2boogie/VerificationResultTransformer.java
```diff
@@ -345,7 +345,7 @@ private IProgramExecution<IAction, Term> reduceRtInconsistencyProgramExecution(f
 			if (hasInvalidBranchEncoders(pe)) {
 				mLogger.info("Computing branch encoders");
 				final TraceCheck<IAction> tcl =
-						TraceCheck.createTraceCheck(mServices, toolkit, mgdScriptTc, truePred, falsePred, trace);
+						TraceCheck.createTraceCheck(mServices, toolkit, mgdScriptTc, truePred, falsePred, trace, null);
```
danieldietsch: Wont that fail as it wants to compute a rcfgprogramexecution, and it needs control locations for that

trunk/source/PEAtoBoogie/src/de/uni_freiburg/informatik/ultimate/pea2boogie/VerificationResultTransformer.java
```diff
@@ -361,18 +361,18 @@ private IProgramExecution<IAction, Term> reduceRtInconsistencyProgramExecution(f
 			final List<IAction> cleanedTrace = removeUnrelatedVariables(sequentialTrace, reqCheck, mgdScriptTc);
 
 			mLogger.info("Computing reduced program execution");
-			final TraceCheck<IAction> tc =
-					TraceCheck.createTraceCheck(mServices, toolkit, mgdScriptTc, truePred, falsePred, cleanedTrace);
+			final TraceCheck<IAction> tc = TraceCheck.createTraceCheck(mServices, toolkit, mgdScriptTc, truePred,
+					falsePred, cleanedTrace, null);
```
danieldietsch: Same as above

trunk/source/PEAtoBoogie/src/de/uni_freiburg/informatik/ultimate/pea2boogie/VerificationResultTransformer.java
```diff
@@ -361,18 +361,18 @@ private IProgramExecution<IAction, Term> reduceRtInconsistencyProgramExecution(f
...
 
 			// should be unreachable
 			mLogger.fatal("Reduced program execution is not 'sat'");
 			final TraceCheck<IAction> tcOrig =
-					TraceCheck.createTraceCheck(mServices, toolkit, mgdScriptTc, truePred, falsePred, trace);
-			final TraceCheck<IAction> tcSeq =
-					TraceCheck.createTraceCheck(mServices, toolkit, mgdScriptTc, truePred, falsePred, sequentialTrace);
+					TraceCheck.createTraceCheck(mServices, toolkit, mgdScriptTc, truePred, falsePred, trace, null);
+			final TraceCheck<IAction> tcSeq = TraceCheck.createTraceCheck(mServices, toolkit, mgdScriptTc, truePred,
+					falsePred, sequentialTrace, null);
```
danieldietsch: Same as above

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
```diff
@@ -283,17 +283,21 @@ protected final void readInitialProof() throws AutomataLibraryException {
...
+
 	@Override
 	protected Pair<LBool, IProgramExecution<L, Term>> isCounterexampleFeasible()
 			throws AutomataOperationCanceledException {
 
 		IStatisticsDataProvider refinementEngineStats = null;
-		final ITARefinementStrategy<L> strategy = mStrategyFactory.constructStrategy(getServices(), mCounterexample,
-				mAbstraction, new SubtaskIterationIdentifier(mTaskIdentifier, getIteration()),
-				mPredicateFactoryInterpolantAutomata, getPreconditionProvider(), getPostconditionProvider());
+		final var locations = getControlLocationsFromCounterexample(mCounterexample);
```
danieldietsch: Why is this needed? Shouldnt such a method not be part of IRun?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/CegarAbsIntRunner.java
```diff
@@ -515,11 +515,11 @@ public IInterpolatingTraceCheck<LETTER> getInterpolantGenerator() {
 		private IInterpolatingTraceCheck<LETTER> createInterpolantGenerator() {
 			if (mResult.hasReachedError()) {
 				// analysis was not strong enough
-				return new AbsIntFailedInterpolantGenerator<>(mPredicateUnifierAbsInt, mCex.getWord(),
+				return new AbsIntFailedInterpolantGenerator<>(mPredicateUnifierAbsInt, mCex,
 						ItpErrorStatus.ALGORITHM_FAILED, null);
 			}
 			// we were strong enough!
-			final Word<LETTER> word = mCex.getWord();
+			final Word<LETTER> word = mCex;
```
danieldietsch: inline word

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
```diff
@@ -574,4 +576,15 @@ private static SearchStrategy getSearchStrategy(final IPreferenceProvider mPrefs
 			throw new IllegalArgumentException();
 		}
 	}
+
+	private List<IcfgLocation> getIcfgLocationsFromRun(final NestedRun<L, IPredicate> run) {
+		return run.getStateSequence().stream().map(p -> ((ISLPredicate) p).getProgramPoint())
+				.collect(Collectors.toList());
```
danieldietsch: Isnt that the kind of hack we wanted to get rid of?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
```diff
@@ -574,4 +576,15 @@ private static SearchStrategy getSearchStrategy(final IPreferenceProvider mPrefs
...
+	}
+
+	@Override
+	protected List<?> getControlLocationsFromCounterexample(final IRun<L, ?> run) {
+		return run.getStateSequence().stream().map(s -> PredicateUtils.getLocations((IPredicate) s))
```
danieldietsch: And why do we need it twice?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/SifaRunner.java
```diff
@@ -82,18 +82,17 @@ public final class SifaRunner<L extends IIcfgTransition<?>> implements IInterpol
...
+		final Set<L> pathProgramSet = currentCex.asSet();
 		final PathProgramConstructionResult ppConstructionResult = PathProgram.constructPathProgram("sifa-path-program",
 				icfg, pathProgramSet, Collections.emptySet(), x -> true);
 		final PathProgram pp = ppConstructionResult.getPathProgram();
 
-		final List<IcfgLocation> locationOfInterestList =
-				TraceCheckUtils.getSequenceOfProgramPoints(currentCex.getWord()).stream()
-						.map(a -> ppConstructionResult.getLocationMapping().get(a)).collect(Collectors.toList());
+		final List<IcfgLocation> locationOfInterestList = TraceCheckUtils.getSequenceOfProgramPoints(currentCex)
+				.stream().map(a -> ppConstructionResult.getLocationMapping().get(a)).collect(Collectors.toList());
```
danieldietsch, line 95: Looks like we need the sequence of control locations here as well. Why is it missing here?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/acceleratedtracecheck/AcceleratedTraceCheck.java
```diff
@@ -197,17 +194,18 @@ private IPredicate[] constructInterpolants(final NestedRun<L, IPredicate> counte
...
 			assert result.size() == aseg.getStartPosition();
 			final IPredicate precondition = interpolantsForAcceleratedTrace.get(aseg.getStartPosition() - offset - 1);
 			final IPredicate postcondition = interpolantsForAcceleratedTrace.get(aseg.getStartPosition() - offset);
 			// Run for the trace from the first letter (inclusive) to the last letter
 			// (inclusive) of the loop body
-			final NestedRun<L, IPredicate> subRun = counterexample.getSubRun(aseg.getStartPosition(),
-					aseg.getEndPosition() + 1);
-			final TraceCheckSpWp<L> inter = checkTrace(precondition, postcondition, subRun);
+			final NestedWord<L> subWord = counterexample.getSubWord(aseg.getStartPosition(), aseg.getEndPosition() + 1);
+			final List<?> subCCS = controlConfigurationSequence == null ? null
```
danieldietsch: Why is controlConfigurationSequence allowed to be null?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/concurrency/CegarLoopForPetriNet.java
```diff
@@ -580,6 +582,17 @@ public IStatisticsDataProvider getCegarLoopBenchmark() {
...
 
+	@Override
+	protected List<?> getControlLocationsFromCounterexample(final IRun<L, ?> run) {
+		final var pnRun = (PetriNetRun<L, IPredicate>) run;
+
+		// Take the places in the run that belong to the initial abstraction.
+		// Remove any places belonging to the interpolant automata.
+		return pnRun.getStateSequence().stream()
+				.map(m -> m.stream().filter(mProgramPointPlaces::contains).collect(Collectors.toSet()))
+				.collect(Collectors.toList());
```
danieldietsch, line 593: I always thought that the control location sequence has to have length of word.size()+1. How do you ensure that here?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/errorabstraction/ErrorAutomatonBuilder.java
```diff
@@ -270,7 +270,7 @@ private TracePredicates getPredicates(final IUltimateServiceProvider services, f
 			final IIcfgSymbolTable symbolTable, final IPredicate truePredicate,
 			final NestedWord<L> trace, final IPredicate precondition, final IPredicate postcondition,
 			final List<IPredicatePostprocessor> postprocessors, final PredicateTransformerType predicateTransformer) throws AssertionError {
-		final DefaultTransFormulas<L> dtf = new DefaultTransFormulas<>(trace, precondition, postcondition,
+		final DefaultTransFormulas<L> dtf = new DefaultTransFormulas<>(trace, null, precondition, postcondition,
```
danieldietsch: I still dont understand why transformulas have an optional field for control location sequences.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/interpolantautomata/builders/AbsIntNonSmtInterpolantAutomatonBuilder.java
```diff
@@ -62,18 +60,22 @@ public class AbsIntNonSmtInterpolantAutomatonBuilder<LETTER>
 	private final IUltimateServiceProvider mServices;
 	private final ILogger mLogger;
 	private final NestedWordAutomaton<LETTER, IPredicate> mResult;
-	private final IRun<LETTER, ?> mCurrentCounterExample;
+	private final Word<LETTER> mCurrentCounterExample;
+	private final List<?> mControlConfigurationSequence;
 	private final PredicateFactory mPredicateFactory;
 	private final ManagedScript mBoogie2Smt;
 
+	// TODO Does this really need an IRun?
```
danieldietsch: I dont see why, I guess a word is enough.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/interpolantautomata/builders/TotalInterpolationAutomatonBuilder.java
```diff
@@ -104,16 +104,15 @@ public class TotalInterpolationAutomatonBuilder<LETTER extends IIcfgTransition<?
 	private final boolean mCollectInterpolantStatistics;
 
 	public TotalInterpolationAutomatonBuilder(final INestedWordAutomaton<LETTER, IPredicate> abstraction,
-			final NestedRun<LETTER, IPredicate> nestedRun, final CfgSmtToolkit csToolkit,
+			final List<IPredicate> run, final CfgSmtToolkit csToolkit,
```
danieldietsch: Perhaps call it stateSequence?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/pathinvariants/InterpolatingTraceCheckPathInvariantsWithFallback.java
```diff
@@ -88,22 +91,22 @@ public InterpolatingTraceCheckPathInvariantsWithFallback(final IPredicate precon
...
+				mSimplificationTechnique);
 		mInterpolantComputationStatus = pathInvariantsGenerator.getInterpolantComputationStatus();
 		final IPredicate[] interpolants = pathInvariantsGenerator.getInterpolants();
 		if (interpolants == null) {
 			assert !pathInvariantsGenerator.getInterpolantComputationStatus()
 					.wasComputationSuccessful() : "null only allowed if computation was not successful";
 		} else {
 			if (interpolants.length != mTrace.length() - 1) {
-				throw new AssertionError("inkorrekt number of interpolants. "
+				throw new AssertionError("incorrekt number of interpolants. "
```
danieldietsch: Got only one correct ;)

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/IpAbStrategyModuleTotalInterpolation.java
```diff
@@ -94,7 +92,7 @@ public IpAbStrategyModuleResult<LETTER> buildInterpolantAutomaton(final List<Qua
 			final INestedWordAutomaton<LETTER, IPredicate> castedAbstraction =
 					(INestedWordAutomaton<LETTER, IPredicate>) mAbstraction;
 			@SuppressWarnings("unchecked")
-			final NestedRun<LETTER, IPredicate> castedCex = (NestedRun<LETTER, IPredicate>) mCounterexample;
+			final List<IPredicate> castedCex = (List<IPredicate>) mCounterexampleRun;
```
danieldietsch, line 95: Would be nice if we could get rid of these casts

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/Mcr.java
```diff
@@ -102,7 +102,8 @@ private McrTraceCheckResult<L> exploreInterleavings(final List<L> initialTrace)
 		while (counterexample != null) {
 			mLogger.info("---- MCR iteration " + iteration + " ----");
 			iteration++;
-			result = mResultProvider.getResult(counterexample);
+			// TODO Is it correct to not pass any control locations here?
+			result = mResultProvider.getResult(counterexample.getWord(), null);
```
danieldietsch: @schuessf I am not sure, but perhaps you remember? 

Independent of this, passing null just seems wrong.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/StrategyFactory.java
```diff
@@ -129,31 +132,51 @@ public PathProgramCache<L> getPathProgramCache() {
...
 	 * Constructs a {@link IRefinementStrategy} that can be used in conjunction with a {@link IRefinementEngine}.
+	 *
+	 * @param counterexample
+	 *            A trace that will be checked for feasibility and for which, if it is infeasible, a refinement result
+	 *            will be constructed.
+	 * @param controlConfigurationSequence
+	 *            Optionally, the sequence of control configurations visited by the trace in the program that is being
+	 *            verified. Used to judge the quality of proofs ("perfect") and for assert order modulation.
+	 * @param abstraction
+	 *            TODO ???
```
danieldietsch: Various strategies require the initial abstraction, e.g., to extract the complete alphabet, or to perform more complex generalizations.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/StrategyFactory.java
```diff
@@ -129,31 +132,51 @@ public PathProgramCache<L> getPathProgramCache() {
 
 	/**
 	 * Constructs a {@link IRefinementStrategy} that can be used in conjunction with a {@link IRefinementEngine}.
+	 *
+	 * @param counterexample
+	 *            A trace that will be checked for feasibility and for which, if it is infeasible, a refinement result
+	 *            will be constructed.
+	 * @param controlConfigurationSequence
+	 *            Optionally, the sequence of control configurations visited by the trace in the program that is being
```
danieldietsch: I don't like configurations here (or in the other cases). A configuration of, e.g., a nested word automaton is IIRC the location and the state of the stack. We are only interested in the location (because we can reconstruct, e.g. the stack from the counterexample / word).

danieldietsch (changes requested):
I did not find the place where we really see a benefit from the removal of IRun. Can you point me to it so that I can have a closer look?

In contrast, I see many places that become more convoluted and difficult to understand, in particular because word and sequence of control locations is now separated. 
This leads to passing null or returning pairs, having even more parameters, and generally obfuscating the relation between things.  
It was a major concern for me to not let that happen. 

Please introduce some wrapper class that combines both and perhaps provides helpful abstractions. In contrast to IRun, it can make its control location sequence optional, but then we should ensure that we fail early, i.e., that strategy modules that require control sequences fail if they are not provided with them. 

It would also be nice if we could reduce these pesky casts even further, but I am not sure that this is easy.

Heizmann:
> I did not find the place where we really see a benefit from the removal of IRun. Can you point me to it so that I can have a closer look?

TraceChecks were made for for sequences of IActions. Perfect Interpolants and Assert Orders and an additional bonus that are sometimes desired and that require and additional sequence of control configurations.
* We must separate the sequence of IActions from the sequence of control configurations in order to maintain support for applications that provide only a sequence of IActions.
* A *run* is a concept from automata theory. It is misleading to utilize a object with this name for other applications. (Would you use ComplexNumber for storing temperature and humidity only because the ComplexNumber can store two reals?)

> In contrast, I see many places that become more convoluted and difficult to understand, in particular because word and sequence of control locations is now separated. This leads to passing null or returning pairs, having even more parameters, and generally obfuscating the relation between things. It was a major concern for me to not let that happen.
> 
> Please introduce some wrapper class that combines both and perhaps provides helpful abstractions. In contrast to IRun, it can make its control location sequence optional, but then we should ensure that we fail early, i.e., that strategy modules that require control sequences fail if they are not provided with them.

Our TraceChecks are work in progress. We have to try to keep them as simple and as modular as possible. Let us pass null for the sequence of control configurations if there is no sequence of control configurations.
Wrapper classes and additional abstractions will make our TraceChecks even more convoluted.

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/AnnotateAndAsserter.java
Heizmann: For concurrent programs we do not have sequences of locations but sequence of *SomethingElse*. We do not yet know what *SomethingElse* should be.
* One good choice for *SomethingElse* is a set of locations, one location per active thread.
* Another choice might be that we additionally states of a monitor automaton.

danieldietsch:
I am not arguing for keeping IRun. I am arguing for adding a new type that describes a trace and optional metadata.

trunk/source/Library-AcceleratedInterpolation/src/de/uni_freiburg/informatik/ultimate/lib/acceleratedinterpolation/AcceleratedInterpolation.java
maul-esel: I've introduced such a datatype and named it `Counterexample<L>` (other naming suggestions welcome).

trunk/source/Library-LassoRanker/src/de/uni_freiburg/informatik/ultimate/lassoranker/nontermination/FixpointCheck2.java
maul-esel: Fixed.

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/InterpolatingTraceCheck.java
maul-esel: Fixed.

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/TraceCheck.java
maul-esel: This method is only used from PEAToBoogie's `VerificationResultTransformer::reduceRtInconsistencyProgramExecution`.

trunk/source/PEAtoBoogie/src/de/uni_freiburg/informatik/ultimate/pea2boogie/VerificationResultTransformer.java
maul-esel: The control configurations that are the subject of this PR should not be used for the purpose of computing executions. Since they have type `Object`, the only thing that should be done for them is to do equality-comparisons (e.g. to detect loops when configurations repeat).

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
maul-esel: No. The states of the `IRun` should not be used for this (see second bullet point under "Motivation" above).

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
maul-esel: The point was to not hardcode such logic deep inside a trace check, which has no idea what the control configurations of the original program look like, and thus did the wrong thing for some of our CEGAR loops.

On the other hand, the CEGAR loops know about the control configurations of the program they are verifying, so they can provide sensible, specialized implementations.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
maul-esel: The previous method is only used by our _danger invariant_ support. It requires a list of `IcfgLocation`s and only works for sequential programs.

This is the method that should be used for trace checks. It does not return a list of `IcfgLocation`s but of arbitrary objects -- in order to support both sequential and concurrent programs, we use a list of `Set<IcfgLocation>`.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/SifaRunner.java
maul-esel: I do not know anything about our SIFA implementation, so I tried to leave it as it is. In particular, this class does something more complex with `locationOfInterestList`, and I do not know how to map that.

As SIFA so far only works for sequential programs generally, I think generalizing this is not high-priority.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/acceleratedtracecheck/AcceleratedTraceCheck.java
maul-esel: Because it is meant to be optional. Do you think it should be required in this class?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/concurrency/CegarLoopForPetriNet.java
maul-esel: `pnRun.getStateSequence()` already has to satisfy this invariant. Since I am only mapping the sequence entry-by-entry, the length does not change.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/errorabstraction/ErrorAutomatonBuilder.java
maul-esel: The type `NestedFormulas` (superclass of `DefaultTransFormulas`) is used in the assert orders. So in order to use the correct control configurations for the assert orders that rely on the control flow, I am passing them into the `NestedFormulas` (now via a `Counterexample<L>` object that replaces the `Word<L>` previously stored there).

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/StrategyFactory.java
maul-esel: Ah, interesting, this is supposed to be the initial abstraction? I thought it should be the current abstraction. Good to know, I'll document it :+1:

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/StrategyFactory.java
maul-esel: See Matthias' comment and the discussion about the naming above.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/Mcr.java
schuessf: I don't quite remember this, but here `StrategyModuleMcr::getResult` is called, which calls `StrategyFactory::constructStrategy` with the same counterexample, so I think we should not pass null here, but the whole counterexample.

maul-esel: The `counterexample` here is a `NestedRun` returned from the emptiness check of the must-happens-before automaton (`mhbAutomaton`). Since it is not a run of the program, I did not know which sequence of control configurations should be used here.

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/AnnotateAndAsserter.java
maul-esel: To add to this: In GemCutter, the _SomethingElse_ has yet another type: it is a pair of a set of locations (one location per active thread) and a sleeo set (a set of letters).

danieldietsch: `configuration` just seems awfully close to the mixing of concepts we want to avoid. 
For GemCutter, I would argue that the pair of sets of location and the set of letters is the location of the explicit product automaton, so here `location` would be ok. I think that `location` is in general closer to the "something" we want to describe than `configuration`. But right now I don't know a better name.

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/TraceCheck.java
danieldietsch: Ok, this application needs this parameter combination.

trunk/source/PEAtoBoogie/src/de/uni_freiburg/informatik/ultimate/pea2boogie/VerificationResultTransformer.java
danieldietsch: Hm, so how are we going to compute rcfg computations in the future?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
danieldietsch: Ok, that makes perfect sense. Nevertheless, why is it not a `NestedRun<L, ISLPredicate>`?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
danieldietsch: Hm, I still dont understand it, in particular in combination with the implementation in NwaCegarLoop, which does exactly what you are saying it should not do.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
danieldietsch: Could we make that explicit by adding type params for this method?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/SifaRunner.java
danieldietsch: Nevertheless, for consistency, it should use the sequence of control locations (although it will fail for any sequence that is different from a sequence of program locations).

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/acceleratedtracecheck/AcceleratedTraceCheck.java
danieldietsch: Yes, I think it should be required in this class.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/concurrency/CegarLoopForPetriNet.java
danieldietsch: Ah yes, I see it now. I did misunderstand the code the frist time.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/Mcr.java
danieldietsch: As it is now passed as `Counterexample` it should be ok anyways.

trunk/source/PEAtoBoogie/src/de/uni_freiburg/informatik/ultimate/pea2boogie/VerificationResultTransformer.java
maul-esel: Can you explain how the computation of ICFG executions is connected to this? I don't see it.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/interpolantautomata/builders/AbsIntNonSmtInterpolantAutomatonBuilder.java
```diff
@@ -95,24 +96,19 @@ private NestedWordAutomaton<LETTER, IPredicate> getPathProgramAutomatonNew(
 			final INwaOutgoingLetterAndTransitionProvider<LETTER, IPredicate> oldAbstraction,
 			final IPredicateUnifier predicateUnifier, final IEmptyStackStateFactory<IPredicate> emptyStackFactory) {
 		mLogger.info("Creating interpolant automaton from AI using abstract post for generalization");
-
-		final NestedRun<LETTER, IPredicate> cex = (NestedRun<LETTER, IPredicate>) mCurrentCounterExample;
-		final List<IPredicate> stateSequence = cex.getStateSequence();
-
-		if (stateSequence.size() <= 1) {
+		if (mControlConfigurationSequence.size() <= 1) {
```
danieldietsch: Can this still happen?

danieldietsch (approved):

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
maul-esel: One should not use the states of the difference automaton _verbatim_, because they contain the interpolant automaton states as well as the program control flow.

The point of this method is to extract only the part of the run state sequence that reflects the control flow in the program, and throw away everything else.
- The implementation in `NwaCegarLoop` computes the set of all Icfg locations of all active threads at each point in the run.
  + Previously, the trace checks implemented something else: they took just one of the Icfg locations at each point, from one of the threads.
 - In `CegarLoopForPetriNet`, the computation is based on the markings of the Petri net, etc.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
maul-esel: Because the `mCounterexample` run created by `NwaCegarLoop` is of type `NestedRun<L, IPredicate>` (as set by `::isAbstractionEmpty`).

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
maul-esel: I don't think so, the computed lists are too different.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/SifaRunner.java
maul-esel: This would require some changes in the callers (the CEGAR loops). Right now, `NwaCegarLoop` always passes a `List<Set<IcfgLocation>>` for the control configurations. We would have to change it such that
- for sequential programs, it passes a `List<IcfgLocation>`;
- and for concurrent programs, it passes a `List<Set<IcfgLocation>>` or `List<IcfgLocation[]>`

I'll look into it. Then we could also perhaps restructure these two similar methods in `NwaCegarLoop` that you remarked on.

maul-esel: On second thought [first reponse deleted], I do not think that is a good idea. It will break the contract that the control configurations in a `Counterexample<L>` are only used in equality comparisons.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/acceleratedtracecheck/AcceleratedTraceCheck.java
maul-esel: Done.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
maul-esel: I have restructured the code slightly, so that for sequential programs, the same method is reused; and concurrent programs are handled separately.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/interpolantautomata/builders/AbsIntNonSmtInterpolantAutomatonBuilder.java
maul-esel: It is still possible to have a `Counterexample<L>` with the empty word, and exactly 1 state.

Of course, this class should never be used for such a counterexample, but to detect it if it is seems exactly the point of the check here.

trunk/source/PEAtoBoogie/src/de/uni_freiburg/informatik/ultimate/pea2boogie/VerificationResultTransformer.java
danieldietsch: You are right, its actually not needed. I thought it was when I was looking at TraceCheck recently, but I was wrong

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
danieldietsch: :+1:

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
danieldietsch: :) -- and we cant change that because...?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/SifaRunner.java
danieldietsch: It might be worthwhile to think about this use case. 

I don't see why we restricting ourselves to `Object` in `Counterexample<L>`. What is preventing us from having a `Counterexample<L,C>`  where `C` is the type of control location? 

Nevertheless, it is fine here either way.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/interpolantautomata/builders/AbsIntNonSmtInterpolantAutomatonBuilder.java
danieldietsch: Hm, would it make sense to forbid this in `Counterexample` instead? Here, at least one should check if `cex` has a `getStateSequence``to avoid the NPE.

maul-esel: We would not get an NPE: The call to `getControlConfigurations()` in the constructor of this class would `throw new IllegalStateException("Counterexample with control configurations is required, but this counterexample does not have control configurations.");`.

Nevertheless, it does seem reasonable to only allow non-empty `Counterexample`s. I will add such a check.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
maul-esel: Because the type of states is not necessarily `ISLPredicate` -- this CEGAR loop can also be used with concurrent programs, just not the danger invariant support (to which this method belongs).

danieldietsch: Yes, so you are saying, we would have to propagate the type too far up?

maul-esel: Yes. Basically, moving it up would replace the (checked) cast here with an unchecked cast further up, which I want to avoid.

Jenkinsfile
```diff
@@ -11,6 +11,10 @@ pipeline {
   environment {
     PATH = "${env.WORKSPACE}/releaseScripts/default/adds:${env.PATH}"
   }
+  tools {
```
schuessf: This change is already on dev, right?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/Mcr.java
```diff
@@ -102,7 +102,8 @@ private McrTraceCheckResult<L> exploreInterleavings(final List<L> initialTrace)
 		while (counterexample != null) {
 			mLogger.info("---- MCR iteration " + iteration + " ----");
 			iteration++;
-			result = mResultProvider.getResult(counterexample);
+			// TODO Is it correct to not pass any control locations here?
```
schuessf: I think it is correct. Just add a comment that states this.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/StrategyFactory.java
```diff
@@ -235,19 +255,29 @@ public class StrategyModuleFactory {
...
 
+		/**
+		 * @param counterexample
+		 *            A trace that will be checked for feasibility and for which, if it is infeasible, a refinement
+		 *            result will be constructed.
+		 * @param controlConfigurationSequence
+		 *            Optionally, the sequence of control configurations visited by the trace in the program that is
+		 *            being verified. Used to judge the quality of proofs ("perfect") and for assert order modulation.
+		 * @param abstraction
+		 *            TODO ???
```
schuessf: Why is this TODO here?

schuessf (approved):
Thanks for your work. I don't have much to add here, just some minor comment.

Jenkinsfile
maul-esel: Yes, just cherry-picked here for the Nightly build. I'll squash when I merge the PR.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/StrategyFactory.java
maul-esel: Because I wanted to document this parameter. I did so for the `constructStrategy` methods above, but forgot it here. Fixed now, and I also updated the rest of the documentation block for the removal of the `controlConfigurationSequence` parameter.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/Mcr.java
maul-esel: :+1: done

PR #672, opened 2024-08-10
Migrate Ultimate to Java 21 and update framework components

bahnwaerter:
This series of patches migrates Ultimate to Java 21 and updates the framework components to the following versions:

  - Eclipse 4.32
  - Eclipse CDT 11.6
  - Jetty 12

The patches also improve the build process by unifying build automation for Ant tasks in Eclipse and Maven builds. Specific Ant settings and the installation of the Tycho plugin in the Eclipse IDE are no longer required for building Ultimate. In addition, dependencies from the P2 repositories are resolved automatically. Therefore, only one PDE target platform file is now required for all supported target architectures.

PR #694, opened 2024-11-15
Use external dependencies from Maven

schuessf:
Currently, there are many different ways how external dependencies are used in Ultimate:
* as proper Maven dependencies in the [target file](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/BA_SiteRepository/Ultimate_E4.32_Java21.target#L62) (added by @bahnwaerter in #672)
* using our own P2 repo (see [here](https://struebli.informatik.uni-freiburg.de/p2-repository-ultimate-dependencies-21/plugins/)), e.g. `org.yaml.snakeyaml`
* as cloned source files (e.g. `Library-ApacheCommonsLang`)
* as JAR files (e.g. `trove-3.0.3.jar` in `Library-Automata`)

However, I presume the first solution is the most optimal one, as one can simply add external dependencies without any changes on some servers (like the second solution) or without copying any files into the source files of Ultimate that can be also easily updated (just change the version number in the target file). Therefore I attempted to only use Maven dependencies, where possible, there are currently some exceptions:
* `Library-Jung` and `Library-ApacheCommonsCLI` were patched to our needs (e.g. 8b6c0bf), so we cannot replace them with external dependencies directly.
* ~~While Eldarica is available on Maven, `Library-CHC/Eldarica-assembly-2.0.8.jar` contains actually more dependencies.~~
* `BA_SharedJARs` contains multiple JARs, but I am not sure, if they can be replaced with Maven dependencies or where these JARs are actually needed.

~~There is still one open issue with this PR: Building works totally fine for me using `Ultimate_E4.32_Java21.target` as a target file, however if I use `Ultimate_E4.32_Java21_Win32.target` the external dependenices cannot be resolved (e.g. `org.apache.commons.lang3`), even though `Ultimate_E4.32_Java21.target` is included in the target file.~~ (This was fixed by adding the dependencies to `includeBundles`)


trunk/source/BA_SiteRepository/Ultimate_E4.32_Java21.target
```diff
@@ -5,11 +5,6 @@
 		<vmArgs>-ea -Xms4M -Xmx4G</vmArgs>
 	</launcherArgs>
 	<locations>
-		<location includeAllPlatforms="true" includeConfigurePhase="true" includeSource="false" type="InstallableUnit">
-			<repository id="ultimate-dependencies" location="https://struebli.informatik.uni-freiburg.de/p2-repository-ultimate-dependencies/"/>
-			<unit id="com.lmax.disruptor" version="3.3.5"/>
-			<unit id="org.yaml.snakeyaml" version="2.2.0"/>
-		</location>
```
danieldietsch, line 12: Can you leave the location but without units? So we still remember its there, just in case :)

trunk/source/Library-Automata/META-INF/MANIFEST.MF
```diff
@@ -9,7 +9,8 @@ Require-Bundle: de.uni_freiburg.informatik.ultimate.lib.core,
  de.uni_freiburg.informatik.ultimate.smtinterpol,
  de.uni_freiburg.informatik.ultimate.lib.smtlib,
  de.uni_freiburg.informatik.ultimate.lib.smtlibutils,
- de.uni_freiburg.informatik.ultimate.lib.util
+ de.uni_freiburg.informatik.ultimate.lib.util,
+ wrapped.net.sf.trove4j.trove4j
```
danieldietsch, line 13: Is trove actually used?

trunk/source/Library-Automata/build.properties
```diff
@@ -2,6 +2,5 @@ source.. = src/
 output.. = bin/
 bin.includes = .,\
                META-INF/,\
-               src/de/uni_freiburg/informatik/ultimate/automata/petrinet/petruchio/petruchio.jar,\
-               src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/optncsb/trove-3.0.3.jar
+               src/de/uni_freiburg/informatik/ultimate/automata/petrinet/petruchio/petruchio.jar
```
danieldietsch, line 5: What about petruchio.jar? Is this not available from Maven? Would this be a candidate for packaging in our p2 repo?

trunk/source/Library-PEA/META-INF/MANIFEST.MF
```diff
@@ -9,7 +9,8 @@ Require-Bundle: de.uni_freiburg.informatik.ultimate.lib.core,
  net.sourceforge.czt,
  de.uni_freiburg.informatik.ultimate.boogie.preprocessor,
  de.uni_freiburg.informatik.ultimate.lib.boogieast,
- de.uni_freiburg.informatik.ultimate.lib.util
+ de.uni_freiburg.informatik.ultimate.lib.util,
+ wrapped.xerces.xercesImpl
```
danieldietsch: Would it also work with Import-Package?

danieldietsch:
* Did you per chance compare the size of our deployments before/after this change?
* Would it work with the platform-dependent target files if you added the new libraries as included bundles in the platform-dependent files?

danieldietsch (approved):

trunk/source/Library-Automata/META-INF/MANIFEST.MF
schuessf: Yes, see [here](https://github.com/ultimate-pa/ultimate/blob/0a57b278650c6e4f7c1dc59465c12955e1a35a6e/trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/optncsb/automata/AccGenBuchi.java#L33). I am however not sure, if this automaton operation is _actually_ used 😉

trunk/source/Library-Automata/build.properties
schuessf: Yes, it is not available on Maven (it is pretty old). But if we can use our P2 repo for such JARs, I am absolutely in favour of this.

schuessf: In that case, we can also store the Eldarica JAR on our P2 (or figure out what depencies it actually contains).

schuessf:
> Did you per chance compare the size of our deployments before/after this change?

It is slightly smaller, e.g. 117MB instead of 119MB for the Debug-E4

> Would it work with the platform-dependent target files if you added the new libraries as included bundles in the platform-dependent files?

Yes, it does work, see 171987b0f4a3b13e258863858c6e07799ed3c1e2.

trunk/source/Library-Automata/build.properties
danieldietsch: But this can be a separate PR; it just would be nice to get rid of these "hidden" jars.

trunk/source/BA_SiteRepository/Ultimate_E4.32_Java21.target
schuessf: If we plan to use it later on (for JARs that are not available on Maven), this makes sense 😉

trunk/source/BA_SiteRepository/Ultimate_E4.32_Java21.target
```diff
@@ -60,6 +58,24 @@
...
+					<artifactId>disruptor</artifactId>
+					<version>3.4.4</version>
+					<type>jar</type>
+				</dependency>
+				<dependency>
+					<groupId>net.sf.trove4j</groupId>
+					<artifactId>trove4j</artifactId>
+					<version>3.0.3</version>
+					<type>jar</type>
+				</dependency>
```
maul-esel, line 84: So I guess it turned out this dependency is still used after all?

maul-esel (approved):
:+1: thanks for taking care of this.

I agree we should keep pushing on this in further PRs for the other JARs, and perhaps we can also adapt some code such that we do not need a patched version e.g. of `ApacheCommonsCLI`?

danieldietsch:
> [...] and perhaps we can also adapt some code such that we do not need a patched version e.g. of `ApacheCommonsCLI`?

For ApacheCommonsCLI rather not. We have many modifications to that part of Apache Commons, which we use for our command line interface. Even moving it to a separate repo would be counter-productive, as changing and testing that when, e.g., we add a new option type, would be much more tedious.

trunk/source/BA_SiteRepository/Ultimate_E4.32_Java21.target
schuessf: Yes, it is used, e.g. [here](https://github.com/ultimate-pa/ultimate/blob/0a57b278650c6e4f7c1dc59465c12955e1a35a6e/trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/optncsb/automata/AccGenBuchi.java#L33).

schuessf:
> I agree we should keep pushing on this in further PRs for the other JARs

I agree, this PR was just meant as a start, where I tried to simply use unmodified Maven dependencies where possible, without changing any code. There are still some dependencies left, where this is not possible, because they are not on maven ([petruchio.jar](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/petruchio/petruchio.jar), [czt_1_5_0_bin.jar](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/Library-CZT/czt_1_5_0_bin.jar), `JavaCup`) or because we use some patched version (`Library-Jung`). Maybe we could simply add them to our P2 repository in a future PR.

schuessf:
> > [...] and perhaps we can also adapt some code such that we do not need a patched version e.g. of `ApacheCommonsCLI`?
> 
> For ApacheCommonsCLI rather not. We have many modifications to that part of Apache Commons, which we use for our command line interface. Even moving it to a separate repo would be counter-productive, as changing and testing that when, e.g., we add a new option type, would be much more tedious.

It looks like, we adapted `ApacheCommonsCLI` only in four commits (804d6ba, 2c29ed9, ff91644, 8b6c0bf). Instead of changing the source files, we could also use the unpatched versions and for example extend the classes there to our needs (if this is possible). But again: This is not really urgent 😉

maul-esel:
For CZT, we should talk to the ReqAnalyzer people -- when asking @Langenfeld, he was not sure whether it is still needed.

danieldietsch:
> > > [...] and perhaps we can also adapt some code such that we do not need a patched version e.g. of `ApacheCommonsCLI`?
> > 
> > 
> > For ApacheCommonsCLI rather not. We have many modifications to that part of Apache Commons, which we use for our command line interface. Even moving it to a separate repo would be counter-productive, as changing and testing that when, e.g., we add a new option type, would be much more tedious.
> 
> It looks like, we adapted `ApacheCommonsCLI` only in four commits ([804d6ba](https://github.com/ultimate-pa/ultimate/commit/804d6bad9bf664ce333005ad4406c11826169de9), [2c29ed9](https://github.com/ultimate-pa/ultimate/commit/2c29ed9ed40ab1586266dd340aa5d5b6aef23872), [ff91644](https://github.com/ultimate-pa/ultimate/commit/ff916444f2bbd6b65887aa4f5ef2c855cdbd7127), [8b6c0bf](https://github.com/ultimate-pa/ultimate/commit/8b6c0bffde0bc6a82263e503df0e1cb2074aa00d)). Instead of changing the source files, we could also use the unpatched versions and for example extend the classes there to our needs (if this is possible). But again: This is not really urgent 😉

My rule of thumb would be:
* Unmodified libraries with "good" autogenerated manifests should go in the target files.
* Unmodified libraries where we have to create a manifest by ourselves or those that are not available via Maven should go in the p2 dependency repo.
* Modified libraries should be part of our repo, because it simplifies development.

For `ApacheCommonsCLI` I am not sure if we want to deal with changes in upstream as long as it does what we want.

danieldietsch:
> For CZT, we should talk to the ReqAnalyzer people -- when asking @Langenfeld, he was not sure whether it is still needed.

CZT is only needed for Library-PEA, and there for the Z support. I am pretty sure that they dont use that, but they would need to remove it from Library-PEA. Question is: do they want to do that soonish, or would it be ok if someone else does it? 
It would be nice to get rid of it.
@Langenfeld @henkele ?

bahnwaerter (approved):
Thanks @schuessf for starting this work to beautify the code base and its dependencies. I particularly like that we can now integrate libraries like Trove cleanly.

bahnwaerter:
> > For CZT, we should talk to the ReqAnalyzer people -- when asking @Langenfeld, he was not sure whether it is still needed.
> 
> CZT is only needed for Library-PEA, and there for the Z support. I am pretty sure that they dont use that, but they would need to remove it from Library-PEA. [...]

That was also the statement of @jhoenicke that he only needed this back then, before the PEA-to-Boogie translation was available. Nowadays it is no longer used and it seems as if the CZT project is dead.

trunk/source/Library-CHC/META-INF/MANIFEST.MF
```diff
@@ -5,16 +5,18 @@ Bundle-SymbolicName: de.uni_freiburg.informatik.ultimate.lib.chc
 Bundle-Vendor: Ultimate Program Analysis Team
 Bundle-Version: 0.3.0
 Bundle-RequiredExecutionEnvironment: JavaSE-21
-Bundle-ClassPath: .,
- Eldarica-assembly-2.0.8.jar
+Bundle-ClassPath: .
```
bahnwaerter: Is this still necessary if Elderica is no longer included as a local `*.jar`?

trunk/source/BA_FeatureUltimateCommon/feature.xml
```diff
@@ -126,10 +126,6 @@
          id="com.github.jhoenicke.javacup"
          version="0.0.0"/>
 
-   <plugin
-         id="org.apache.commons.lang"
-         version="0.0.0"/>
-
```
bahnwaerter, line 132: It is correct to remove this here, but it should be added as a general dependency under `BA_FeatureDependenciesCommon/feature.xml`. Otherwise the products will be packaged incorrectly.

bahnwaerter (changes requested):

trunk/source/Library-CHC/META-INF/MANIFEST.MF
schuessf: Yes. it seems like it can be omitted

maul-esel:
IIRC the usage of eldarica via the JAR didn't really work; I think we were using the wrong API. In the end, we used eldarica via command line. Hence, the JAR dependency _might_ be obsolete and we _might_ just replace it with a binary in `releaseScripts/default/adds`.

trunk/source/BA_FeatureUltimateCommon/feature.xml
schuessf: Okay, that's somethint I don't quite understand: Do all the external dependencies belong to `BA_FeatureDependenciesCommon/feature.xml`? Currently this is not the case, none of the dependencies in the target-file are mentioned there.

And what do you meant with "packaged incorrectly"? I suppose that the SnakeYAML is currently working anyways in the SV-COMP 😉

schuessf:
> IIRC the usage of eldarica via the JAR didn't really work; I think we were using the wrong API. In the end, we used eldarica via command line. Hence, the JAR dependency _might_ be obsolete and we _might_ just replace it with a binary in `releaseScripts/default/adds`.

Currently, [EldaricaChcScript](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/Library-CHC/src/de/uni_freiburg/informatik/ultimate/lib/chc/eldarica/EldaricaChcScript.java) actually uses the API provided by the JAR (it imports `ap.*`, `lazabs.*`). But if we want to use the binary instead, I can also revert this change for Eldarica.

maul-esel:
Yes, that implementation is still there. Since it didn't work well, I introduced [EldaricaCliChcScript](https://github.com/ultimate-pa/ultimate/blob/413081681f424fe05429144cce149eb7144d0d9b/trunk/source/Library-CHC/src/de/uni_freiburg/informatik/ultimate/lib/chc/EldaricaCliChcScript.java) on a branch, but that branch needs some more work before it's merge-ready.

If the implementation on `dev` works with your changes, then let's keep them -- it's anyway better to have the dependency in maven rather than checking in the JAR. Once I get around to cleaning up & merging that other branch, we can see if we remove `EldaricaChcScript`, and then also remove these dependencies.

bahnwaerter (approved):

PR #699, opened 2024-12-03
Refactor assert orders

schuessf:
This PR aims to refactor the assert orders. Currently the situation is as follows:
* We have several assert orders (see `AssertCodeBlockOrderType`). `NOT_INCREMENTALLY` is handled in the class `AnnotateAndAsserter`, while all the others are handled in the subclass `AnnotateAndAsserterWithStmtOrderPrioritization` (that hardly makes use of the methods of the superclass).
* The assert orders in `AnnotateAndAsserterWithStmtOrderPrioritization` are just methods (e.g. `annotateAndAssertOutsideLoopFirst1`) that doe not only the partitioning of the trace, but also asserts the SSA directly and thus returns a `LBool`.

This situation is not nice for several reasons:
* `NOT_INCREMENTALLY` does not need any special handling, it can be also seen as an assert order that partitions the trace into a single partition.
* `AnnotateAndAsserterWithStmtOrderPrioritization` contains many methods that just belong to a single assertion order.
* We were thinking to combine two assertion order, which is not possible for now, as the methods are not really modular.

Therefore the following changes were made:
* Assert orders are now an implementation of the new interface `IAssertOrder` with a single methods that performs the partitioning, but not the assertion itself.
* Code that is shared between different assert orders was moved from `AnnotateAndAsserterWithStmtOrderPrioritization` to the new class `AssertOrderUtils`
* For `NOT_INCREMENTALLY` there exists also an implementation of `IAssertOrder`, `AssertOrderNotIncrementally`. Therefore `AnnotateAndAsserter` is no more required and thus was deleted and `AnnotateAndAsserterWithStmtOrderPrioritization` was renamed to `AnnotateAndAsserter`
* In `AnnotateAndAsserter` we first create an instance of the desired `IAssertOrder` in `getAssertOrder`, perform the partioning and then assert the partioned SSA in `annotateAndAssert` (which was previosuly performed by every assert order itself).

*Note*: I ran already the nightly build, but there was a small bug (connected to #692). I just fixed it and re-ran the nightly build.


trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/assertorders/AssertOrderSmtFeatureHeuristic.java
```diff
@@ -0,0 +1,175 @@
...
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.cfg.structure.IAction;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.tracecheck.ITraceCheckPreferences.SmtFeatureHeuristicPartitioningType;
+import de.uni_freiburg.informatik.ultimate.lib.smtlibutils.solverbuilder.SMTFeatureExtractionTermClassifier;
+import de.uni_freiburg.informatik.ultimate.lib.smtlibutils.solverbuilder.SMTFeatureExtractionTermClassifier.ScoringMethod;
+import de.uni_freiburg.informatik.ultimate.lib.tracecheckerutils.singletracecheck.NestedFormulas;
+import de.uni_freiburg.informatik.ultimate.logic.Term;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.Triple;
+
+/**
+ * @author Betim Musa (musab@informatik.uni-freiburg.de)
```
danieldietsch: Could you check if the SMTFeatureHeuristic was really by Betim? I remember that Julian Loeffler implemented that ; perhaps he just forgot to add himself as author. I believe that was 1e8a6187ee..fc59022d9a on AnnotateAndAsserterWithStmtOrderPrioritization.java

danieldietsch (approved):
Looks very good and is a really nice change!

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/assertorders/AssertOrderSmtFeatureHeuristic.java
schuessf: Good point! I just used the same author for every `IAssertOrder` as in `AnnotateAndAsserterWithStmtOrderPrioritization`, but I can check the blame, who is actually the author.

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/assertorders/AssertOrderUtils.java
```diff
@@ -0,0 +1,120 @@
...
+		final Map<Integer, Set<Integer>> depth2Statements = new HashMap<>();
+
+		dfsPartitionStatementsAccordingToDepth(0, trace.length(), 0, rwt, depth2Statements, pps);
+
+		return depth2Statements;
+	}
+
+	/**
+	 * Partition the statement positions between lowerIndex and upperIndex according to their depth. (See documentation
+	 * for the meaning of 'depth'). The result is stored in the map 'depth2Statements'. The partitioning is done
```
maul-esel: Where is this documentation regarding the meaning of "depth"?

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/AnnotateAndAsserter.java
```diff
@@ -26,114 +28,170 @@
...
 			} else {
-				if (mTrace.isReturnPosition(i) && mTrace.isPendingReturn(i)) {
-					pendingReturnPositions.add(i);
-				}
 				mAnnotSSA.setFormulaAtNonCallPos(i, mAnnotateAndAssertCodeBlocks.annotateAndAssertNonCall(i));
 			}
 		}
 
-		assert callPositions.containsAll(mTrace.getCallPositions());
-		assert mTrace.getCallPositions().containsAll(callPositions);
```
maul-esel, line 106: Did you intend to remove these checks?

maul-esel (approved):
Nice :)

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/AnnotateAndAsserter.java
schuessf: Oh, that's interesting: These checks were not present in `AnnotateAndAsserterWithStmtOrderPrioritization` (which is the basis for the new `AnnotateAndAsserter`). Therefore I also removed the variable `callPositions`, because it was unused. But I think it does not hurt to add these checks again.

schuessf: Wait, these assertions state that all calls of the trace are already asserted. This is only reasonable, if the whole SSA is asserted here. We could keep the second assertion, but it would be trivially satisfied, since for all `i`, if `trace.isCallPosition(i)` is true, then `i` is element of `trace.getCallPositions()`. So I think it is reasonable to remove these assertions.

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/assertorders/AssertOrderUtils.java
schuessf: Should be fixed, this part was removed and some documentation added to `partitionStatementsAccordingDepth`.

bahnwaerter (approved):

PR #700, opened 2024-12-06
Apply CC fixes to existing code

schuessf:
During the update to Java 21 (see #672), we also slightly changed our coding conventions.
This can lead to unnecessary diffs, when editing existing files (many existing file did not even follow our old coding conventions...).
In this PR I applied the new coding conventions automatically (starting with `CACSL2BoogieTranslator`) and we can discuss possible changes to the coding conventions here.

---

Eclipse does not actually allow to apply the save actions automatically to multiple files. But I found a way 😉
* There is a similar command `Clean Up` that can be applied on a whole project under the `Source` menu (Note: There is also a command `Format` there, but it only applies the formatter).
* The `Clean Up` actions can be set up under `Preferences` -> `Java` -> `Code Style` -> `Cleanup` and they are almost the same as the save actions, but their is no common way to set them.
* We could also add these cleanup settings to our settings that the users should import, but we don't want to keep them in sync manually and also the import of the cleanup actions did not work.
* Therefore I wrote a small python script (see below) that creates an `.xml` file that can be imported under the cleanup-settings from our `.epf` file.
* After this file is imported, you can simply select some files / plugins / the whole workspace to run `Source` -> `Clean up`, which should apply our save actions on multiple files automatically.

```py
def create_cleanup_settings(in_file, out_file):
    lines = ['<?xml version="1.0" encoding="UTF-8" standalone="no"?>',
             '<profiles version="2">',
             '\t<profile kind="CleanUpProfile" name="Ultimate" version="2">']
    with open(in_file) as f:
        settings = re.findall(r'/instance/org.eclipse.jdt.ui/sp_cleanup.(.*?)=(.*?)\s', f.read())
    for n, s in settings:
        lines.append(f'\t\t<setting id="cleanup.{n}" value="{s}"/>')
    lines.append('\t</profile>')
    lines.append('</profiles>')
    with open(out_file, "w") as f:
        print("\n".join(lines), file=f)
```


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/CLocation.java
```diff
@@ -188,19 +188,17 @@ private IASTNode getMergedNode(final CLocation otherCloc) {
 			return otherNode;
 		} else if (otherNode == null) {
 			return myNode;
+		} else // we have two nodes and want to merge them; if one of both is a translation unit, we take the other
```
schuessf: Merging else and if is a good idea in general, but it looks somehow weird, if there is a comment in between, but I don't know how to handle this.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/PostProcessor.java
```diff
@@ -222,10 +222,10 @@ public ArrayList<Declaration> postProcess(final ILocation loc, final IASTNode ho
 
 			mLogger.info("Analyzing all entry points");
 			if (mProcedureManager.hasProcedure(SFO.MAIN)) {
-				final String msg =
-						"You selected the library mode (i.e., each procedure can be starting procedure and global "
-								+ "variables are not initialized). This program contains a \"main\" procedure. Maybe you "
-								+ "wanted to select the \"main\" procedure as starting procedure.";
+				final String msg = """
```
schuessf: That's really a nice usage of multiline string, we were able to get rid of the quotes `\"`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/container/c/CType.java
```diff
@@ -186,14 +188,7 @@ public boolean isVoidType() {
...
 	public int hashCode() {
-		final int prime = 31;
-		int result = 1;
-		result = prime * result + (mIsConst ? 1231 : 1237);
-		result = prime * result + (mIsExtern ? 1231 : 1237);
-		result = prime * result + (mIsInline ? 1231 : 1237);
-		result = prime * result + (mIsRestrict ? 1231 : 1237);
-		result = prime * result + (mIsVolatile ? 1231 : 1237);
-		return result;
+		return Objects.hash(mIsConst, mIsExtern, mIsInline, mIsRestrict, mIsVolatile);
```
schuessf, line 191: That's so much nicer, `Boolean::hashCode` returns exactly `1231` or `1237` and `Objects::hash` computes exactly this combination.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/preferences/CACSLPreferenceInitializer.java
```diff
@@ -61,11 +62,12 @@ public class CACSLPreferenceInitializer extends UltimatePreferenceInitializer {
 	public static final String LABEL_CHECK_MEMORY_LEAK_IN_MAIN =
 			"Check for the main procedure if all allocated memory was freed";
 	public static final String LABEL_SVCOMP_MEMTRACK_COMPATIBILITY_MODE = "SV-COMP memtrack compatibility mode";
-	public static final String DESC_SVCOMP_MEMTRACK_COMPATIBILITY_MODE = "Report UNKNOWN instead of UNSAFE if not all "
-			+ "allocated memory was freed at the end of the main procedure. Rationale: at the SV-COMP we have to check "
-			+ "if the program lost track of allocated memory. If this is set to false we are unsound (at SV-COMP) in "
-			+ "cases where not all memory is freed but pointers to that memory are live at the end of the "
-			+ "main procedure.";
+	public static final String DESC_SVCOMP_MEMTRACK_COMPATIBILITY_MODE = """
```
schuessf: While I think these multiline Strings are generally nicer, I don't think we should change existing code, as also the line breaks are somehow strange now.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/CACSL2BoogieBacktranslator.java
```diff
@@ -747,7 +748,7 @@ public IBacktranslatedCFG<String, CACSLLocation> translateCFG(final IBacktransla
 		// printCFG(cfg, mLogger::info);
 		final boolean oldValue = mGenerateBacktranslationWarnings;
 		mGenerateBacktranslationWarnings = false;
-		IBacktranslatedCFG<String, CACSLLocation> translated = translateCFG(cfg, (a, b, c) -> translateCFGEdge(a, b, c),
+		IBacktranslatedCFG<String, CACSLLocation> translated = translateCFG(cfg, (IFunction<Map<IExplicitEdgesMultigraph<?, ?, String, ? extends BoogieASTNode, ?>, Multigraph<String, CACSLLocation>>, IMultigraphEdge<?, ?, String, BoogieASTNode, ?>, Multigraph<String, CACSLLocation>, Multigraph<String, CACSLLocation>>) this::translateCFGEdge,
```
schuessf: This change is really bad! In general, `this::translateCFGEdge` is a bit nicer than `(a, b, c) -> translateCFGEdge(a, b, c)`, but the save actions introduced an unncessary cast here and for long class names `C`, I actually prefer `c -> f(c)` over `C::f`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/FunctionTableBuilder.java
```diff
@@ -65,13 +65,11 @@ public int visit(final IASTDeclaration declaration) {
...
-					// we only update the table with a declaration, if there is no entry for that name yet.
-					// otherwise we might only keep the declaration and omit the implementation from
-					// reachableDeclarations.
-					if (!mFunctionTable.containsKey(rslvKey)) {
-						mFunctionTable.put(rslvKey, d);
-					}
+				// we only update the table with a declaration, if there is no entry for that name yet.
+				// otherwise we might only keep the declaration and omit the implementation from
+				// reachableDeclarations.
+				if ((d instanceof IASTFunctionDeclarator) && !mFunctionTable.containsKey(rslvKey)) {
```
schuessf, line 71: Oddly enough, Eclipse inserts unnecessary parentheses here that would be removed again, if we run the save actions afterwards again 😆

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -134,7 +134,7 @@ public BacktranslatedExpression translateEnsuresExpression(
 
 	private Optional<Expression> computeOldVarEqualities(final String proc,
 			final Set<de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression> modifiableGlobals) {
-		final var modifiableNames = modifiableGlobals.stream().map(e -> e.getIdentifier()).collect(Collectors.toSet());
+		final var modifiableNames = modifiableGlobals.stream().map(de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression::getIdentifier).collect(Collectors.toSet());
```
schuessf: `e -> e.getIdentifier())` is definitely nicer than `de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression::getIdentifier`...
Also, Eclipse does not respect the maximum line length here.

schuessf: Maybe even without this change, we should run `Clean up` twice.

danieldietsch: Up to the fixpoint :)

Heizmann:
The current rules remove variables that are seemingly unnecessary as for example in the following example.
```
Term fewerQuantifiers = myComplexMethod(arg0, arg1);
return fewerQuantifiers;
```
However these variables are not unnecessary they 
* serve as documentation and 
* they ease debugging. When I step through the code, I can immediately see the value of this variable.

bahnwaerter:
> However these variables are not unnecessary they
> * serve as documentation and
> * they ease debugging. When I step through the code, I can immediately see the value of this variable.

Such measures tend to indicate an unfavorable naming of a function or its excessive complexity. In this case, one should rather give the function a meaningful name or reduce its complexity. To debug the return value of a function, you can use watch expressions in the Eclipse debugger. Watch expressions are also useful for expressions like `foo.getBar().equals("xyz");` where you only want to inspect `foo.getBar()` without rewriting the code for debugging.

Heizmann:
> Such measures tend to indicate an unfavorable naming of a function or its excessive complexity. In this case, one should rather give the function a meaningful name or reduce its complexity. To debug the return value of a function, you can use watch expressions in the Eclipse debugger. Watch expressions are also useful for expressions like `foo.getBar().equals("xyz");` where you only want to inspect `foo.getBar()` without rewriting the code for debugging.

Sounds plausible but
* You may not be able to modify the methods that you call.
* The method name refers to its capabilities in general, but your return value may have additional properties that the method does not ensure.
* In practice, watching expressions is (for a human) more difficult and time-consuming than inspecting values.
* You cannot watch an expression that has side effects without changing the system's state.

danieldietsch:
> ... watch expressions ...

Newer Eclipse versions also display the return value of a called function without a watch expression. The only downside is that you see it only when returning from the function, i.e., not anymore in the next step (as far as I know). 
As the next step is another function (after all, its a return), this again should never be a problem: either the other function also returns this value immediately, or it does something with it (and can then save the value in a local variable if this is appropriate for debugging/readability or necessary for other reasons. 

> You may not be able to modify the methods that you call.

In Ultimate, you are probably always able to rename the methods you are calling. 
 

I think in many circumstances it is helpful and increases readability if you use appropriately named local variables instead of inlining expressions. If the return of your method is itself a method call, however, then the description of that value should already be clear from the method you are in (not the one you are calling), because this is now the value you are returning. If it is not clear from the caller name, the callers docs, the callees name or the callees docs, it is probably more appropriate to add a line comment there, as this value is not some intermediate result but _the_ result of the function.

Heizmann:
> Newer Eclipse versions also display the return value of a called function without a watch expression. The only downside is that you see it only when returning from the function, i.e., not anymore in the next step (as far as I know). As the next step is another function (after all, its a return), this again should never be a problem: either the other function also returns this value immediately, or it does something with it (and can then save the value in a local variable if this is appropriate for debugging/readability or necessary for other reasons.

The question is not only what is doable, but what is efficiently doable. Debugging is time-consuming, we have to make sure that developers can debug efficiently. (E.g., maybe you want to set a conditional breakpoint that puts the return value of the method in relation to other values)

danieldietsch:
Yes, it seems that Eclipse does not support referencing return values in breakpoint conditions.

schuessf:
After yesterday's discussion, I changed the coding conventions to make them slightly less strict (see 4661161; incl. variables before returns) and re-applied the coding conventions to `CACSL2BoogieTranslator`.

schuessf:
@danieldietsch @maul-esel @Heizmann @bahnwaerter What is your opinion on this PR? Do you agree with the applied coding conventions? And what is the best way to apply them to our project?

danieldietsch:
The best way to apply is unclear to me -- there was some project that could apply Eclipse CCs from the command line, but not all. Perhaps one could write a small eclipse plugin that opens all Java files and applies the conventions, then we had one commit. 
The alternative is: just setting it, and applying whenever changes are made (preferably with a separate commit). 

Wrt. to the convetions themselves: I like them :)

schuessf:
> The best way to apply is unclear to me -- there was some project that could apply Eclipse CCs from the command line, but not all. Perhaps one could write a small eclipse plugin that opens all Java files and applies the conventions, then we had one commit.

I think it is only possible to apply the formatter from the command line, but not the save actions. But it is possible to run `Clean up` from Eclipse (on selected files/plugins).

So, should we have a single huge commit with the CCs, some smaller ones (e.g., for each plugin), or do we just apply the CCs when changing the corresponding code? And if we decide to apply the CCs on the whole project, there are possibly some plugins that we don't want to be included (e.g., external dependencies like `JavaCup`).

danieldietsch:
If we can apply it in one huge commit, we should do so (and then add it to something like ``.git-blame-ignore-revs``)
```
❯ rg -i blame  ~/.gitconfig
73:[blame]
74:    ignoreRevsFile = .git-blame-ignore-revs
```

Do we still have substantial external dependencies with code? If so, then we should not apply it to those.

maul-esel:
I'm also in favor of applying this in one huge commit, and adding that commit to the (already existing) blame-ignore file.

Parts of the code that should be excluded:
- JavaCup
- Library-ApacheCommonsCLI
- files generated by ASTBuilder
  - Most of these are not checked in, except those inside ASTBuilder itself (Node, Grammar, ...)
- SMTInterpol, SMTInterpolTest
  - IIRC other parts of the code are also shared with SmtInterpol (Library-SMTLIB I think? and Library-SMTLIBTest?)
  - What about SmtParser?

danieldietsch:
We could ask @jhoenicke if SMTInterpol also wants to use these CCs in Eclipse, then we could also apply them to Library-SMTLib (we probably should anyways).

schuessf:
Okay, if you are all in favour of a single large commit, I will continue with this PR and squash it in the end and add it to `.git-blame-ignore-revs` afterwards.
For now I applied the CCs to 3 more plugins. You can still review the changes and I will add more plugins later on.

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/absint/vpdomain/WeakEquivalenceEdgeLabel.java
```diff
@@ -204,49 +204,44 @@ Set<NODE> projectAwaySimpleElement(final NODE elemToRemove) {
...
+			 * actually, even removeSimpleElement is not enough, because we might be removing a[i], a (in that order)
+			 * then during removing a[i] we add a node a[q], but dont insert a, because it is in remInfo then when we
+			 * remove a, the removeSimpleElement will just say the cc does not have a and do nothing
 			 *
-			 *  old plan: compute all dependents, and remove them one by one
-			 *  current plan: do removeSimpleElement, but take care that no wrong nodes are added
+			 * old plan: compute all dependents, and remove them one by one current plan: do removeSimpleElement, but
+			 * take care that no wrong nodes are added
 			 */
-//			if (mWeakEquivalenceGraph.mEmptyDisjunct instanceof WeqCongruenceClosure<?>) {
```
schuessf, line 222: One thing I just noticed: Some of the larger changes are because of outcommented code, but it would be great to completely remove it...

Heizmann:
Where can I see the new coding conventions?

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/absint/vpdomain/WeakEquivalenceEdgeLabel.java
Heizmann: We should not remove commented code automatically. Commented code indicates that this class is probably of low quality and the author did not finish his work properly. If we remove the commented code, this is not visible any more.

danieldietsch: I am in favor of removing commented code. I also don't think it says anything reliable about the quality of the unit.

schuessf:
> Where can I see the new coding conventions?

In the [settings file](https://github.com/ultimate-pa/ultimate/blob/dev/releaseScripts/eclipse_settings.epf) that should be imported, and of course in the IDE if you did so.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ContractResult.java
```diff
@@ -48,7 +48,8 @@ public class ContractResult extends Result {
 	/**
 	 * Constructor.
 	 *
-	 * @param specs a specification array.
+	 * @param specs
+	 *            a specification array.
```
schuessf, line 52: One thing I noticed here: We always use tabs for indentation, but the parameter description is indented using spaces. This was not adapted recently, but is already in the coding conventions for a long time, therefore changing it, might cause a lot of changes...

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/absint/vpdomain/WeakEquivalenceEdgeLabel.java
Heizmann: I am very surprised to read this. Why do you think it says nothing about the quality of the code? What do you think is the reason that we see commented code in some places?

danieldietsch: Oh, usually because someone doesn't know how to use Git to compare with bases, does not create branches, or does not have enough monitors. Reasons I have seen such code: 
* Wanted to transport unfinished code to another machine, did not want to create branch that does not compile 
* Wanted to implement a feature, did not finish / was interrupted by other things, did not create branch 
* Tried various variants, arrived at the final one, switched to another task and forgot to remove the old one
* Added debug code and commented because "one might need it if another issue arises" 
* Reimplemented something in a better way, kept old code commented as reference during that time and forgot to remove it 

As a rule of thumb, I try to remove all commented code whenever I encounter it. 

EDIT: It might indicate that the programmer is somewhat sloppy, but it doesn't tell you anything about the current state of the code. There might also be a lot of changes by other people in that file, but they didn't remove the comments because they were afraid someone might need them.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/result/ContractResult.java
danieldietsch: Hm, the amount of changes does not really concern me (should it?) -- and it looks nicer, so why not?

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/absint/vpdomain/WeakEquivalenceEdgeLabel.java
Heizmann: I see that you have the well-intended ambition that our code looks good. But I think it is much more important that the code is indeed good.
I have some doubts about the reasons that you mention are predominantly responsible. In the past, it happened quite often that I had to implement things quickly and did not have the time to test or finish the code properly. I would prefer that the code reminds me that this work was not finished and shows me the alternatives I considered when interrupted.

I guess commented code from other authors is sometimes there for the very same reason. Since I am the person who most likely has to maintain the unmaintained code, I would prefer the following policy.

If someone sees commented code and is confident that this class works well, he or she can remove the commented code. Otherwise, he or she should leave the code as it is (and inform the author if disturbed by the commented code).

Heizmann:
I am fine with the coding conventions, but we should not apply them automatically to `UltimateTest`. There we often violate several coding conventions deliberately to improve the readability of testsuites.

Heizmann:
Note: Afterwards we also have to update the wiki https://github.com/ultimate-pa/ultimate/wiki/CodingConvention

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/absint/vpdomain/WeakEquivalenceEdgeLabel.java
danieldietsch: I mean, that goes without saying :) -- but really, you are selling yourself short. Its not often that there is commented code anyways.

schuessf:
> I am fine with the coding conventions, but we should not apply them automatically to `UltimateTest`. There we often violate several coding conventions deliberately to improve the readability of testsuites.

I agree that we don't apply the coding conventions to every file (like in `UltimateTest`). But it would be better to use `// @formatter:off` for such files in the future.

> Note: Afterwards we also have to update the wiki https://github.com/ultimate-pa/ultimate/wiki/CodingConvention

What do you want to change here? The list there is incomplete as it written (`We actually use many more conventions`) and I cannot see any conflicting statements there.

schuessf:
I applied the coding conventions with the exclusions as suggested.
Now, quite a large number of files were changed, therefore I don't think a detailed review is feasible (even though the GitHub UI can display the changes surprisingly well 😉), but maybe you can have a quick look.

trunk/source/Library-SmtLibUtils/src/de/uni_freiburg/informatik/ultimate/lib/smtlibutils/arrays/MultiDimensionalNestedStore.java
```diff
@@ -183,37 +175,21 @@ public boolean equals(final Object obj) {
 	}
 
 	public static MultiDimensionalNestedStore of(final Term term) {
-		MultiDimensionalNestedStore result = convert1mdseq(term);
```
schuessf: This removal is concerning, as it changed the behavior and caused some errors in the [Nightly build](https://jenkins.sopranium.de/blue/organizations/jenkins/Ultimate%2FUltimate%20Nightly/detail/PR-700/1/tests/). It seems like `Remove redundant comparison statement` was responsible for this, therefore I will disable this settings and re-run the formatter.

Heizmann:
> What do you want to change here?
Add a note that the coding conventions are defined by the settings file and that only a subset is listed. There is also a reference to SonarQube that I do not understand.
Foremost, we should make clear what contributors have to do.
* Import our settings.
* Will the autosave actions do everything that is needed?
* Do contributors have to apply the formatter?
* (They do not really have to take care for the individual conventions)

Heizmann:
> I agree that we don't apply the coding conventions to every file (like in `UltimateTest`). But it would be better to use `// @formatter:off` for such files in the future.

I agree. When you edit a file, the autosave result encourages you to include `@formatter` annotations. However, I believe that this should be done individually by each person who edits the files. I am open to the idea of automatically adding `@formatter` annotations, but I currently do not have a solution for how this could be implemented.

trunk/source/Library-SmtLibUtils/src/de/uni_freiburg/informatik/ultimate/lib/smtlibutils/arrays/MultiDimensionalNestedStore.java
Heizmann: Concerning indeed. Is this a bug of Eclipse? Is there a way to apply only rules that to not change the behavior?

PR #701, opened 2025-01-03
Sleep-Reduced Thread-Modular Proofs and related CHC refactoring

maul-esel:
This PR contains the integration of sleep set reduction for thread-modular proofs of concurrent programs, as described in our POPL'24 paper. It has been refactored to use the same symbolic independence relations introduced by PR #696.

The PR also contains the refactoring of program-to-CHC translation that was needed to implement the above techique without too much code duplication; and some fixes to our interfaces to CHC solvers.
:arrow_right: @schuessf as you have worked on thread-modular proofs via CHC before, could you / do you want to take a look?


trunk/source/ChcSolver/src/de/uni_freiburg/informatik/ultimate/plugins/chcsolver/ChcSolverObserver.java
```diff
@@ -88,11 +101,10 @@ public boolean process(final IElement root) throws Throwable {
 	private IChcScript getBackend(final HornAnnot annotation) {
 		switch (mPrefs.getBackend()) {
 		case ELDARICA:
-			return new EldaricaChcScript(mServices, annotation.getScript().getScript());
+			// return new EldaricaChcScript(mServices, annotation.getScript().getScript());
```
schuessf: Can `EldaricaChcScript` be removed now (including the Maven dependencies), and the outcommented code here?

trunk/source/Library-CHC/src/de/uni_freiburg/informatik/ultimate/lib/chc/EldaricaCliChcScript.java
```diff
@@ -0,0 +1,207 @@
...
+
+			return mLastResult;
+
+		} catch (final IOException e) {
+			throw new RuntimeException(e);
+		}
+	}
+
+	private String getCommand(final File file) {
+		var command = "eld";
```
schuessf: This only works, if `eld` is in the `PATH` of the system, should we include an Eldarica build in `releaseScripts`?

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/HcLocalVar.java
```diff
@@ -39,23 +41,27 @@ public Sort getSort() {
...
 	public String toString() {
-		return IcfgToChcConcurrentUtils.getReadableString(mVariable) + (mIndex + 1);
+		return IcfgToChcConcurrentUtils.getReadableString(mVariable) + (mInstance.getInstanceNumber() + 1);
 	}
 
 	@Override
 	public int hashCode() {
-		return Objects.hash(mIndex, mVariable);
+		final int prime = 97;
+		return prime * Objects.hash(mInstance, mVariable);
```
schuessf: Is there any reason why you multiply the hash code from `Objects::hash` with 97?

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/HcLocationVar.java
```diff
@@ -36,23 +41,27 @@ public Sort getSort() {
...
 	public String toString() {
-		return "loc_" + IcfgToChcConcurrentUtils.getReadableString(mProcedure) + "_" + (mIndex + 1);
+		return "loc_" + mInstance;
 	}
 
 	@Override
 	public int hashCode() {
-		return Objects.hash(mIndex, mProcedure);
+		final int prime = 79;
+		return prime * Objects.hash(mInstance);
```
schuessf: Why do you multiply the has code with 79 here?

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/IcfgLiptonReducer.java
```diff
@@ -0,0 +1,185 @@
...
+import de.uni_freiburg.informatik.ultimate.lib.tracecheckerutils.partialorder.petrinetlbe.IcfgCompositionFactory;
+
+/**
+ * Performs Lipton reduction on a concurrent program, and returns a new ICFG for the reduced program.
+ *
+ * @author Dominik Klumpp (klumpp@informatik.uni-freiburg.de)
+ * @author Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ *
+ */
+public class IcfgLiptonReducer {
```
schuessf, line 65: Good idea to extract this to an own class 👍
As this this is independent of CHCs, can you move this to some other plugin (maybe Library-IcfgTransformer)?

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/ThreadModularHornClauseProvider.java
```diff
@@ -0,0 +1,843 @@
...
+			@Override
+			public String toString() {
+				return "<[ dummy edge: assume true; ]>";
+			}
+		};
+	}
+
+	// add actual constraints for spec edges, do nothing if not a spec edge
+	protected void transformSpecEdgeClause(final IcfgEdge edge, final HornClauseBuilder clause) {
+		// TODO Fix this for
```
schuessf, line 383: Do you want to address this TODO in this PR?

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/ThreadModularHornClauseProvider.java
```diff
@@ -0,0 +1,843 @@
...
+		return getCandidateInstancesForClause(edge.getPrecedingProcedure()).stream()
+				.map(t -> new Triple<>(Map.of(t, edge.getSource()), Map.of(t, edge.getTarget()), t))
+				.collect(Collectors.toList());
+	}
+
+	private List<Pair<ThreadInstance, IcfgLocation>> getErrorLocations() {
+		return mIcfg.getProcedureErrorNodes().entrySet().stream()
+				.flatMap(e -> e.getValue().stream().map(l -> new Pair<>(e.getKey(), l)))
+				.flatMap(e -> mInstances.stream().filter(i -> i.getTemplateName().equals(e.getKey()))
+						.<Pair<ThreadInstance, IcfgLocation>> map(i -> new Pair<>(i, e.getValue())))
```
schuessf, line 476: Is this type parameter necessary here?

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/ThreadModularHornClauseProvider.java
```diff
@@ -0,0 +1,843 @@
...
+				.collect(Collectors.toList());
+	}
+
+	private List<Triple<ThreadInstance, IcfgLocation, UnmodifiableTransFormula>> getErrorConditions() {
+		return mIcfg.getProcedureErrorNodes().entrySet().stream()
+				.flatMap(e -> e.getValue().stream().map(l -> new Pair<>(e.getKey(), l)))
+				.flatMap(e -> e.getValue().getIncomingEdges().stream()
+						.map(t -> new Triple<>(e.getKey(), t.getSource(), t.getTransformula())))
+				.flatMap(e -> mInstances.stream().filter(i -> i.getTemplateName().equals(e.getFirst()))
+						.<Triple<ThreadInstance, IcfgLocation, UnmodifiableTransFormula>> map(
```
schuessf, line 486: Is this type parameter necessary here?

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/ThreadModularHornClauseProvider.java
```diff
@@ -0,0 +1,843 @@
...
+			return bodyArgs.stream().map(v -> clause.getBodyVar(v).getTerm()).collect(Collectors.toList());
+		});
+	}
+
+	protected HornClauseBuilder buildNonInterferenceClause(final IIcfgTransition<?> transition,
+			final BiFunction<HornClauseBuilder, ThreadInstance, List<Term>> getInterferencePrecondition) {
+		final var clause = createBuilder(mInvariantPredicate, "non-interference clause for transition "
+				+ transition.hashCode() + " with transformula " + transition.getTransformula());
+
+		// TODO support transitions with multiple predecessors (joins)
```
schuessf, line 675: Does this TODO mean that we just ignore joins and thus overapproximate the behavior of the program here?

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/preferences/IcfgToChcPreferenceInitializer.java
```diff
@@ -0,0 +1,204 @@
...
+
+	public static final String LABEL_THREADMODULAR_LEVEL = "Thread-Modular Proof Level";
+	private static final String DESC_THREADMODULAR_LEVEL =
+			"The level at which thread-modular proofs should be computed";
+	private static final int DEF_THREADMODULAR_LEVEL = 2;
+
+	public static final String LABEL_SYMMETRY_CLAUSES = "Use symmetry clauses";
+	private static final boolean DEF_SYMMETRY_CLAUSES = false;
+
+	// TODO Currently unused
```
schuessf, line 92: Are there any plans to use the settings right now?

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/preferences/IcfgToChcPreferenceInitializer.java
```diff
@@ -0,0 +1,204 @@
...
+
+	public static final String LABEL_NONDET_SLEEP_UPDATES = "Use nondeterministic sleep update encoding";
+	private static final String DESC_NONDET_SLEEP_UPDATES = """
+			If enabled, sleep set updates are encoded such that non-deterministically, \
+			non-commuting edges may or may not be added to the sleep set. \
+			This does not affect CHC satisfiability but may sometimes be simpler for the CHC solver. \
+			It is particularly recommended in connection with necessary-and-complete conditional independence.""";
+	private static final boolean DEF_NONDET_SLEEP_UPDATES = false;
+
+	// TODO Currently unused
```
schuessf, line 129: Are there any plans to use the settings right now?

trunk/source/Library-CHC/src/de/uni_freiburg/informatik/ultimate/lib/chc/ChcSolutionNormalizer.java
```diff
@@ -0,0 +1,83 @@
...
+				.map(fun -> normalize(model.getFunctionDefinition(fun.getName()))).collect(Collectors.toSet()));
+	}
+
+	public FunctionDefinition normalize(final FunctionDefinition definition) {
+		final var body = new UnfTransformer(mScript).transform(definition.getBody());
+		return new FunctionDefinition(definition.getName(), definition.getParams(), body);
+	}
+
+	public Derivation normalize(final Derivation derivation) {
+		// TODO
```
schuessf, line 80: Is the normalization relevant for now?

schuessf (commented):
Thanks for your work! Your changes drastically improve the code (that was rather ad-hoc by me 😄). I have just a few comments left.

trunk/source/ChcSolver/src/de/uni_freiburg/informatik/ultimate/plugins/chcsolver/ChcSolverObserver.java
maul-esel: At least currently, it is not being used. If we want to keep using eldarica as a solver long-term (and perhaps on a more fine-granular level than here), we should probably investigate again if we can access it programmatically, such that we can access derivation trees and other, more detailed output.

I'll remove the code for now. We can always go back to it via git later. For the dependencies, see the next comment below.

trunk/source/Library-CHC/src/de/uni_freiburg/informatik/ultimate/lib/chc/EldaricaCliChcScript.java
maul-esel: We need to make this a bit more portable, I agree.

- `eld` is a bash script, so it won't work on Windows. It sets some parameters and then runs a JAR -- we should implement that directly.
- Do we want to check in the eldarica JAR like we do for the SMT solver binaries, or wouldn't it be better to still have it as a Maven dependency?

I just thought about whether we could actually just call eldarica's `main` method from Java, but that might cause all kinds of unexpected behaviour (e.g. memory remaining allocated, re-invocations behaving incorrectly due to old cached stuff, coupling of required java versions), so... maybe not.

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/HcLocalVar.java
maul-esel: Instances of different classes implementing `IHcReplacementVar` are quite likely to end up in the same hash set / map. Even if their internal data is the same, they are not equal and their hashcodes should not collide. Thus, every implementation multiplies its hash code with a different prime number.

I'll add comments explaining this.

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/HcLocationVar.java
maul-esel: (see above)

trunk/source/IcfgToChc/src/de/uni_freiburg/informatik/ultimate/plugins/icfgtochc/concurrent/IcfgLiptonReducer.java
maul-esel: :+1: I'll have a look

trunk/source/Library-CHC/src/de/uni_freiburg/informatik/ultimate/lib/chc/EldaricaCliChcScript.java
schuessf: > Do we want to check in the eldarica JAR like we do for the SMT solver binaries, or wouldn't it be better to still have it as a Maven dependency?

Yes, the Maven dependency is much nicer of course. But I am not sure how you could actually call the binary in that case.
If it is not straightforward to include Eldarica in `releaseScripts`, we might really just assume that `eld` is in the PATH of the system for now (maybe you can throw an useful exception in that case?).
The best solution would be of course to get rid of `EldaricaCliChcScript` again and call Eldarica programmatically. If we cannot get it working, we could include the JAR along with our custom launchers (incl. for Windows) in `releaseScripts`. But this is more of a long term question 😉

PR #703, opened 2025-01-07
Refactor unchecked reads to improve support for dereferences in ACSL

schuessf:
In our C translation, we translate dereferences to calls to Boogie procedures, e.g., `x = a[i];` is translated to:
```
call #t~mem0 := read~int({ base: ~a~0!base, offset: ~a~0!offset + 4 * ~i~0 }, 4);
~x~0 := #t~mem0;
```
where `read~int` has the following signature (for memsafety we have a stronger contract):
```
procedure read~int(#ptr : $Pointer$, #sizeOfReadType : int) returns (#value : int);
ensures #value == #memory_int[#ptr];
```
This read-call is produced by`ExpressionResultTransformer::switchToRValue`.

For ACSL, we attempted to handle this the same way as for C. However, there was the following issue:
This approach with read-calls works for some ACSL statements (e.g., the translation of `//@ assert x == a[i];` looks similar to above). There are however other ACSL statements / expressions, the translation with calls does not work, as it produces not only an expression, but also auxiliary statements (e.g., function contracts, loop invariants, quantifiers in the future).
Therefore I refactored the `MemoryHandler` to provide a method `getUncheckedReadCall` that creates an array access without any calls or additional checks (also used for the translation of `memcpy` and `memmove`) and use this method in `ExpressionResultTransformer::switchToRValueUnchecked` that is now called in `ACSLHandler`.

### DIFF
--- a/trunk/source/BoogiePreprocessor/src/de/uni_freiburg/informatik/ultimate/boogie/preprocessor/memoryslicer/MemoryArrayReplacer.java
+++ b/trunk/source/BoogiePreprocessor/src/de/uni_freiburg/informatik/ultimate/boogie/preprocessor/memoryslicer/MemoryArrayReplacer.java
@@ -239,6 +239,15 @@ public class MemoryArrayReplacer extends BoogieTransformer {
 		return super.processStatement(statement);
 	}
 
+	@Override
+	protected Expression processExpression(final Expression expr) {
+		if (expr instanceof final IdentifierExpression id && id.getIdentifier().startsWith("#memory")) {
+			throw new MemorySliceException(
+					"Found direct access to memory array (without a read-call), which is not supported yet.");
+		}
+		return super.processExpression(expr);
+	}
+
 	private AssignmentStatement handleArrayWrite(final AssignmentStatement as) {
 		if (as.getLhs().length != 1) {
 			return null;
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
@@ -407,7 +407,10 @@ public class ACSLHandler implements IACSLHandler {
 	private ExpressionResult dispatchSwitch(final IDispatcher main,
 			final de.uni_freiburg.informatik.ultimate.model.acsl.ast.Expression node, final ILocation loc) {
 		final ExpressionResult expr = (ExpressionResult) main.dispatch(node, main.getAcslHook());
-		return mExprResultTransformer.switchToRValue(expr, loc, main.getAcslHook());
+		// Perform an unchecked switch to RValue (i.e., without checking for memsafety).
+		// This also ensures that there are no read-calls for dereferences and thus allows us to use also dereferences
+		// inside ACSL expressions that have to be side-effect-free (e.g., loop invariant or contracts)
+		return mExprResultTransformer.switchToRValueUnchecked(expr, loc, main.getAcslHook());
 	}
 
 	@Override
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/MemoryHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/MemoryHandler.java
@@ -764,15 +764,14 @@ public class MemoryHandler {
 	 * expression of the returned ResultExpression. Note that we only read simple types from the heap -- when reading
 	 * e.g. an array, we have to make readCalls for each cell.
 	 *
-	 * @param tPointer
+	 * @param address
 	 *            the address to read from.
-	 * @param pointerCType
-	 *            the CType of the pointer in tPointer
+	 * @param resultType
+	 *            the CType of the pointer
 	 *
 	 * @return all declarations and statements required to perform the read, plus an identifierExpression holding the
 	 *         read value.
 	 */
-	// 2015-10
 	public ExpressionResult getReadCall(final Expression address, final ICType resultType) {
 		final ILocation loc = address.getLocation();
 		final ExpressionResultBuilder resultBuilder = new ExpressionResultBuilder();
@@ -795,6 +794,16 @@ public class MemoryHandler {
 		return resultBuilder.build();
 	}
 
+	/**
+	 * Handles a read to the given address without any additional memory checks. Therefore, this method always returns a
+	 * single Expression (i.e., access in the corresponding memory array) without any additional statements etc.
+	 *
+	 * @param address
+	 *            the address to read from.
+	 * @param resultType
+	 *            the CType of the pointer
+	 * @return an ExpressionResult consisting only of a array access to the memory array.
+	 */
 	public ExpressionResult getReadUnchecked(final Expression address, final ICType resultType) {
 		final int byteSize;
 		if (resultType instanceof CPointer) {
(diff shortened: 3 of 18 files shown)


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -438,69 +438,72 @@ private static int getCASTBinaryExprOperator(
 	private ExpressionResult dispatchSwitch(final IDispatcher main,
 			final de.uni_freiburg.informatik.ultimate.model.acsl.ast.Expression node, final ILocation loc) {
 		final ExpressionResult expr = (ExpressionResult) main.dispatch(node, main.getAcslHook());
+		if (expr.getLrValue() instanceof final HeapLValue heapValue) {
+			final CType type = CEnum.replaceEnumWithInt(expr.getLrValue().getCType().getUnderlyingType());
+			// Perform a custom switch to RValue without any read calls
+			// This allows us to use also dereferences inside ACSL expressions that have to be side-effect-free
+			// (e.g., loop invariant or contracts)
```
maul-esel: But this will now also be used for ACSL asserts, right?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -438,69 +438,72 @@ private static int getCASTBinaryExprOperator(
...
+		if (expr.getLrValue() instanceof final HeapLValue heapValue) {
+			final CType type = CEnum.replaceEnumWithInt(expr.getLrValue().getCType().getUnderlyingType());
+			// Perform a custom switch to RValue without any read calls
+			// This allows us to use also dereferences inside ACSL expressions that have to be side-effect-free
+			// (e.g., loop invariant or contracts)
+			if (type instanceof CPrimitive || type instanceof CPointer) {
+				final Expression memoryAccess = ExpressionFactory.constructNestedArrayAccessExpression(loc,
+						mCHandler.getMemoryArrayForType(type), new Expression[] { heapValue.getAddress() });
+				return new ExpressionResultBuilder(expr).resetLrValue(new RValue(memoryAccess, type)).build();
+			}
```
maul-esel: Do I understand correctly that this code replaces the calls to `mExprResultTransformer.switchToRValue` (and related methods) that were previously made?

If so, I am a bit unsure if this is a good place to have this. For instance, it could easily get out-of-sync with changes in `switchToRValue`. Would it be possible to make this a specialized method e.g. `mExprResultTransformer.switchToRValueWithoutSideEffects`?

maul-esel (commented):
:+1: Thanks.

I'm not deeply familiar with the ACSL translation, so unfortunately I can only offer some superficial comments (see below).

Could you also add some test cases here?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: Yes, it will be used for any ACSL expression.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: > Do I understand correctly that this code replaces the calls to `mExprResultTransformer.switchToRValue` (and related methods) that were previously made?

Yes, any call to `mExprResultTransformer.switchToRValue` in `ACSLHandler` was replaced with a call to this method.

> If so, I am a bit unsure if this is a good place to have this. For instance, it could easily get out-of-sync with changes in `switchToRValue`. Would it be possible to make this a specialized method e.g. `mExprResultTransformer.switchToRValueWithoutSideEffects`?

We cannot get completely out-of-sync, this method still has the call to `mExprResultTransformer.switchToRValue` as a base case (which does currently nothing, but if this changes, it will also be visible in this method).
We could also make it a public method in `ExpressionResultTransformer`, but this `switchToRValue` does not have any memsafety-checks (even if the setting is enabled). Therefore I am not sure, if it is a good idea to have such a public method in `ExpressionResultTransformer`.

schuessf:
I just added two test cases. The should actually pass, since the `MemorySlicer` is not enabled by default.

schuessf:
Update: This is not only important that we can support e.g., array accesses inside quantifiers or loop invariants (which was the initial motiviation), but also if we want to assume that there is no undefined behaviour in the program, this should not have an impact on the ACSL invariants (we don't want them to block).

Heizmann (approved):

danieldietsch (approved):

trunk/source/BoogiePreprocessor/src/de/uni_freiburg/informatik/ultimate/boogie/preprocessor/memoryslicer/MemoryArrayReplacer.java
```diff
@@ -239,6 +239,15 @@ protected Statement processStatement(final Statement statement) {
 		return super.processStatement(statement);
 	}
 
+	@Override
+	protected Expression processExpression(final Expression expr) {
+		if (expr instanceof final IdentifierExpression id && id.getIdentifier().startsWith("#memory")) {
```
maul-esel, line 244: Nitpick: Use the constant `SFO.MEMORY` instead of a string literal.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/MemoryHandler.java
```diff
@@ -748,6 +747,18 @@ public Pair<RValue, CallStatement> getUltimateMemAllocInitCall(final ILocation a
...
+		if (type instanceof final CPrimitive primitive) {
+			mRequiredMemoryModelFeatures.reportDataOnHeapRequired(primitive.getType());
+			return mMemoryModel.getDataHeapArray(primitive.getType());
+		}
+		throw new AssertionError("There is no memory array for the type " + type);
+	}
+
 	/**
 	 * Generates a call of the read procedure and writes the returned value to a temp variable, returned in the
 	 * expression of the returned ResultExpression. Note that we only read simple types from the heap -- when reading
```
maul-esel, line 764: If I understand correctly, we are no longer necessarily calling a procedure here. If so, we should update the documentation.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/MemoryHandler.java
```diff
@@ -768,13 +779,26 @@ public ExpressionResult getReadCall(final Expression address, final CType result
...
+				byteSize = mTypeSizes.getSizeOfPointer();
+			} else if (resultType instanceof final CPrimitive prim) {
+				byteSize = mTypeSizes.getSize(prim.getType());
+			} else {
+				throw new AssertionError("We only support reading primitive or pointer types");
+			}
+			return new ExpressionResult(new RValue(
+					readFromHeap(loc, determineMemoryArrayForType(resultType), address, resultType, byteSize),
+					resultType));
+		}
```
maul-esel: Instead of having this parameter, could we not just have a different method?

maul-esel (approved):
Looks good :+1: I just left some minor comments.

trunk/source/BoogiePreprocessor/src/de/uni_freiburg/informatik/ultimate/boogie/preprocessor/memoryslicer/MemoryArrayReplacer.java
schuessf: `SFO` is part of `CACSL2BoogieTranslator` which should not be imported by `BoogirPreprocessor`. There are more cases like that in `MemorySlicer` and this is a bit of a conceptional issue (but it does work 🙂).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/MemoryHandler.java
schuessf: Yes, this should be possible in principle. At some point we probably need a case distinction, but I will check it again.

PR #704, opened 2025-01-07
Support quantifiers in ACSL

schuessf:
This PR adds support for quantifiers in ACSL.
* We already had some rules in our grammar to parse quantifiers. These rules just always yielded a `NotDefinedExpression`. There was also a AST object for a `QuantifierExpression` that I just adapted to our needs.
* To translate quantifiers from ACSL to Boogie, I added a new method to `ACSLHandler`. The crucial steps in this translation are that we have to handle the quantified variables in the expression inside the quantifier properly (track the variables in a `ScopedHashMap` and first lookup when handling an `IdentifierExpression`) and that we add type constraints over the quantified variables (ACSL may use bounded types that are translated to an unbounded type in Boogie).

Note that we can only handle quantified expressions, where the inner expressions is side-effect free (which also does not allow any auxiliary statements). Therefore, quantified expressions that contain dereferences are currently not supported, but this should be fixed in #703.

### DIFF
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
@@ -121,9 +121,11 @@ import de.uni_freiburg.informatik.ultimate.model.acsl.ast.LoopStatement;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ast.LoopVariant;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ast.MallocableExpression;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ast.OldValueExpression;
+import de.uni_freiburg.informatik.ultimate.model.acsl.ast.QuantifierExpression;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ast.RealLiteral;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ast.Requires;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ast.ValidExpression;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.ScopedHashMap;
 
 /**
  * @author Markus Lindenmann
@@ -173,6 +175,8 @@ public class ACSLHandler implements IACSLHandler {
 	private final LocationFactory mLocationFactory;
 	private final CHandler mCHandler;
 	private final CExpressionTranslator mCExpressionTranslator;
+
+	private final ScopedHashMap<String, LRValue> mBoundVariables = new ScopedHashMap<>();
 
 	public ACSLHandler(final boolean witnessInvariantMode, final FlatSymbolTable symboltable,
 			final ExpressionTranslation expressionTranslation, final ITypeHandler typeHandler,
@@ -650,6 +654,10 @@ public class ACSLHandler implements IACSLHandler {
 	@Override
 	public Result visit(final IDispatcher main,
 			final de.uni_freiburg.informatik.ultimate.model.acsl.ast.IdentifierExpression node) {
+		final var boundVar = mBoundVariables.get(node.getIdentifier());
+		if (boundVar != null) {
+			return new ExpressionResult(boundVar);
+		}
 		final ILocation loc = mLocationFactory.createACSLLocation(node);
 		final String id = lookupId(main, node, loc);
 
@@ -707,6 +715,47 @@ public class ACSLHandler implements IACSLHandler {
 		default:
 			throw new IncorrectSyntaxException(loc, "The type of specType should be in some type!");
 		}
+	}
+
+	@Override
+	public Result visit(final IDispatcher main, final QuantifierExpression node) {
+		mBoundVariables.beginScope();
+		final ILocation loc = mLocationFactory.createACSLLocation(node);
+		final List<VarList> quantifiedVars = new ArrayList<>();
+		final List<Expression> typeConstraints = new ArrayList<>();
+		for (final var decl : node.getVariables()) {
+			// For each quantified variable in the ACSL expression, create a corresponding Boogie variable and store it
+			// in the mBoundVariables to be used when handling IdentifierExpressions.
+			final String name = decl.getName();
+			final CType cType = AcslTypeUtils.translateAcslTypeToCType(decl.getType());
+			if (!(cType instanceof CPrimitive)) {
+				throw new UnsupportedSyntaxException(loc, "Only quantified variables of primitive type are supported.");
+			}
+			final DeclarationInformation declInfo = new DeclarationInformation(StorageClass.QUANTIFIED, null);
+			final BoogieType boogieType = mTypeHandler.getBoogieTypeForCType(cType);
+			mBoundVariables.put(name, new LocalLValue(new VariableLHS(loc, boogieType, name, declInfo), cType, false));
+			quantifiedVars.add(new VarList(loc, new String[] { name }, mTypeHandler.cType2AstType(loc, cType)));
+			// Collect the type constraints for the given CType (if any)
+			final var id = ExpressionFactory.constructIdentifierExpression(loc, boogieType, name, declInfo);
+			final var constraint = mExpressionTranslation.getTypeConstraint(loc, id, cType);
+			if (constraint.isPresent()) {
+				typeConstraints.add(constraint.get());
+			}
+		}
+		final ExpressionResult subResult =
+				mExprResultTransformer.rexIntToBool(dispatchSwitch(main, node.getSubformula(), loc), loc);
+		if (!subResult.hasNoSideEffects()) {
+			throw new UnsupportedSyntaxException(loc, "Unable to handle quantified expressions with side-effects.");
+		}
+		// Create a quantifier expression in Boogie
+		// As Boogie uses mathematical integers, we add type constraints inside the quantifier, i.e., we produce
+		// (exists ... typeConstraints && subResult) and (forall ... typeConstraints ==> subResult)
+		final Expression inner = ExpressionFactory.newBinaryExpression(loc,
+				node.isUniversal() ? Operator.LOGICIMPLIES : Operator.LOGICAND,
+				ExpressionFactory.and(loc, typeConstraints), subResult.getLrValue().getValue());
+		final Expression result = ExpressionFactory.quantifier(loc, node.isUniversal(), quantifiedVars, inner);
+		mBoundVariables.endScope();
+		return new ExpressionResult(new RValue(result, new CPrimitive(CPrimitives.BOOL), true));
 	}
 
 	@Override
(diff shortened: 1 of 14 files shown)


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainDispatcher.java
```diff
@@ -535,7 +535,7 @@ public Result dispatch(final ACSLNode n, final IASTNode cHook) {
 				return mAcslHandler.visit(this, (IfThenElseExpression) n);
 			}
 			if (n instanceof QuantifierExpression) {
-				return mAcslHandler.visit(this, n);
+				return mAcslHandler.visit(this, (QuantifierExpression) n);
```
maul-esel, line 538:
```suggestion
			if (n instanceof QuantifierExpression quantified) {
				return mAcslHandler.visit(this, quantified);
```

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
```diff
@@ -353,8 +354,15 @@ ExpressionResult convertToBool(final ILocation loc, final ExpressionResult expr)
...
-			ExpressionResultBuilder expressionResultBuilder);
+	public void addAssumeValueInRangeStatements(final ILocation loc, final Expression expr, final CType ctype,
+			final ExpressionResultBuilder expressionResultBuilder) {
+		final var constraint = getTypeConstraint(loc, expr, ctype);
+		if (constraint.isPresent()) {
+			expressionResultBuilder.addStatement(new AssumeStatement(loc, constraint.get()));
+		}
+	}
+
+	public abstract Optional<Expression> getTypeConstraint(ILocation loc, Expression expr, CType cType);
```
maul-esel: Could you add a documentation comment for this method?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -677,6 +685,47 @@ private String lookupId(final IDispatcher main,
...
+		final List<VarList> quantifiedVars = new ArrayList<>();
+		final List<Expression> typeConstraints = new ArrayList<>();
+		for (final var decl : node.getVariables()) {
+			// For each quantified variable in the ACSL expression, create a corresponding Boogie variable and store it
+			// in the mQuantifiedVariables to be used when handling IdentifierExpressions.
+			final String name = decl.getName();
+			final CType cType = AcslTypeUtils.translateAcslTypeToCType(decl.getType());
+			final DeclarationInformation declInfo = new DeclarationInformation(StorageClass.QUANTIFIED, null);
+			final BoogieType boogieType = mTypeHandler.getBoogieTypeForCType(cType);
+			// TODO: Handle quantified variables on heap
```
maul-esel: What exactly does this mean, and what would have to be done?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -173,6 +175,8 @@ private enum SPEC_TYPE {
 	private final CHandler mCHandler;
 	private final CExpressionTranslator mCExpressionTranslator;
 
+	private final ScopedHashMap<String, LRValue> mQuantifiedVariables = new ScopedHashMap<>();
```
maul-esel: Maybe `mBoundVariables`, so it also makes sense if support for `let` or `lambda` is added.

maul-esel (commented):
:+1: Thanks for taking care of this. Looks good to me, I've only left some smaller comments.

Could you maybe add some small test cases? (maybe some using verification/Automizer and some for Referee?)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainDispatcher.java
schuessf: I agree that this new Java feature would be nice here, but I would rather keep it as it as, to be consistent with the rest of `MainDispatcher`:
```java
if (n instanceof BinaryExpression) {
	return mAcslHandler.visit(this, (BinaryExpression) n);
}
```

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: Actually I don't know 😄
This wold happen, if we quantify over pointers and I am not sure about the ACSL semantics in that case.
Also, I don't think we support pointer types in ACSL properly, so we would probably crash earlier.
So maybe I can just replace the TODO with a comment that we don't support pointers for now and thus yield a `LocalLValue`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainDispatcher.java
maul-esel: Yeah, I was tempted to suggest rewriting the entire method (or multiple methods) with pattern-matching switch-cases, but better to keep this PR simple :grin:

schuessf:
Regarding the test cases: I wanted to wait for #703 to write some tests with quantifiers over arrays, but I will just add some simple tests to demonstrate the basic functionality,

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
danieldietsch: I am also confused what "quantified variables on heap" means. Is this a quantified pointer? 
But I agree, that sounds rather difficult. Perhaps we should crash as soon as we encounter it, i.e., throw UnsupportedSyntax .

danieldietsch (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: I just replaced this TODO with an exception. I don't think it can be triggered for now, as we already crash before for non-primitive types (definitely in `AcslTypeUtils::translateAcslTypeToCType`, maybe already during parsing). But this may change in the future and the thrown exception also serves as some form of documentation.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/IntegerTranslation.java
```diff
@@ -419,25 +419,24 @@ public CPrimitive getCTypeOfPointerComponents() {
...
-			final ExpressionResultBuilder expressionResultBuilder) {
+	public Optional<Expression> getTypeConstraint(final ILocation loc, final Expression expr, final CType cType) {
 		if (!mSettings.assumeNondeterministicValuesInRange() || !cType.getUnderlyingType().isIntegerType()) {
 			// only integer types can be out of range
-			return;
+			return Optional.empty();
 		}
 		final CPrimitive cPrimitive = (CPrimitive) CEnum.replaceEnumWithInt(cType.getUnderlyingType());
 		if (mTypeSizes.isUnsigned(cPrimitive)) {
 			// only signed types can be out of range
```
maul-esel, line 429: Is this also the case for quantified variables?

I remember that at one point, we had two approaches to unsigned variables: either (1) insert appropriate modulo operations in every comparison (so variables may have values out-of-range but it cannot be observed by the program), or (2) insert appropriate modulo operations whenever an unsigned value is computed.

It seems to me that while (1) may work without type constraints in quantified expressions, (2) probably would not.

trunk/source/CodeCheck/src/de/uni_freiburg/informatik/ultimate/plugins/generator/codecheck/preferences/CodeCheckPreferenceInitializer.java
```diff
@@ -121,7 +121,7 @@ public enum RedirectionStrategy {
 	public static final boolean DEF_USESEPARATETRACECHECKSOLVER = true;
 	public static final SolverMode DEF_CHOOSESEPARATETRACECHECKSOLVER = SolverMode.Internal_SMTInterpol;
 	public static final String DEF_SEPARATETRACECHECKSOLVERCOMMAND = "";
-	public static final String DEF_SEPARATETRACECHECKSOLVERTHEORY = "QF_AUFLIA";
+	public static final String DEF_SEPARATETRACECHECKSOLVERTHEORY = "ALL";
```
maul-esel: Is it really intended to change the default setting here? If so, are there any settings files that also need to be adapted?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/IntegerTranslation.java
schuessf: Currently, we use (1) in `IntegerTranslation`. There was some idea for an alternative translation that uses (2), but which is not implemented on `dev` (only started on some older branch).
There is also a test for unsigned quantified variables:
`//@ assert \forall unsigned x; x >= 0;` is translated to `assert forall x :: int x % 4294967296 >= 0;`, which holds trivially.

trunk/source/CodeCheck/src/de/uni_freiburg/informatik/ultimate/plugins/generator/codecheck/preferences/CodeCheckPreferenceInitializer.java
schuessf: In all the relevant settings (e.g. for SV-COMP) `ALL` was already used. The most simple solution to fix the failing test would be of course to just adapt the settings files that is responsible for the tests. However, to be consistent with Automizer (see `RcfgPreferenceInitializer`), I adapted the default setting for Kojak instead, but maybe this is nothing that should be done in this PR.

PR #707, opened 2025-02-09
ASTBuilder: Generate sealed class hierarchy

maul-esel:
This PR changes the `ASTBuilder` such that it marks the generated AST classes as `sealed` (for abstract supertypes such as `Expression`) resp. `final` (for concrete classes such as `IdentifierExpression`).

The intent is to have a precise, complete and closed description of the elements that can make up an AST. As a major consequence, this enables exhaustiveness checks for pattern matching! Code can now switch over all types of e.g. Boogie expressions like this:
```java
switch (expr) {
case IdentifierExpression idExpr -> ...
case BooleanLiteral boolLit -> ...
... (more cases) ...
}
```
**without needing to include a `default` case**, as the compiler can be sure whether or not all cases are handled. This doesn't just get rid of an annoying line of code, it will also have the benefit that if new AST classes are added, the compiler immediately reports all the case distinctions where the new classes must be handled -- rather than silently going to the `default` and crashing (or worse, behaving incorrectly) at runtime.

My hope is that such pattern matching can be particularly useful when (back-)translating between C, ACSL, Boogie and SMT. I was fiddling around in the `IcfgBuilder` code when I had this idea; there are several places there that could benefit from pattern matching.

## Other Changes
To make this change work, I had to get rid of two hand-written classes that extend the Boogie resp. ACSL expressions and were used in backtranslation. One of them is now in the  official ACSL AST (as the backtranslation creates instances that are passed on to other code), the other no longer extends the Boogie AST.

## Limitations
I stopped short of even more pattern-matching support by making the classes records -- this would be a more complex change:
- Records cannot have superclasses (other than `java.lang.Record`). So we would have to replace abstract base classes like `Expression` with (sealed) interfaces. Some functionality could be implemented in those interfaces with default methods, but all fields would have to be generated in each concrete implementation.
- Records can only have immutable fields, and our AST classes have some write-once fields particularly for the `ILocation`. These cannot easily be made final because the object graph is cyclic -- the node points to the location, and the location points to the node.

As there seem to be some plans to support better pattern matching for (non-record) classes in future Java versions, perhaps the goal of better AST pattern matching can then be achieved in an easier way.

### DIFF
--- a/trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/acsl.ast
+++ b/trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/acsl.ast
@@ -274,6 +274,11 @@ Expression ::=
 	| { /** This can be used as call forall parameter, or as if or 
 	     * while condition.  In all other places it is forbidden. */ 
 		WildcardExpression }
+	| { /** This is used to represent addresses in backtranslated program executions.
+	     *  There is no corresponding grammar rule for the parser. */
+	    FakePointerExpression }
+		base : Expression
+		offset : Expression
 	);
  
 ACSLType ::= 
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
@@ -751,20 +751,17 @@ public class ACSLHandler implements IACSLHandler {
 	@Override
 	public Result visit(final IDispatcher main, final Contract node) {
 		final ILocation loc = mLocationFactory.createACSLLocation(node);
-		final ArrayList<Specification> spec = new ArrayList<>();
-		// First we catch the case that a contract is at a FunctionDefinition
-		if (node instanceof IASTFunctionDefinition) {
-			final String msg = "Syntax Error, Contracts on FunctionDefinition are not allowed";
-			throw new IncorrectSyntaxException(loc, msg);
-		}
 
+		final ArrayList<Specification> spec = new ArrayList<>();
 		for (final ContractStatement stmt : node.getContractStmt()) {
 			spec.addAll(Arrays.asList(((ContractResult) main.dispatch(stmt, main.getAcslHook())).getSpecs()));
 		}
+
 		if (node.getBehaviors() != null && node.getBehaviors().length != 0) {
 			final String msg = "Not yet implemented: Behaviour";
 			throw new UnsupportedSyntaxException(loc, msg);
 		}
+
 		// TODO : node.getCompleteness();
 		mSpecType = ACSLHandler.SPEC_TYPE.NOT;
 		return new ContractResult(spec.toArray(new Specification[spec.size()]));
--- a/.gitignore
+++ b/.gitignore
@@ -96,6 +96,7 @@ trunk/examples/concurrent/bpl/weaver-benchmarks/weaver
 /trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/Decreases.java
 /trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/Ensures.java
 /trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/Expression.java
+/trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/FakePointerExpression.java
 /trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/FieldAccessExpression.java
 /trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/FreeableExpression.java
 /trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/FunctionApplication.java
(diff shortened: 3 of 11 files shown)


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -751,13 +751,15 @@ public Result visit(final IDispatcher main, final QuantifierExpression node) {
...
+		// TODO The check below is broken. The compiler detects this and reports an error.
 		// First we catch the case that a contract is at a FunctionDefinition
-		if (node instanceof IASTFunctionDefinition) {
-			final String msg = "Syntax Error, Contracts on FunctionDefinition are not allowed";
-			throw new IncorrectSyntaxException(loc, msg);
-		}
+		// if (node instanceof IASTFunctionDefinition) {
+		// final String msg = "Syntax Error, Contracts on FunctionDefinition are not allowed";
+		// throw new IncorrectSyntaxException(loc, msg);
+		// }
```
maul-esel: Making the type hierarchy sealed causes the compiler to detect that this  `instanceof` check  is meaningless (i.e., always `false`) and report an error. The check was always meaningless, it's only now being reported.

Can this check be simply removed? If not, where could this be checked instead? The problem with checking it here is that we need the C AST node that is annotated with a `Contract`, but we do not have it here (and we accidentally used the ACSL AST node instead).

danieldietsch:
This is a really nice change!

danieldietsch (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
schuessf: I think this check should be removed. It seems like the intention was to crash on ACSL contracts at function definitions, but this should be supported by now (e.g. see [ACSL-contractOnDelarationAndDefinition.c](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/examples/programs/regression/c/ACSL-contractOnDelarationAndDefinition.c)).

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/acsl.ast
```diff
@@ -274,6 +274,11 @@ Expression ::=
 	| { /** This can be used as call forall parameter, or as if or 
 	     * while condition.  In all other places it is forbidden. */ 
 		WildcardExpression }
+	| { /** This is used to represent addresses in backtranslated program executions.
```
schuessf, line 277: I am not quite sure, if this should be in the AST. If we also wanted exhaustive switchs in `ACSLHandler`, we also need to have a case there, even though `FakePointerExpression` does only occur in the backtranslation.
Would it be possible to have either an `Expression` or a `FakePointerExpression` in `BacktranslatedExpresison` instead?

schuessf (commented):
Nice change, I have just one open question regarding the change in the AST.

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/acsl.ast
maul-esel: > If we also wanted exhaustive switchs in ACSLHandler, we also need to have a case there

Yes, that is definitely a downside of the current solution.

> Would it be possible to have either an `Expression` or a `FakePointerExpression` in `BacktranslatedExpression` instead?

Perhaps.
- By moving `FakePointerExpression` outside of the AST, we would forbid nesting it in a larger expression. Currently, we only use `FakePointerExpression` in backtranslation of executions, where such nesting probably never occurs. If we also wanted to use `FakePointerExpression` in invariants, things might be more difficult; though we probably never want to export such invariants as these expressions only make sense in the context of our tools (e.g. for debugging).
- If we change `BacktranslatedExpression` to alternatively allow a `FakePointerExpression`, this of course also affects all `BacktranslatedExpression`s representing invariants. But that is perhaps no major issue; we could e.g. provide a method that returns an `Expression` and throws an exception eagerly if, by some bug, a `FakePointerExpression` makes it into an invariant.

I would suggest doing these changes in a separate PR though, as this one is really supposed to focus on ASTBuilder.

schuessf: `FakePointerExpression`s are only used as values in program states, they are not used in invariants or nested in any other expressions. But you are right, we could move this to a separate PR.

schuessf (approved):

schuessf:
I just started the Nightly build to be sure that there is no issue with this change 🙂

maul-esel:
:+1: Thanks. I always forget that it has to be re-enabled after creating the PR. The [previous build for the branch](https://jenkins.sopranium.de/job/Ultimate/job/Ultimate%20Nightly/job/wip%252Fdk%252Fsealed-ast/1/) looks good though.

maul-esel:
If there are no further objections and the CI does not show new problems, I will merge this tonight.

schuessf:
> 👍 Thanks. I always forget that it has to be re-enabled after creating the PR. The [previous build for the branch](https://jenkins.sopranium.de/job/Ultimate/job/Ultimate%20Nightly/job/wip%252Fdk%252Fsealed-ast/1/) looks good though.

Oh, if you ran it on the branch before, there are probably to crucial changes. I just did not find any reference in this PR to a succesful Nightly. But it does not hurt to wait for another Nightly run either.

bahnwaerter (approved):

PR #709, opened 2025-03-05
Add auxiliary attribute to labels

Heizmann:
Add auxiliary attribute to all Boogie labels that are auxiliary labels (an auxiliary label is a label that does not represent a label of the original program but is introduced by our translations).

I ommited `DSITransformer` and `Cookiefy` since I am not familiar with these projects.

### DIFF
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
@@ -147,10 +147,8 @@ import de.uni_freiburg.informatik.ultimate.boogie.ast.GotoStatement;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.HavocStatement;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.IfStatement;
-import de.uni_freiburg.informatik.ultimate.boogie.ast.Label;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.LeftHandSide;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.LoopInvariantSpecification;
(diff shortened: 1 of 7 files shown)


trunk/source/Library-BoogieAST/src/de/uni_freiburg/informatik/ultimate/boogie/BoogieUtils.java
```diff
@@ -35,25 +36,59 @@
...
+	public static NamedAttribute constructAuxiliaryLabelAttribute(final ILocation loc, final boolean isAuxiliary) {
+		return new NamedAttribute(loc, AUXILIARY_LABEL,
+				new Expression[] { ExpressionFactory.createBooleanLiteral(loc, isAuxiliary) });
+	}
+
+	/**
+	 * Construct {@link Label} and add attribute that indicates that this label is an auxiliary label.
+	 */
+	public static Label constuctAuxiliaryLabel(final ILocation loc, final String name) {
+		return new Label(loc, name, new NamedAttribute[] { constructAuxiliaryLabelAttribute(loc, true) });
```
schuessf: What is the reason for this change? Previosuly, the idea was that `{:auxiliary_label" true}` can be omitted (iirc). Therefore, there was also no change necessary to the existing code (except in the C translation, where we actually want `{:auxiliary_label" false}`), since the attributes are only optional in the constructor of `Label`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -2025,10 +2024,6 @@ public Result visit(final IDispatcher main, final IASTLabelStatement node) {
 		final ArrayList<Statement> stmt = new ArrayList<>();
 		final ArrayList<Declaration> decl = new ArrayList<>();
 		final List<Overapprox> overappr = new ArrayList<>();
-		final String label = node.getName().toString();
-		// Mark the label with the attribute { :auxiliary_label false}
-		stmt.add(new Label(loc, label, new NamedAttribute[] { new NamedAttribute(loc, "auxiliary_label",
```
schuessf, line 2030: You removed the label entirely. Instead, we want a non-auxiliary label here.

PR #714, opened 2025-04-14
Refactor handling of standard functions and predefined types

schuessf:
In the translation from C to Boogie, we also need to model standard functions. The whole handling currently happens in the single class  `StandardFunctionHandler`, which is really large for this reason (4k LOC).

Therefore, I tried to refactor this by splitting `StandardFunctionHandler`. Now there is an abstract class `FunctionModelProvider` that simply returns a list of supported and unsupported functions (and which also provides fields and helper functions that are useful for different implementations). For each library that is supported in ULTIMATE, there is now a different implementation of `FunctionModelProvider` (e.g., `StdlibFunctionModelProvider` for to handle function declared in `stdlib.h`).

`StandardFunctionHandler` is still preserved, but instead of handling all the functions, it now contains only a `List<FunctionModelProvider>` from which the Map of supported functions is built and thus the size of the class is heavily decreased.


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
```diff
@@ -139,86 +69,70 @@
...
-			final LocationFactory locationFactory, final ITypeHandler typeHandler,
-			final CExpressionTranslator cEpressionTranslator, final DataRaceChecker dataRaceChecker) {
+			final TypeSizes typeSizes, final FlatSymbolTable symboltable, final TranslationSettings settings,
+			final ExpressionResultTransformer expressionResultTransformer, final LocationFactory locationFactory,
+			final ITypeHandler typeHandler, final CExpressionTranslator cEpressionTranslator,
+			final DataRaceChecker dataRaceChecker) {
+		final List<FunctionModelProvider> functionModelProviders = List.of(
+				new LinuxFunctionModelProvider(auxVarInfoBuilder, nameHandler, expressionTranslation, memoryHandler,
+						typeSizeAndOffsetComputer, procedureManager, typeSizes, settings, expressionResultTransformer,
+						typeHandler, cEpressionTranslator, dataRaceChecker),
```
danieldietsch: It looks like we could use some container object for the different FunctionModelProvider

danieldietsch (changes requested):
Looks like a sensible change. But you should add headers to the new files.

Some questions though:
* How do we decide on a name for new FMPs? It might be nicer if they were aligned with a certain import instead of by semantic feature (it would also allow us to support different variants of stdlib in the future). 
* How/when do we move helper functions to the FMP base class? In particular, if I need to implement a new C function, I now have to search all the FMPs to find possibly useful helper functions and then decide whether I should pull them up or not. 

it might be nice to have a companion class that contains all "helpers", i.e., functions that actually create Boogie code, and pass that helper to the different FMPs which then use them to create the functions in that specific import. 
This would avoid a somewhat strange base class and the question of where to put these helper functions (as well as if someone wants to write a new function, they already see the available building blocks). 
But it might result in a larger file (which I personally do not find problematic, as long as the purpose is clear enough).

maul-esel:
:+1: Thanks for taking this on.

- Perhaps similar to @danieldietsch, I would have expected the functions to be grouped by the corresponding standard library header (where applicable). Though perhaps for large headers, it might make sense to split them into smaller units. But even then, it would be good to see which units belong to which header, and perhaps to even have one class for the entire header that bundles the smaller units.
- (Standard library) header files can also declare more things than just functions, such as types and macros. As e.g. our model of a standard function that takes a `pthread_t` is typically tied to the representation of `pthread_t`, it would make sense to handle these different declarations together. Could the FMPs be extended in this direction?

  Currently, some builtin types are hard-coded in `TypeHandler`: https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/TypeHandler.java#L281-L302 As we move towards non-preprocessed files, a lot more of these will be needed.

schuessf:
> But you should add headers to the new files.

I will, once the file structure of this PR is final :wink:

> How do we decide on a name for new FMPs? It might be nicer if they were aligned with a certain import instead of by semantic feature (it would also allow us to support different variants of stdlib in the future).

> Perhaps similar to @danieldietsch, I would have expected the functions to be grouped by the corresponding standard library header (where applicable). Though perhaps for large headers, it might make sense to split them into smaller units. But even then, it would be good to see which units belong to which header, and perhaps to even have one class for the entire header that bundles the smaller units.

This was the idea, there are the following implementations:
* `AssertFunctionModelProvider` for `assert.h`
* `AtomicFunctionModelProvider` for `stdatomic.h`
* `FloatFunctionModelProvider` for `math.h` and `fenv.h` (didn't seem reasonable to split; can be renamed the very least)
* `GccBuiltinFunctionModelProvider` for GCC builtin functions (see mostly [here](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html) and [here](https://gcc.gnu.org/onlinedocs/gcc-3.4.5/gcc/Other-Builtins.html))
* `LinuxFunctionModelProvider` for functions that were found [here](https://man7.org/linux/man-pages/man3/) (not sure, if we should consider them as standard function, but they appear in SV-COMP)
* `PthreadFunctionModelProvider` for `pthread.h`
* `SetjmpFunctionModelProvider` for `setjmp.h` (we don't actually support and, therefore this class is very small)
* `SocketFunctionModelProvider` for `socket.h` (see [here](https://pubs.opengroup.org/onlinepubs/009604499/basedefs/sys/socket.h.html))
* `StdioFunctionModelProvider` for `stdio.h`
* `StdlibFunctionModelProvider` for `stdlib.h`
* `StringFunctionModelProvider` for `string.h`
* `SvcompFunctionModelProvider` for SV-COMP specific functions (see [here](https://sv-comp.sosy-lab.org/2025/rules.php))
* `TimeFunctionModelProvider` for `time.h`
* `VariadicFunctionModelProvider` for `stdarg.h` 

> it might be nice to have a companion class that contains all "helpers", i.e., functions that actually create Boogie code, and pass that helper to the different FMPs which then use them to create the functions in that specific import. This would avoid a somewhat strange base class and the question of where to put these helper functions (as well as if someone wants to write a new function, they already see the available building blocks). But it might result in a larger file (which I personally do not find problematic, as long as the purpose is clear enough).

I can certainly can split `FunctionModelProvider` into an interface and a non-abstract helper class. This would also reduce the number of arguments in the constructors. But how should it be possible to extract "functions that actually create Boogie code"? This is most of the work done in the implementations of  `FunctionModelProvider` and even if it is possible, it is really much work.

> (Standard library) header files can also declare more things than just functions, such as types and macros. As e.g. our model of a standard function that takes a `pthread_t` is typically tied to the representation of `pthread_t`, it would make sense to handle these different declarations together. Could the FMPs be extended in this direction?
> Currently, some builtin types are hard-coded in `TypeHandler`: https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/TypeHandler.java#L281-L302 As we move towards non-preprocessed files, a lot more of these will be needed.

Yes, I also thought about this. It should be possible, we just need to move the instatiations of FMPs outside of the `StandardFunctionHandler` (maybe this is a good idea anyway) and add another method that yields a map of defined types. I will have a look, but maybe we should move this to a separate PR.

danieldietsch:
> I can certainly can split FunctionModelProvider into an interface and a non-abstract helper class. This would also reduce the number of arguments in the constructors. 

:+1:

> But how should it be possible to extract "functions that actually create Boogie code"? This is most of the work done in the implementations of FunctionModelProvider and even if it is possible, it is really much work.

After looking at all the methods that are now in the subclasses: yes, they should stay there. Its just that I expect some similarities between methods in different headers, and hence the question might arise: when do we want to share code between these? This PR is probably a good time to provide some guidance (in the form of comments somewhere).

maul-esel:
> Yes, I also thought about this. It should be possible, we just need to move the instatiations of FMPs outside of the `StandardFunctionHandler` (maybe this is a good idea anyway) and add another method that yields a map of defined types. I will have a look, but maybe we should move this to a separate PR.

I agree we don't have to do everything in this PR. But perhaps moving the instantiation of these classes (I think you did this already) and giving them a more general name (as we intend to provide other things besides function models) would be a good foundation.

schuessf:
Okay, so I reworked this PR again:
* `FunctionModelProvider` was split into an interface and a helper class.
* There is now an interface `ILibraryModel` with methods for functions (as before), but also types.
* The type definitions were moved from `TypeHandler` to the matchding implementation of `ILibraryModel` (e.g., the handling of `pthread_t` is now in `PthreadLibraryModel`)
* `StandardFunctionHandler` was removed and replaced by `LibraryModelHandler` that can also handles types (creates a map internally, similar to functions)
* `TypeHandler` uses `LibraryModelHandler` for the translation of types (this required a method `setLibraryModelHandler`, as some implementations of `ILibraryModel` already require `TypeHandler`)

I am still not quite sure about the naming (maybe you have some comments there 😉) and the documentation and header is still missing, but I will add them, once we agreed about these changes in general.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/standardfunctions/StandardFunctionHandler.java
schuessf: The list of all these objects was now moved into a method of [CHandler](https://github.com/ultimate-pa/ultimate/pull/714/files#diff-eac2932341760ee7522f505fdef04efb90ec6eaf8e279341fb02af1b032730e1R609-R637). Do you still think, we should move this to a separate object?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/StdlibLibraryModel.java
```diff
@@ -0,0 +1,500 @@
...
+				IASTBinaryExpression.op_greaterThan, expr, resultType,
+				mTypeSizes.constructLiteralForIntegerType(loc, resultType, BigInteger.ZERO), resultType);
+		final Expression negated =
+				mExpressionTranslation.constructUnaryExpression(loc, IASTUnaryExpression.op_minus, expr, resultType);
+		final Expression iteExpression = ExpressionFactory.constructIfThenElseExpression(loc, positive, expr, negated);
+		return builder.setLrValue(new RValue(iteExpression, resultType)).build();
+	}
+
+	private TypesResult handleSize_t(final IASTNamedTypeSpecifier node, final ILocation loc) {
+		return new TypesResult(new PrimitiveType(loc, BoogieType.TYPE_REAL, SFO.REAL), node.isConst(), false,
```
schuessf: I just found that we use Boogie's `real` as a type here, but `size_t` is an `int` type of course. This seems like a bug that has been present quite long in `TypeHandler` (not sure if it actually triggered something)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/PthreadLibraryModel.java
```diff
@@ -0,0 +1,500 @@
...
+		erb.addAllExceptLrValue(arg1);
+		erb.addStatement(unlockMutex);
+		erb.setLrValue(new RValue(mTypeSizes.constructLiteralForIntegerType(loc, returnType, value),
+				new CPrimitive(CPrimitives.INT)));
+		return erb.build();
+	}
+
+	private TypesResult handlePthread_t(final IASTNamedTypeSpecifier node, final ILocation loc) {
+		final var cType = mTypeHandler.getThreadIdType();
+		return new TypesResult(mTypeHandler.cType2AstType(loc, cType), node.isConst(), false, cType);
```
schuessf: Currently, `TypeModel` is quite similar to `FunctionModel`: It contains a name and a function to produce a `Result` (should be always a `TypesResult`). But it looks like it could be simplified. All those function seem to produce some `CType` `cType` and return `new TypesResult(mTypeHandler.cType2AstType(loc, cType), node.isConst(), false, cType)`. If we don't need this flexibility to use a function, we could also directly use a `CType` in `TypeModel` instead.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/StdlibLibraryModel.java
danieldietsch: Huh, strange, but I would also think that this is a bug.

danieldietsch (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/PthreadLibraryModel.java
schuessf: So, I just tried replacing the functional interface with a simple `ICType` (see 89341c3). This simplifies the code, but we cannot define types with a custom `ASTType`, but I am not sure if we need that either. This removed also the bug above with `BoogieType.TYPE_REAL`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LibraryModelHandler.java
```diff
@@ -0,0 +1,167 @@
...
+	/**
+	 * Check if the given function has an "integrated" specification or implementation and return a {@link Result} that
+	 * contains a translation of the function if this is the case.
+	 *
+	 * Return null otherwise.
+	 *
+	 * @param main
+	 * @param node
+	 * @param astIdExpression
+	 * @return
```
maul-esel: Please fill in the blanks after there `@param` / `@return` tags

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LibraryModelHandler.java
```diff
@@ -0,0 +1,167 @@
...
+	 */
+	public Result translateStandardFunction(final IDispatcher main, final IASTFunctionCallExpression node,
+			final IASTIdExpression astIdExpression) {
+		assert node.getFunctionNameExpression() == astIdExpression
+				: "astIdExpression is not the name of the called function";
```
maul-esel: If this is required, why pass two parameters at all?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LibraryModelHandler.java
```diff
@@ -0,0 +1,167 @@
...
+		final String name = astIdExpression.getName().toString();
+
+		final IFunctionModelHandler functionModel = mFunctionModels.get(name);
+		if (functionModel != null) {
+			final String transformedName = mSymboltable.applyMultiparseRenaming(node.getContainingFilename(), name);
+			final IASTNode funDecl = mFunctionTable.get(transformedName);
+			if (funDecl instanceof IASTFunctionDefinition) {
+				// it is a function that already has a body
+				if (!mSettings.checkErrorFunction() || !"reach_error".equals(transformedName)) {
```
maul-esel: Could `reach_error` not also be handled by `SvcompLibraryModel`?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LibraryModelHandler.java
```diff
@@ -0,0 +1,167 @@
...
+		if (old != null) {
+			throw new AssertionError("Accidentally overwrote definition for " + key);
+		}
+	}
+
+	/**
+	 * @author Daniel Dietsch (dietsch@informatik.uni-freiburg.de)
+	 */
+	@FunctionalInterface
+	interface IFunctionModelHandler {
```
maul-esel, line 170: Please document this interface.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LinuxLibraryModel.java
```diff
@@ -0,0 +1,252 @@
...
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.result.ExpressionResultTransformer;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.result.LocalLValue;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.result.RValue;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.result.Result;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.util.SFO.AUXVAR;
+import de.uni_freiburg.informatik.ultimate.core.lib.models.annotation.Overapprox;
+import de.uni_freiburg.informatik.ultimate.core.model.models.ILocation;
+
+/**
+ * Modelling of various C functions form Linux, see https://man7.org/linux/man-pages/
```
maul-esel: `from` ?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/FunctionModelHelper.java
```diff
@@ -0,0 +1,358 @@
...
+			final IASTInitializerClause[] arguments) {
+		if (arguments.length != expectedArgs) {
+			throw new IncorrectSyntaxException(loc, name + " is expected to have " + expectedArgs
+					+ " arguments, but was called with " + arguments.length);
+		}
+	}
+
+	/**
+	 * Handle a function call by dispatching all arguments and then calling a function with no arguments that has the
+	 * name of the function and is marked with the {@link Overapprox} annotation. Additionally it is assumed that the
```
maul-esel: This documentation is no longer correct, as far as I can determine. Could you please update it?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/SvcompLibraryModel.java
```diff
@@ -0,0 +1,294 @@
...
+		result.add(new FunctionModel("__VERIFIER_nondet_ulong",
+				(main, node, loc, name) -> handleVerifierNonDet(main, loc, new CPrimitive(CPrimitives.ULONG))));
+		result.add(new FunctionModel("__VERIFIER_nondet_ulonglong",
+				(main, node, loc, name) -> handleVerifierNonDet(main, loc, new CPrimitive(CPrimitives.ULONGLONG))));
+		result.add(new FunctionModel("__VERIFIER_nondet_uint128",
+				(main, node, loc, name) -> handleVerifierNonDet(main, loc, new CPrimitive(CPrimitives.UINT128))));
+		result.add(new FunctionModel("__VERIFIER_nondet_ushort",
+				(main, node, loc, name) -> handleVerifierNonDet(main, loc, new CPrimitive(CPrimitives.USHORT))));
+
+		// TODO: These are no predefined types, thus the return value may depend on the benchmark
```
maul-esel, line 159: We should get rid of such special handling. If there are still benchmarks with these, either they have to be changed, or the newly proposed extension of `__VERIFIER_nondet` to more types should cover them.

(Not an issue for this PR though.)

maul-esel (approved):
Thanks again for the updates.

I've looked through the changes in more detail, but found only small issues.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LibraryModelHandler.java
schuessf: Good point, I simply copied this from `StandardFunctionHandler`, but I will check it again.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LibraryModelHandler.java
schuessf: It is handled by `SvcompLibraryModel`. This code here is just responsible that a warning is shown if `reach_error` is defined (as done in many tasks of SV-COMP). It was just moved from `StandardFunctionHandler`, but I might be useful to always show this warning, not just for `reach_error`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/SvcompLibraryModel.java
schuessf: According to the [SV-COMP rules](https://sv-comp.sosy-lab.org/2025/rules.php), the functions are allowed for some "non-standard" types (e.g., `size_t`, `loff_t`, `u32`), while not for others (e.g., `u8`, `u16`), even though such functions occur in the benchmarks. Therefore, either the rules or the benchmarks should be fixed (which requires some effort), and thus I suggest to keep this for now (this was also present in the `StandardFunctionHandler` previously).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LibraryModelHandler.java
schuessf: Okay, I changed this in 21e8758, but this actually changes the behaviour: If a function from a standard library is redefined, we no longer do an actual call to this function, but take our model from this standard function instead and show a warning. But if I understand the C standard correctly, this is undefined behaviour anyways (C11 7.1.3.2):

> No other identifiers are reserved. If the program declares or defines an identifier in a
context in which it is reserved (other than as allowed by 7.1.4), or defines a reserved
identifier as a macro name, the behavior is undefined.

Previouly, we only overrode `reach_error` for unreach-call and no function from a standard library.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LibraryModelHandler.java
schuessf: This argument was indeed not necessary, therefore I removed it (740c4ea).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/LibraryModelHandler.java
maul-esel: Ah, that is risky. I'm sceptical that it is a good idea to override functions that are actually defined in the C code.
- This may for now be ok for standard library functions.
- But it already gets somewhat critical with SV-COMP functions -- redefining them is certainly not undefined. Thus the warning for `reach_error`.
- E.g. for linux functions it may depend on the verified program. If we are analysing some kind of kernel implementation, we should definitely not substitute parts of it with our model.

schuessf: I agree, this was not done fully on purpose. Therefore I reverted this change and just clarified the existing code instead (9030a5b).

PR #717, opened 2025-05-06
Produce invariants at labels

Heizmann:


maul-esel (approved):

schuessf (approved):
Looks good!
With this change you removed the only calls to `IcfgUtils.getPotentialCycleProgramPoints` , therefore this method could be removed (not necessarily in this PR).
Also, this method seems to be the only place, where we check for `LoopEntryType.GOTO`, therefore I wonder if we can remove `LoopEntryType` at all and also the annotation for gotos. This would also make sense, since we want invariants at labels and not at gotos. But this is also something we can check after merging this PR.

PR #719, opened 2025-05-07
Group settings of C translation

Heizmann:
Screenshot_20250507_203353


danieldietsch (approved):

bahnwaerter (approved):

maul-esel (approved):
:+1: for grouping them, and splitting the "Semantics" group (can change the expected verdict) from "Optimizations" (should not change the expected verdict).

I'm still not entirely convinced of the "Optimizations" label -- e.g. why is the choice of integer vs bitvector translation an optimization? Which value is the "optimized" one? And generally, if these are optimizations (which to me sounds like some value is the "optimized" one, the others are sub-optimal), why would I ever want to choose sub-optimal values for these settings (except for internal benchmarking)?

But if you are convinced Optimizations is the right term, I'm fine with it.

schuessf (approved):
Thanks for this change! The settings look much better now 👍
As discussed, we should maybe rename/merge some settings and give them a description, but this should be done in another PR.

Heizmann:
> But if you are convinced Optimizations is the right term, I'm fine with it.

I am not convinced. I am open for other ideas. We can change that later.

PR #720, opened 2025-05-08
Extend library models to support unpreprocessed C programs

schuessf:
Currently, we only support preprocessed C programs. While handling multi-files or all kinds of macros in general can be quite complicated, this PR allows at least includes of supported system libraries (e.g., `#include <stdlib.h>`) without the need for preprocessing. This way we should be already prepared for the upcoming SV-COMP, as the rules should be relaxed to allow such unpreprocessed tasks with only system-includes 😉

Therefore, this PR includes the following changes:
* The method `getConstantModels` was added to the interface `ILibraryModel` (similar to the change for types in #714). There was already support for some constants in [CHandler](https://github.com/ultimate-pa/ultimate/blob/37e010c/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java#L1743) that were just moved to the matching implementation of `ILibraryModel` (e.g., `NULL` is now handled in `StdlibLibraryModel`).
* Support for additional constants was added (e.g., `PTHREAD_MUTEX_INITIALIZER` is now handled in `PthreadLibraryModel`).
* Additional types that are defined in the standard headers are also supported (e.g., `VariadicLibraryModel` does not only handle `__builtin_va_list`, but also `va_list`, `AtomicLibraryModel` can now handle types like `atomic_int`).
* During preprocessing function macros are replaced with some compiler builtin functions (e.g., `isless` is replaced with `__builtin_isless`), therefore we only handled the latter before. In this PR, support for such macros was also added (the same way as corresponding builtin functions are handled).


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/AtomicLibraryModel.java
```diff
@@ -73,10 +74,8 @@
  * @author Dominik Klumpp (klumpp@informatik.uni-freiburg.de)
  */
 public class AtomicLibraryModel implements ILibraryModel {
-	/**
-	 * See MEMORY_ORDER_SEQ_CST in stdatomic.h
-	 */
-	private static final int MEMORY_ORDER_SEQ_CST = 5;
+	private static final String[] MEMORY_ORDER = { "memory_order_relaxed", "memory_order_consume",
+			"memory_order_acquire", "memory_order_release", "memory_order_acq_rel", "memory_order_seq_cst" };
```
bahnwaerter: Representing memory orders as a string array can be dangerous, since the `memory_order` from `stdatomic.h` is represented as an `enum`. Wouldn't it be better to use an `enum` with an attached string value in Java?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/AtomicLibraryModel.java
```diff
@@ -456,7 +450,7 @@ private ExpressionResult applyMemoryOrders(final ILocation loc, final Expression
 		// create condition checking whether all memory orders are supported
 		final CPrimitive intType = new CPrimitive(CPrimitives.INT);
 		final Expression seqCst = mExpressionTranslation.constructLiteralForIntegerType(loc, intType,
-				BigInteger.valueOf(MEMORY_ORDER_SEQ_CST));
+				BigInteger.valueOf(Arrays.asList(MEMORY_ORDER).indexOf("memory_order_seq_cst")));
```
bahnwaerter: A Java enum would be more suitable for representation here to access the numbered entries here. With a string array, this can quickly become dangerous if the order of the strings in the array is accidentally changed.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/MathLibraryModel.java
```diff
@@ -509,6 +534,26 @@ private Result handleFloatBuiltinIsLessGreater(final IDispatcher main, final IAS
 
 	@Override
 	public Collection<TypeModel> getTypeModels() {
-		return List.of();
+		return List.of(
+				// most efficient floating-point type at least as wide as float -> We choose float
+				new TypeModel("float_t", new CPrimitive(CPrimitives.FLOAT)),
+				// most efficient floating-point type at least as wide as double -> We choose double
+				new TypeModel("double_t", new CPrimitive(CPrimitives.DOUBLE)));
```
bahnwaerter, line 541: The two types are defined by a `typedef` to an implementation-defined type. Typically, this is the classic type `float` or `double`. In the future, we can have the implementation determined automatically via `typedef` resolution and don't need to hard-code the two special types. This already works in simple cases. Will it work here as well?

In the GNU C Library 2.41, the `typedef`s for `float_t` and `double_t` from `math.h` look like this:
```
# if __GLIBC_FLT_EVAL_METHOD == 0 || __GLIBC_FLT_EVAL_METHOD == 16
typedef float float_t;
typedef double double_t;
# elif __GLIBC_FLT_EVAL_METHOD == 1
typedef double float_t;
typedef double double_t;
# elif __GLIBC_FLT_EVAL_METHOD == 2
typedef long double float_t;
typedef long double double_t;
# elif __GLIBC_FLT_EVAL_METHOD == 32
typedef _Float32 float_t;
typedef double double_t;
# elif __GLIBC_FLT_EVAL_METHOD == 33
typedef _Float32x float_t;
typedef _Float32x double_t;
# elif __GLIBC_FLT_EVAL_METHOD == 64
typedef _Float64 float_t;
typedef _Float64 double_t;
# elif __GLIBC_FLT_EVAL_METHOD == 65
typedef _Float64x float_t;
typedef _Float64x double_t;
# elif __GLIBC_FLT_EVAL_METHOD == 128
typedef _Float128 float_t;
typedef _Float128 double_t;
# elif __GLIBC_FLT_EVAL_METHOD == 129
typedef _Float128x float_t;
typedef _Float128x double_t;
# else
#  error "Unknown __GLIBC_FLT_EVAL_METHOD"
# endif
#endif
```
In such cases, SV-COMP must also specify C preprocessor defines in the task definitions for non-preprocessed C programs in the future.

bahnwaerter (changes requested):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/AtomicLibraryModel.java
schuessf: To define the `CEnum` `memory_order`, I need exactly such a string array, therefore this was the simplest solution.
If I would use an enum instead, this would be more complex and I cannot really see the advantage.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/AtomicLibraryModel.java
schuessf: But the same could happen, if you accidentally change the order in the enum, right?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/MathLibraryModel.java
schuessf: Exactly, these types are implementation defined. Therefore we just choose one type (`float` / `double`) as an implementation to match the standard.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/AtomicLibraryModel.java
maul-esel: I would also propose an enum, but additionally, avoid using `indexOf` to get the literal. The enum constructor could take a string argument for the name, and an integer for the literal. Then we can have getters on the enum to retrieve them.

Thinking long-term, we may also want to avoid using literals at some point, and just use symbolic Boogie constants instead. This would be more general, and it would avoid proofs in which e.g. a mysterious `5` appears, when really what we mean is `memory_order_seq_cst`.

bahnwaerter: > But the same could happen, if you accidentally change the order in the enum, right?

Yes, that could happen there too. But it is less likely that a tool like a code formatter changes the order of the elements (since they are numbered implicitly).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/MathLibraryModel.java
bahnwaerter: For now, that's perfect. We can still adjust this in the future.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/AtomicLibraryModel.java
schuessf: > I would also propose an enum, but additionally, avoid using `indexOf` to get the literal. The enum constructor could take a string argument for the name, and an integer for the literal. Then we can have getters on the enum to retrieve them.

Okay, I chnaged it as suggested. Retrieving the field names in the correct order requires more work now, therefore `getTypeModels()` is more complicated now -- thus I am still not quite convinced that this is a better solution 😉

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/PostProcessor.java
```diff
@@ -377,10 +378,6 @@ private static ArrayList<Declaration> declareCurrentRoundingModeVar(final ILocat
 	}
 
 	private ArrayList<Declaration> createUltimateSetCurrentRoundingProcedure(final ILocation loc, final IASTNode hook) {
```
maul-esel, line 383: The rounding-mode code in this class also looks like it should be part of a library model. It seems like it would require a further extension of the design though:

- It apparently declares a global variable, which library models do not yet support. Maybe some kind of `ILibraryStateModelHandler` is needed?
- `::postProcess()` only includes this code if `mTypeHandler.areFloatingTypesNeeded()` returns `true`. IIRC this was quite crucial, as otherwise the trace checks fail for any solver that doesn't support floating points; even when they are not at all needed in the program. Perhaps specialized `IFunctionModelHandler` / `ILibraryStateModelHandler` / `ILibraryTypeModelHandler` implementations could do a similar check?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/AtomicLibraryModel.java
```diff
@@ -480,7 +493,29 @@ private ExpressionResult applyMemoryOrders(final ILocation loc, final Expression
...
+				new TypeModel("atomic_ushort", CPrimitive.constructAtomicType(CPrimitives.USHORT)),
+				new TypeModel("atomic_int", CPrimitive.constructAtomicType(CPrimitives.INT)),
+				new TypeModel("atomic_uint", CPrimitive.constructAtomicType(CPrimitives.UINT)),
+				new TypeModel("atomic_long", CPrimitive.constructAtomicType(CPrimitives.LONG)),
+				new TypeModel("atomic_ulong", CPrimitive.constructAtomicType(CPrimitives.ULONG)),
+				new TypeModel("atomic_llong", CPrimitive.constructAtomicType(CPrimitives.LONGLONG)),
+				new TypeModel("atomic_ullong", CPrimitive.constructAtomicType(CPrimitives.ULONGLONG)),
+				new TypeModel("memory_order",
+						new CEnum("memory_order",
+								Arrays.stream(MemoryOrder.values()).sorted(Comparator.comparing(MemoryOrder::getValue))
```
maul-esel: Regarding the discussion of how to model this enum: I am a bit confused now -- what role does the order of the names in the `CEnum` instance play? As you do provide constant models, I guess the fields are not implicitly numbered from `1` to `n`. Is the order at all relevant?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/ILibraryModel.java
```diff
@@ -27,16 +27,63 @@
...
+	@FunctionalInterface
+	interface IConstantModelHandler {
+		/**
+		 * Models the constant as an {@link ExpressionResult}.
+		 *
+		 * @param loc
+		 *            A location.
+		 * @return An {@link ExpressionResult} as a model of the constant.
+		 */
+		ExpressionResult handleConstant(ILocation loc);
```
maul-esel, line 84: Is there a specific reason why the return type here is more restrictive than in `IFunctionModelHandler`? In principle, it makes sense to me that for a constant, an expression is returned. But what are the functions for which this is not the case? Could any of these scenarios also apply to constants?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/ILibraryModel.java
```diff
@@ -52,24 +99,47 @@ public record TypeModel(String typeName, ICType cType) {
 		// empty
 	}
 
+	/**
+	 * Model of a predefined constant, consisting of the name of the type and our translated model, i.e., a
+	 * {@code ICType}) and an {@code Expression}.
```
maul-esel: This doesn't seem to match the record declaration.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/MathLibraryModel.java
maul-esel: I guess how we will resolve this (in the future) depends on whether some fixed standard library header files are given (then, we probably want the `#define`s and resolve against those headers) or not (then, we may want to have settings, similar to how we can now define the sizes of various types).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/StdintLibraryModel.java
```diff
@@ -0,0 +1,98 @@
...
+				// maximum width unsigned integer type
+				new TypeModel("uintmax_t", uint64),
+
+				// unsigned integer type capable of holding a pointer
+				new TypeModel("uintptr_t", uint32));
+	}
+
+	@Override
+	public Collection<ConstantModel> getConstantModels() {
+		// TODO: Add INTN_MIN etc.
```
maul-esel, line 94: Is this TODO planned for later, or this PR?

maul-esel (approved):
:+1: thanks for getting us ready for the future :) The PR looks good. I've left some questions and comments, but (except maybe the doc issue) all of these may potentially be delayed to future iterations.

Let me know if we should run further evaluations against this branch before it is merged.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/PostProcessor.java
schuessf: Yes, I think that is more complicated -- and I am also not really familiar with this class.
For now, my goal was simply to support constants like `FE_DOWNWARD` instead of the literals that result from GCC preprocessing. But it is a good idea to try to move this code also to the library model.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/AtomicLibraryModel.java
schuessf: Oh, you are right we actually don't care about the fields or their order at all in `CEnum`, I somehow confused this with `CStructOrUnion` (where we lookup the fields, but the order still does not matter).
Instead of an `enum`, we could define `memory_order` also using just `int`, but I am not sure if that is better.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/ILibraryModel.java
schuessf: What do you mean with "more restrictive"? Because `ExpressionResult` is returned here instead of just `Result`? An arbitrary `Result` does not really make sense for constants, as an `LRValue` should be returned, which is only possible for an `ExpressionResult`. On the either hand, to model `void` function it might be perfectly fine to return a `SkipResult`.

And it does also not work to return just a  `LRValue`, as the handling of  `__FUNCTION__` in (`GccBuiltinLibraryModel`) requires an aux-var and maybe we need the flexibility to use additional Boogie statements in the future.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/ILibraryModel.java
schuessf: Thanks for the hint! I changed the signature, but forgot to update the documentation.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/MathLibraryModel.java
schuessf: Yes, if somewhere else in the file (or multiple files, if we support this) a type is defined, we still use this instead of our library model. I changed this in 787bd85, otherwise it can be an issue for something like `typedef double size_t;`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/StdintLibraryModel.java
schuessf: For later 😉

maul-esel:
Many constants (perhaps mostly compiler built-ins) are also defined in here: https://github.com/ultimate-pa/ultimate/blob/c237c589506a02c25c3eb69cdbfb97ddaaa01ac4/trunk/source/CDTParser/src/de/uni_freiburg/informatik/ultimate/cdt/parser/GccStaticLanguageSettingsProvider.java#L52-L57

Some of them seem to overlap with constants defined in library models (e.g. the memory order constants in the snippet above). Others seem like they should change depending on the architecture settings (e.g. `__SIZEOF_POINTER__`). Yet others seem contradictory, like defining both `_WIN32` and `_WIN64` as `1`.

How do we proceed with these?

trunk/source/CDTParser/src/de/uni_freiburg/informatik/ultimate/cdt/parser/GccStaticLanguageSettingsProvider.java
```diff
@@ -262,6 +262,7 @@ public GccStaticLanguageSettingsProvider() {
 				createEntry("_stdcall", "__attribute__((__stdcall__))"),
 				createEntry("_thiscall", "__attribute__((__thiscall__))"),
 				createEntry("__builtin_va_arg(ap,type)", "*((typeof(type) *)((ap += sizeof(type)) - sizeof(type)))"),
+				createEntry("va_arg(ap,type)", "*((typeof(type) *)((ap += sizeof(type)) - sizeof(type)))"),
```
maul-esel, line 257: Could you please add comments here to distinguish which entries (such as this one) have been added by us manually (and for what reason), as opposed to the remaining ones which, according to the class documentation, have at some point been generated automatically?

schuessf:
> Some of them seem to overlap with constants defined in library models (e.g. the memory order constants in the snippet above). Others seem like they should change depending on the architecture settings (e.g. `__SIZEOF_POINTER__`). Yet others seem contradictory, like defining both `_WIN32` and `_WIN64` as `1`.
> 
> How do we proceed with these?

They don't quite overlap: The entries of `memory_order` are called `memory_order_seq_cst` etc., `__ATOMIC_SEQ_CST` is just a constant known by GCC to define `memory_order_seq_cst`, see in [stdatomic.h by GCC](https://github.com/gcc-mirror/gcc/blob/master/gcc/ginclude/stdatomic.h#L29):
```c
typedef enum
  {
    memory_order_relaxed = __ATOMIC_RELAXED,
    memory_order_consume = __ATOMIC_CONSUME,
    memory_order_acquire = __ATOMIC_ACQUIRE,
    memory_order_release = __ATOMIC_RELEASE,
    memory_order_acq_rel = __ATOMIC_ACQ_REL,
    memory_order_seq_cst = __ATOMIC_SEQ_CST
  } memory_order;
```

But I am still not quite sure, if we actually need any of these entries here. Maybe @danieldietsch knows more about it, since he added this class.

maul-esel:
> They don't quite overlap: [...]

They are not the same constants, but they "overlap" in the sense that it seems like a good idea to define `memory_order_relaxed` and `__ATOMIC_RELAXED` in the same place, doesn't it?

maul-esel:
Many of these macros seem to be documented [here](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html) as GNU C extensions. As I read this page, it seems to be perfectly supported to use these in C programs (_edit_: for some of them. For others, it says "do not use directly"). Therefore, I think we should support them as documented, rather than hard-coding the values we currently have in `GccStaticLanguageSettingsProvider`.

schuessf:
> Many of these macros seem to be documented [here](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html) as GNU C extensions. As I read this page, it seems to be perfectly supported to use these in C programs. Therefore, I think we should support them as documented, rather than hard-coding the values we currently have in `GccStaticLanguageSettingsProvider`.

Sure, they are perfectly fine to use in C programs that should be compiled using GCC. And I agree that we should model them as described in GNU C. I think that they don't occur in the current SV-COMP tasks (but I can check that again). So I think we should really just remove all those macros `in GccStaticLanguageSettingsProvider` and handle them in the corresponding `ILibraryModel`. But I would suggest to add them only on demand (as we also do for library functions), I will just start with some of them (like the mentioned `__ATOMIC_RELAXED`).

schuessf:
> > Many of these macros seem to be documented [here](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html) as GNU C extensions. As I read this page, it seems to be perfectly supported to use these in C programs. Therefore, I think we should support them as documented, rather than hard-coding the values we currently have in `GccStaticLanguageSettingsProvider`.
> 
> Sure, they are perfectly fine to use in C programs that should be compiled using GCC. And I agree that we should model them as described in GNU C. I think that they don't occur in the current SV-COMP tasks (but I can check that again). So I think we should really just remove all those macros `in GccStaticLanguageSettingsProvider` and handle them in the corresponding `ILibraryModel`. But I would suggest to add them only on demand (as we also do for library functions), I will just start with some of them (like the mentioned `__ATOMIC_RELAXED`).

But again: Maybe @danieldietsch can tell, why all these macros in `GccStaticLanguageSettingsProvider` are necessary and where they come from 😉

danieldietsch:
Yes, they are there because they are necessary if one wants to parse a certain gcc version with a certain set of header files. The CDT parser can extract them from system headers, and as we needed a way to provide them without scanning the actual system headers, we just hardcoded them. 

The correct way to get those would be to retrieve them from the include path, but that is a lot more work. We would at least need a switch to disable header parsing when running in (the old) SVCOMP mode.

schuessf:
> Yes, they are there because they are necessary if one wants to parse a certain gcc version with a certain set of header files. The CDT parser can extract them from system headers, and as we needed a way to provide them without scanning the actual system headers, we just hardcoded them.
> 
> The correct way to get those would be to retrieve them from the include path, but that is a lot more work. We would at least need a switch to disable header parsing when running in (the old) SVCOMP mode.

Sure, if we want to be able to translate C files that use such macros, we need to handle them somewhere.
But as we have this interface `ILibraryModel` now, wouldn't this be a better place for such macros? This PR also attempts to support macros and types from certain system headers without explicitly including the header files.

Also I just checked: None of the macros in `GccStaticLanguageSettingsProvider` are currently used in SV-COMP. Therefore I would suggest to remove them and add them to the matching implementation of `ILibraryModel` on demand.

danieldietsch:
These macros are required by some system headers themselves and provided by whatever gcc you are using (afair at least).

maul-esel:
I'm getting a bit confused here. My understanding is that we are discussing macros that are _not_ defined in any standard header file, but directly built into the compiler. This matches my interpretation of the statement

> These macros are required by some system headers themselves and provided by whatever gcc you are using (afair at least).

but seems to contradict the statement that

> The CDT parser can extract them from system headers, [...] The correct way to get those would be to retrieve them from the include path

If my understanding is correct, we would need builtin support for these macros in Ultimate, no matter whether we (a) analyze pre-processed code, (b) analyze non-preprocessed code and load standard header files from a given include directory, or (c) analyze non-preprocessed code _without_ loading standard header files from disk.

The issues now seem to be:
- @schuessf has started adding support for standard header-defined macros (constants) via `ILibraryModel`. Is there any argument against supporting these builtin macros in the same way, and removing (most of) them from `GccStaticLanguageSettings`?
- The existing implementation in `GccStaticLanguageSettings` seems to hard-code values for some macros that are inconsistent with other parts of Ultimate (particularly, the target architecture settings). If, for some reason, we cannot move these to `ILibraryModel`, we should still fix this issue.

danieldietsch:
You are right, its confusing because I wrote it confusing. 

> These macros are required by some system headers themselves and provided by whatever gcc you are using (afair at least).

Is the right statement. 

Hence, your understanding is also correct. I don't think there is an argument against moving them, although you should check if they are somehow used directly by CDT.

schuessf:
> Hence, your understanding is also correct. I don't think there is an argument against moving them, although you should check if they are somehow used directly by CDT.

Great, for this PR I will just move the memory orders from `GccStaticLanguageSettings` to `AtomicLibraryModel`. I don't think this should be any issue for CDT.

Regarding the other macros from `GccStaticLanguageSettings`: I don't think they are necessary, if we don't want to translate the system headers that are provided by GCC, but this can be moved to a future PR.

maul-esel:
Could you maybe also add the following, as their use is not discouraged (like for some of the other macros), and to make sure their values are consistent with the target architecture settings?
image

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/PostProcessor.java
maul-esel: Maybe we can just add a TODO for now and come back to this later.

maul-esel (approved):

schuessf:
@danieldietsch @bahnwaerter Do you have anything to add here or can this PR be merged?

danieldietsch:
Go ahead :)

PR #722, opened 2025-05-08
Fix nondeterminism bug in BlockEncodingV2

Heizmann:
Implemented fix: HashSet<EdgeConstructor> constructors becomes ArrayList<EdgeConstructor>
Considered alternative: Give EdgeConstructor proper hashCode and equals methods. However, it seems to me that MinimizeStatesMultiEdgeMultiNode#getEdgePairs which provides the elements for constructors ensures that there is only one EdgeConstructor per edge pair.


danieldietsch:
Looks plausible -- but why not LinkedHashSet?

danieldietsch (approved):

Heizmann:
> Looks plausible -- but why not LinkedHashSet?

Because there are no operations on `constructors` that could benefit from set operations.

PR #723, opened 2025-05-10
Bugfix: Insert continue label before loop

Heizmann:
When translating while loops that have a continue, we need an additional label (since Boogie does not have a continue statement). Previously, this label was inserted at the beginning of the loop. However, since we want calculate the correct invariant also for continues and we also want to check an invariant after a continue, the label is inserted just before the loop.


schuessf:
The nightly looked fine, so I suggest to merge this.

Issue #667, opened 2024-05-23
Bug: Ultimate C parser crashes when there are Doxygen group comments in the C code


bahnwaerter:
The current C parser crashes when there are Doxygen group comments in the C code. This issue can be reproduced with the latest version of Ultimate Automizer (0.2.4-dev-1ececfe with OpenJDK 11.0.23) using the following input:
```
//#SAFE

typedef long unsigned int __uint32_t;
typedef __uint32_t uint32_t;

/** \ingroup  CMSIS_Core_FunctionInterface
    \defgroup CMSIS_Core_RegAccFunctions CMSIS Core Register Access Functions
  @{
 */

/**
  \brief   Get Control Register
  \details Returns the content of the Control Register.
  \return               Control Register value
 */
static inline uint32_t __get_CONTROL(void)
{
  uint32_t result;

  __asm volatile ("MRS %0, control" : "=r" (result) );
  return(result);
}

/**
  \brief   Set Control Register
  \details Writes the given value to the Control Register.
  \param [in]    control  Control Register value to set
 */
static inline void __set_CONTROL(uint32_t control)
{
  __asm volatile ("MSR control, %0" : : "r" (control) : "memory");
}

/*@} end of CMSIS_Core_RegAccFunctions */

int main(void)
{
  uint32_t control = __get_CONTROL();
  assert(control >= 0x00000000);
  __set_CONTROL(control);
  return 0;
}
```

Considering this example input, the C parser crashes when trying to parse the end marker of a Doxygen group comment (`/*@} end of CMSIS_Core_RegAccFunctions */` in line 34). The following error message is reported during a crash:
```
[...]
[2024-05-23 17:51:32,051 WARN  L194 CommentParser]: Syntax Error: #120[rbrace](0/1 - 0/2)
[...]
[2024-05-23 17:51:32,421 INFO  L174    ResultUtil]:  * Results from CDTParser:
[2024-05-23 17:51:32,421 INFO  L198    ResultUtil]:   - SyntaxErrorResult [Line: 34]: Incorrect Syntax
[2024-05-23 17:51:32,422 INFO  L202    ResultUtil]:     Syntax Error: #120[rbrace](0/1 - 0/2)
```

In this example, the C parser for the comments expects ACSL specifications and instead gets Doxygen group comments that violate the grammar of ACSL. Unfortunately, the parser cannot ignore the Doxygen group comments.

schuessf:
[...]

PR #724, opened 2025-05-13
Add setting whether ACSL assertions should be checked

schuessf:
This PR adds a setting for `CACSL2BoogieTranslator` (in the group "specification") whether ACSL assertions should be checked or not. For compatibility, I enabled this setting by default, as previously ACSL was always checked.
Note: For now, we still parse ACSL even if the setting is disabled. This is an issue, if a comment contains unsupported ACSL (or starts with @ and contains no ACSL at all #667). But disabling parsing could be more complicated, as it is done in `CDTParser` instead of `CACSL2BoogieTranslator`.

### DIFF
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
@@ -95,6 +95,7 @@ import de.uni_freiburg.informatik.ultimate.core.lib.models.annotation.Overapprox
 import de.uni_freiburg.informatik.ultimate.core.model.models.ILocation;
 import de.uni_freiburg.informatik.ultimate.core.model.models.annotation.Spec;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ACSLNode;
+import de.uni_freiburg.informatik.ultimate.model.acsl.ast.ACSLProblemNode;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ast.ACSLResultExpression;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ast.ArrayAccessExpression;
 import de.uni_freiburg.informatik.ultimate.model.acsl.ast.Assertion;
@@ -950,4 +951,9 @@ public class ACSLHandler implements IACSLHandler {
 						CPointer.voidPointer()));
 	}
 
+	@Override
+	public Result visit(final IDispatcher main, final ACSLProblemNode node) {
+		final ILocation loc = mLocationFactory.createACSLLocation(node);
+		throw new UnsupportedSyntaxException(loc, "Error during parsing of ACSL (" + node.getErrorMessage() + ")");
+	}
 }
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ trunk/examples/concurrent/bpl/weaver-benchmarks/weaver
 /trunk/source/.sonarlint/*
 /trunk/source/*.log
 
+/trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/ACSLProblemNode.java
 /trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/ACSLResultExpression.java
 /trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/ACSLTransformer.java
 /trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/ACSLType.java
--- a/trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/acsl.ast
+++ b/trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ast/acsl.ast
@@ -261,3 +261,6 @@ ACSLType ::=
 Declaration ::=
 	type : ACSLType
 	name : String;
+
+ACSLProblemNode ::=
+	errorMessage : String;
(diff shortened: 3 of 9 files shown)


maul-esel:
:+1: thanks for adding this.

Regarding the parsing issues, we've also had this issue that @bahnwaerter reported that other comments not meant to be ACSL (I think it was doxygen) lead to parse errors. How should we handle this?
- Should we add a separate setting to CDTParser to (in the class `CommentParser`) skip parsing of ACSL?
  + Downside: This other setting would effectively override the setting added in this PR. That could lead to confusing behaviour.
- In case we cannot parse a particular comment (that we believe to be ACSL), instead of directly reporting a syntax error, could we instead annotate the C AST with a kind of "problem marker" (instead of an ACSL AST), and only report a syntax error if CACSL2BoogieTranslator encounters the problem marker?

schuessf:
> In case we cannot parse a particular comment (that we believe to be ACSL), instead of directly reporting a syntax error, could we instead annotate the C AST with a kind of "problem marker" (instead of an ACSL AST), and only report a syntax error if CACSL2BoogieTranslator encounters the problem marker?

I think this is a good idea, I will check how to implement this approach.

schuessf:
Okay, I implemented it this way: Instead of throwing reporting an error and cancelling the toolchain in `CommentParser`, we just create an `ACSLProblemNode`. Now we only crash when dispatching such an `ACSLProblemNode` in `ACSLHandler`, i.e., only if the new setting `Check ACSL annotations` is enabled.

Also, I just again, if this setting affects witness validation, where we also make use of our ACSL parsing. Witness validation still works as expected, even if `Check ACSL annotations` is disabled, since the dispatcher is called directly. And if the witness contains a syntax error, we still crash as before.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/ACSLHandler.java
```diff
@@ -950,4 +951,9 @@ public Result visit(final IDispatcher main, final NullPointer node) {
 						CPointer.voidPointer()));
 	}
 
+	@Override
+	public Result visit(final IDispatcher main, final ACSLProblemNode node) {
+		final ILocation loc = mLocationFactory.createACSLLocation(node);
+		throw new UnsupportedSyntaxException(loc, "Error during parsing of ACSL (" + node.getErrorMessage() + ")");
```
maul-esel, line 957: How is this syntax exception handled at runtime? How does the behaviour compare to what happened before these changes (do/did we get generic `ExceptionOrErrorResult`s, or some kind of `UnsupportedSyntaxResult`, etc.) ? And how does it relate to the `catch (final ParseException e1)` in line 2583 of `CHandler`?

danieldietsch: This will be catched and turned into an UnsupportedSyntaxResult.

schuessf: > How is this syntax exception handled at runtime? How does the behaviour compare to what happened before these changes (do/did we get generic ExceptionOrErrorResults, or some kind of UnsupportedSyntaxResult, etc.) ?

The result / output does not change much. The only difference is that we get an `UnsupportedSyntaxResult` (the exception is catched in [`MainTranslator`](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/MainTranslator.java#L124)) instead of a `SyntaxErrorResult`, but this is this result that is usually produced in the C translation.
Before:
```
WARN  L194          CommentParser]: Syntax Error: #45[gt](0/12 - 0/13)
[...]
INFO  L172             ResultUtil]:  --- Results ---
[...]
INFO  L174             ResultUtil]:  * Results from CDTParser:
INFO  L198             ResultUtil]:   - SyntaxErrorResult [Line: 3]: Incorrect Syntax
INFO  L202             ResultUtil]:     Syntax Error: #45[gt](0/12 - 0/13)
```
After:
```
ERROR L321         MainTranslator]: Unsupported Syntax: Error during parsing of ACSL (Syntax Error: #45[gt](0/12 - 0/13))
[...]
INFO  L172             ResultUtil]:  --- Results ---
[...]
INFO  L174             ResultUtil]:  * Results from de.uni_freiburg.informatik.ultimate.plugins.generator.cacsl2boogietranslator:
INFO  L198             ResultUtil]:   - UnsupportedSyntaxResult [Line: 3]: Unsupported Syntax
INFO  L202             ResultUtil]:     Error during parsing of ACSL (Syntax Error: #45[gt](0/12 - 0/13))
```

schuessf: > And how does it relate to the catch (final ParseException e1) in line 2583 of CHandler?

`IDispatcher::nextACSLStatement` is allowed to throw a `ParseException`, but the actual implementations never do so. I guess `ParseException` was used earlier and this is just a leftover. But I am not sure, if the exception handling for `ParseException` should be removed in this PR.

schuessf: I just removed `ParseException` in 9116035 (on dev).

schuessf:
I just rebased this PR again, can we merge it, or do you want some changes (@maul-esel @Heizmann @bahnwaerter @danieldietsch)?

danieldietsch (approved):

danieldietsch:
Merge :)

Heizmann (approved):

bahnwaerter (approved):

PR #729, opened 2025-06-06
Merge Parallel Trace Abstraction into Dev

MaxBarth95:
Updated

I prepared this branch to merge my parallel trace abstraction approach into main.
Feel free to take a look already, make suggestions, and ideally tell me how you want it to be refactored.
Unfortunately my changes  affects quite a few classes.

Before merging, I need to run a few more experiments to ensure there are no unwanted side effects. 

The following changes have been made:

-  ParallelNwaCegarLoop.java
Contains the main CEGAR loop

- CegarNwaWorkerThread.java
Is the corresponding class that does the trace check, interpolation, and generalization of the resulting automaton

- Some changes are made to the refinement strategy stuff. In the CegarNwaWorkerThread.java I can for example
(se commented code) decide to do LoopAcceleration after seeing a PathProgram 7 times.

 - IsEmptyParallel
The search now searches for a Prefix of a counterexamples that is different from all counterexamples in a Set given as parameter.  Then IsEmptyHeuristic is called with waypoints. Those waypoints are the Prefix and the IsEmptyHeuristic search follows those waypoints and after the last waypoints it searches for an accepting state.
It will not backtrack beyond the last waypoint. 

- TransferBetweenMainAndWorker
This is now the more elegant solution to our script problem. Each worker has one such transfer class and it manages all transfers in both direction. It transfers the CfgSmtToolkit, Automata, Counterexamples. To be precise, transformulas, IProgramVars everything that the CFGSmtToolkit touches

- Wintesses
To create a violation witness i need the sort of a value in the TypeSortTranslator. This sorts come from worker smtcfgtoolkit scripts. Since the TypeSortTranslator is created at the very start when translating to Boogie. I need a way to tell the TypeSortTranslator what the script is, that created this sort. I do this by passing it through the programexecution


trunk/source/Library-IcfgTransformer/src/de/uni_freiburg/informatik/ultimate/icfgtransformer/loopacceleration/jordan/JordanLoopAcceleration.java
```diff
@@ -304,14 +312,17 @@ private static ClosedFormOfUpdate computeClosedFormOfUpdate(final ManagedScript
...
+					.get(((HistoryRecordingScript) mgdScript.getScript()).transferTermToWorker(pv.getTermVariable())));
 		}
 		// Mapping in which that maps default TermVariables to their closed form (where
 		// variables are represented by inVars). Since array indices may also contain
 		// read-only variables we have to add from the inVars mapping.
 		final HashMap<TermVariable, Term> substitutionMapping = new HashMap<>(closedFormWithInvarsMap);
 		for (final IProgramVar pv : su.getReadonlyVars()) {
-			final Term oldValue = substitutionMapping.put(pv.getTermVariable(), inVars.get(pv));
+			final Term oldValue =
+					substitutionMapping.put((TermVariable) ((HistoryRecordingScript) mgdScript.getScript())
```
schuessf: I think we should look for a better solution here. We should not cast every script to `HistoryRecordingScript` to access the method `transferTermToWorker`.

Also, where is it necessary to transfer the terms? It looks like many locations are affected, maybe there is also a more elegant solution for that.

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/cfg/transitions/UnmodifiableTransFormula.java
```diff
@@ -227,8 +270,11 @@ private boolean doConstantConsistencyCheck() {
 		for (final IProgramConst programConsts : getNonTheoryConsts()) {
 			consistent &= !programConsts.getDefaultConstant().getFunction().isIntern();
 			assert consistent : "is theory symbol";
-			nonTheoryConstantTerms.add(programConsts.getDefaultConstant());
-			consistent &= constantsInFormula.contains(programConsts.getDefaultConstant());
+			nonTheoryConstantTerms.add((ApplicationTerm) ((HistoryRecordingScript) mMgdScript.getScript())
+					.transferTermToWorker(programConsts.getDefaultConstant()));
+			// assert mFormula.getTheory().equals(programConsts.getDefaultConstant().getTheory());
```
schuessf: Please delete outcommented code

trunk/source/Library-TraceCheckerUtils/META-INF/MANIFEST.MF
```diff
@@ -25,3 +25,6 @@ Require-Bundle: de.uni_freiburg.informatik.ultimate.lib.core,
  de.uni_freiburg.informatik.ultimate.smtinterpol,
  de.uni_freiburg.informatik.ultimate.smtsolver.external
 Automatic-Module-Name: de.uni.freiburg.informatik.ultimate.lib.tracecheckerutils
+Import-Package: de.uni_freiburg.informatik.ultimate.boogie.ast,
```
schuessf: I guess, we should not use any dependency to Boogie in this plugin. Why did you add it?

trunk/source/Library-TraceCheckerUtils/META-INF/MANIFEST.MF
```diff
@@ -25,3 +25,6 @@ Require-Bundle: de.uni_freiburg.informatik.ultimate.lib.core,
...
+ de.uni_freiburg.informatik.ultimate.plugins.generator.rcfgbuilder.cfg,
+ de.uni_freiburg.informatik.ultimate.plugins.generator.rcfgbuilder.preferences
```
schuessf: Why do we need settings of the `RCFGBuilder` here? This should be avoided.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/preferences/TraceAbstractionPreferenceInitializer.java
```diff
@@ -1066,4 +1111,29 @@ public enum OrderOfErrorLocations {
 	public enum PathProgramDumpStop {
 		NEVER, AFTER_FIRST_DUMP, BEFORE_FIRST_DUPLICATE
 	}
+
+	/**
+	 * Differnt modes of Test Generation
+	 *
+	 * @author Max Barth (Max.Barth@gmx.de)
+	 */
+	public enum TestGenerationMode {
```
schuessf: This has nothing to do with parallel trace abstraction, right?
Also, is this enum used anywhere?

schuessf (commented):
Thanks for your work! Running trace abstraction in parallel can be really useful to reduce the walltime in practice (where we don't care about the CPU time).
I did not yet look into this PR in detail, but I have some more general comments.
I can see that it is complicated to manage scripts for different threads, but the current solution does not seem nice (it introduces some possibly unsafe casts) and complicates much of the existing code. Therefore, we should rather look for a more elegant solution before merging this PR.

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/cfg/transitions/UnmodifiableTransFormula.java
MaxBarth95: deleted

trunk/source/Library-TraceCheckerUtils/META-INF/MANIFEST.MF
MaxBarth95: deleted

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/preferences/TraceAbstractionPreferenceInitializer.java
MaxBarth95: deleted

trunk/source/Library-TraceCheckerUtils/META-INF/MANIFEST.MF
MaxBarth95: deleted

trunk/source/Library-IcfgTransformer/src/de/uni_freiburg/informatik/ultimate/icfgtransformer/loopacceleration/jordan/JordanLoopAcceleration.java
MaxBarth95: Yes a better solution would be great, so far i couldnt think of any.

We basically need it everywhere:
- Its necessary before we checksat or assert
- Before we build a formula we have to make sure the components are from the same script
- When we get a termVar representing a IProgramVar 
- When we read a transformula, In, Out vars etc.
- When we create fresh predicates or use an old one

maul-esel:
Thanks for your work on this. Bringing parallelization to Ultimate is a big step, and it will be great to get it merged back into `dev`.

I think what we need here is an understanding for the architecture of your concurrent analysis, so that we can identify suitable interfaces between the workers and the main analysis thread. Then, the code can be refactored to handle all the business of this interaction -- primarily, it seems, the transferring of terms between scripts -- at well-defined points of these interfaces; such that separation of concerns is maintained.

We already have some components in Ultimate using different scripts, perhaps they could serve as inspiration:
- Trace checks use their own scripts. They must transfer transition formulas from the CFG script to the trace check script, and interpolants / counterexamples from the trace check script to the CFG script.
- GemCutter's independence checks use their own script. Here, predicates and transition formulas must be transferred to the script, and (since recently) commutativity condition predicates must be transferred back.

The current approach -- transferring these formulas in the most diverse set of classes, from loop acceleration over TermVarsFuns to SSA building code -- is extremely brittle. Who is to say that there is not some code path that is triggered only by few benchmarks, where the transferral is still missing? Any future change to Ultimate, e.g. some new kind of trace check, might introduce such cases. etc. I think we need a more robust solution.

MaxBarth95:
Sounds like a plan. You dont have to convince me that this needs some refactoring :) 
But before I do a refactoring i want us to agree on a suitable solution.

One thing we can do based on a suggestion of Matthias:
- Instead of using the HistoryRecordingScript we introduce a new Wrapper script.
This checks for every term and sort it sees, if its from its own theory and if not, it transfers from Main to Worker.
This will cleanup some clases, I'm guessing 30%.

- Additionally, we would need a new Substitution that transfers before its applies the substitution.

- Then we we are left with with the cases, where we have IProgramVars and their TermVariables and do renaming if I remember correctly

maul-esel:
That may be a possibility, but it introduces somewhat magic behaviour ;-) And you say it doesn't cover all cases.

Is there a reason against simply transferring every object at the point where the object is handed from the main thread to a worker, or from a worker to the main thread?

schuessf:
I was also thinking about something similar:
Would it work in priniciple to create a copy of the entire CFG, but with a fresh script for each worker? Then the the main thread could just do that and transfer the terms returned from the worker to the main script again.
Always creating an entirely new CFG would be really inefficient of course, but maybe we could just create a CFG implementation that wraps an exisiting CFG and transfers the terms on demand (and caches them).
If this does not work, you can tell me why and this way we might get towards a nicer solution 😉

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmpty.java
```diff
@@ -285,12 +289,41 @@ private IsEmpty(final AutomataLibraryServices services,
...
+			mLogger.info(startMessage());
+		}
+
+		mAcceptingRun = null;
+
+		if (mLogger.isInfoEnabled()) {
+			mLogger.info(exitMessage());
+		}
+	}
+
```
maul-esel, line 303: It seems this constructor is no longer needed? The only extra parameter is unused.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyHeuristic.java
```diff
@@ -83,7 +84,9 @@ public final class IsEmptyHeuristic<LETTER, STATE> extends UnaryNwaOperation<LET
 	private final Predicate<STATE> mIsForbiddenState;
 	private final NestedRun<LETTER, STATE> mAcceptingRun;
 	private final STATE mDummyEmptyStackState;
-
+	protected final List<Pair<STATE, LETTER>> mWayPoints;
+	boolean stillFollowingWayPoint = true;
+	boolean mDontVisitLoops = false;
```
maul-esel: - Please indicate visibility for these fields. Is there any reason to not make them private?
- The first one also does not conform to our naming conventions.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyHeuristic.java
```diff
@@ -64,6 +64,7 @@
 import de.uni_freiburg.informatik.ultimate.util.CoreUtil;
 import de.uni_freiburg.informatik.ultimate.util.HashUtils;
 import de.uni_freiburg.informatik.ultimate.util.datastructures.HashedPriorityQueue;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.Pair;
```
maul-esel, line 67: Are the changes in this class related to parallel TAR? The waypoint stuff seems witness-related?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyHeuristic.java
```diff
@@ -1133,6 +1205,30 @@ static <STATE, LETTER> IHeuristic<STATE, LETTER> getHeuristic(final AStarHeurist
...
+		/**
+		 * If state and trans are in any cex, we increase the cost
+		 *
+		 * The states and trans in the prefix have 0 cost
+		 *
+		 * @param <STATE>
+		 * @param <LETTER>
+		 * @return
+		 */
+		static <STATE, LETTER> IHeuristic<STATE, LETTER> getParallelHeuristic() {
```
maul-esel: Please document the meaning of this heuristic.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyParallel.java
```diff
@@ -0,0 +1,861 @@
...
+ *
+ * The idea is to have a preprocessing that finds a run to an arbitrary state not taken by any other counterexample.
+ * Then we call BFS to find a goal from this arbitrary state. At the end the runs are merged.
+ *
+ * Uses recursion for backtracking
+ *
+ * Non-terminating if every existing counterexample is in the set but the state space is infinite.
+ *
+ * TODO: Has issues with recursive function calls, sometimes finds path that doesnt exist. -> for now we check
+ * isAccepted after the search to find these cases
```
maul-esel: Has this bug been fixed?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyParallel.java
```diff
@@ -0,0 +1,861 @@
...
+		Set<STATE> callPreds = mVisitedCallPairs.get(state);
+		if (callPreds == null) {
+			callPreds = new HashSet<>();
+			mVisitedCallPairs.put(state, callPreds);
+		}
+		assert !callPreds.contains(stateK);
+		callPreds.add(stateK);
+	}
+
+	// TODO This does not work! we need a different data structure
```
maul-esel: Has this been fixed?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyParallel.java
```diff
@@ -0,0 +1,861 @@
...
+	 * nested word.
+	 */
+	@SuppressWarnings("squid:S1698")
+	private NestedRun<LETTER, STATE> getAcceptingRun(final DoubleDecker<STATE> startpair)
+			throws AutomataOperationCanceledException {
+		mInternalSubRun.clear();
+		mCallSubRun.clear();
+		mReturnPredStateK.clear();
+		// mSummaryReturnPred.clear();
+		// mSummaryReturnSymbol.clear();
```
maul-esel: please remove commented-out code

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyParallel.java
```diff
@@ -0,0 +1,861 @@
...
+
+	private class PQState {
+		Integer mScore;
+		STATE mState;
+		STATE mSucc;
+		STATE mStateK;
+		ArrayList<Integer> mCounterexamples = new ArrayList<>();
+		LETTER mSymbol;
+		boolean mCallTransition;
+		boolean mReturnTransition;
```
maul-esel: should these fields be private final?

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/boogie/TypeSortTranslator.java
```diff
@@ -121,7 +124,11 @@ public TypeSortTranslator(final Script script, final IUltimateServiceProvider se
 		}
 	}
 
-	public IBoogieType getType(final Sort sort) {
+	public IBoogieType getType(Sort sort) {
+		if(mCfgScriptFromWorker != null) {
```
maul-esel: Please run auto-formatting on all modified projects.

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/boogie/TypeSortTranslator.java
```diff
@@ -273,4 +280,8 @@ private void cacheSort(final IBoogieType boogieType, final Sort result) {
 		mSort2Type.put(result, boogieType);
 	}
 
+	public void setCfgScriptFromWorker(ManagedScript cfgScriptFromWorker) {
+		mCfgScriptFromWorker = cfgScriptFromWorker;
```
maul-esel: Who calls this? Is it necessary to make the class mutable?

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/cfg/IcfgProgramExecution.java
```diff
@@ -359,5 +361,12 @@ public IBacktranslationValueProvider<L, Term> getBacktranslationValueProvider()
 	public boolean isConcurrent() {
 		return mIsConcurrent;
 	}
+	
+	public void setOriginCfgScript(ManagedScript managedScript) {
+		mScript = managedScript;
```
maul-esel, line 366: Is it necessary to make this class mutable?

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/tracehandling/AutomatonFreeRefinementEngine.java
```diff
@@ -76,6 +76,19 @@ public final class AutomatonFreeRefinementEngine<L extends IAction>
...
+			final IRefinementStrategy<L> strategy, final boolean quickCheck) {
+		mServices = services;
+		mLogger = logger;
+		mStrategy = strategy;
+		mQuickCheck = quickCheck;
+		mRefinementEngineStatistics = new RefinementEngineStatisticsGenerator();
+		mFeasibility = executeStrategy();
+		mRefinementEngineStatistics.finishRefinementEngineRun();
+	}
+
```
maul-esel: This duplicates code from the constructor below. Call `this(...)` instead.

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/tracehandling/AutomatonFreeRefinementEngine.java
```diff
@@ -126,7 +139,7 @@ private LBool executeStrategy() {
 		}
 
 		// trace was feasible, return
-		if (feasibilityResult == LBool.SAT) {
+		if (feasibilityResult == LBool.SAT || mQuickCheck) {
```
maul-esel: This seems wrong -- the log should only be written if the trace is truly feasible.

trunk/source/Library-SmtLibUtils/src/de/uni_freiburg/informatik/ultimate/lib/smtlibutils/ManagedScript.java
```diff
@@ -144,6 +144,10 @@ public LBool assertTerm(final Object lockOwner, final Term term) throws SMTLIBEx
 
 	public LBool checkSat(final Object lockOwner) throws SMTLIBException {
 		assert lockOwner == mLockOwner : generateLockErrorMessage(lockOwner, mLockOwner);
+		if (Thread.interrupted()) {
+			mScript.exit();
+			throw new SMTLIBException("Thread Was Interrupted, crashing before checkSat");
+		}
```
maul-esel: What is the purpose of this code? Why do you check this only here? Would it be possible to use Ultimate's internal timeout mechanism instead?

trunk/source/Library-SmtLibUtils/src/de/uni_freiburg/informatik/ultimate/lib/smtlibutils/ManagedScript.java
```diff
@@ -327,9 +331,11 @@ public TermVariable constructFreshTermVariable(final String name, final Sort sor
 		public TermVariable constructFreshCopy(final TermVariable tv) {
 			String basename = mTv2Basename.get(tv);
 			if (basename == null) {
+				// Variable Names are not synchronized in Parallel CEGAR loop, this warning will be triggered!
```
maul-esel: Is this an issue that could be fixed?

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/InterpolatingTraceCheckCraig.java
```diff
@@ -119,6 +119,7 @@ public InterpolatingTraceCheckCraig(final IPredicate precondition, final IPredic
 			mInterpolantComputationStatus =
 					new InterpolantComputationStatus(ItpErrorStatus.SMT_SOLVER_CANNOT_INTERPOLATE_INPUT, null);
 		}
+		mgdScriptTc.getScript().exit();
```
maul-esel: I see you added this to several trace checks. What is the reason behind this?

trunk/source/RCFGBuilder/src/de/uni_freiburg/informatik/ultimate/plugins/generator/rcfgbuilder/cfg/Call.java
```diff
@@ -86,5 +87,8 @@ public UnmodifiableTransFormula getLocalVarsAssignment() {
 	protected int getNumberOfOpenCalls() {
 		return 1;
 	}
-
+	
+	public void setPayload(IPayload payload) {
+		mPayload = payload;
+	}
```
maul-esel, line 93: Is it necessary to make the payload mutable?

trunk/source/RCFGBuilder/src/de/uni_freiburg/informatik/ultimate/plugins/generator/rcfgbuilder/cfg/Call.java
```diff
@@ -52,7 +53,7 @@ public class Call extends CodeBlock implements IIcfgCallTransition<IcfgLocation>
 	protected CallStatement mCallStatement;
 	protected String mPrettyPrintedStatements;
 
-	Call(final int serialNumber, final BoogieIcfgLocation source, final BoogieIcfgLocation target,
+	public Call(final int serialNumber, final BoogieIcfgLocation source, final BoogieIcfgLocation target,
```
maul-esel, line 56: Why is it necessary to make these constructors public?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/preferences/TAPreferences.java
```diff
@@ -96,6 +96,16 @@ public final class TAPreferences {
...
 
+	// Parallel Trace Abstraction Settings
+	private final int mThreadLimit;
+	private final boolean mParallelCegarLoop;
+	public boolean parallelSearchSrategy;
+	public boolean considerOnlyActiveCounterexamplesInIsEmptyParallel;
+	public boolean minimizeAbstractionPerWorker;
+	public boolean mUseQuickCheckWorker;
+	public int mQuickCheckLoopBound;
+	public int mSearchLoopBound;
```
maul-esel: make these final, private, and follow the naming convention.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/preferences/TraceAbstractionPreferenceInitializer.java
```diff
@@ -512,6 +512,29 @@ public enum CoinflipMode {
 	private static final String DESC_ASSERT_CODEBLOCKS_HEURISTIC_SCORE_THRESHOLD =
 			"If Assert CodeBlocks is set to SMT_FEATURE_HEURISTIC and partitioning strategy is THRESHOLD, two partitions are created, one partition contains all terms >= threshold  and one all terms < threshold";
 
+	// Parallel Trace Abstraction
+	// ========================================================================
+	public static final String LABEL_PARALLEL_CEGAR_LOOP = "Use CEGAR loop for Parallel Trace Abstraction";
+	public static final boolean DEF_PARALLEL_CEGAR_LOOP = false;
```
maul-esel: minor: the `DEF_` constants are typically private

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/preferences/TraceAbstractionPreferenceInitializer.java
```diff
@@ -724,7 +747,7 @@ private static UltimatePreferenceItemContainer getConcurrencySettings() {
 				getPORSettings(), getPetriLbeSettings());
 	}
 
-	private static UltimatePreferenceItemContainer getPORSettings() {
+	public static UltimatePreferenceItemContainer getPORSettings() {
```
maul-esel: why do you need to make this public?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/StrategyFactory.java
```diff
@@ -109,6 +110,22 @@ public class StrategyFactory<L extends IIcfgTransition<?>> {
...
+		mTaPrefs = taPrefsForInterpolantConsolidation;
+		mPrefs = prefs;
+		mInitialIcfg = initialIcfg;
+		mCfgSmtToolkit = initialIcfg.getCfgSmtToolkit();
+		mPredicateFactory = predicateFactory;
+		mPredicateFactoryInterpolAut = predicateFactoryInterpolAut;
+		mPathProgramCache = programCache;
+		mTransitionClazz = transitionClazz;
+	}
+
```
maul-esel, line 118: please avoid the code duplication between constructors

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/StrategyFactory.java
```diff
@@ -154,18 +172,34 @@ public ITARefinementStrategy<L> constructStrategy(final IUltimateServiceProvider
...
+	 * Construct Strategy with given type
 	 *
+	 * @param services
 	 * @param counterexample
-	 *            A trace that will be checked for feasibility and for which, if it is infeasible, a refinement result
-	 *            will be constructed.
-	 *
-	 *            Optionally, accompanied by the sequence of control configurations visited by the trace in the program
-	 *            that is being verified. This sequence is used to judge the quality of proofs ("perfect") and for
-	 *            assert order modulation.
```
maul-esel, line 165: Please do not remove this documentation.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/tracehandling/TraceAbstractionRefinementEngine.java
```diff
@@ -74,6 +75,19 @@ public final class TraceAbstractionRefinementEngine<L extends IIcfgTransition<?>
...
+		mLogger = logger;
+		mStrategy = strategy;
+		mQuickCheck = quickCheck;
+		final AutomatonFreeRefinementEngine<L> afEngine =
+				new AutomatonFreeRefinementEngine<>(services, logger, strategy, quickCheck);
+		mAfeResult = afEngine.getResult();
+		mAfeStatistics = afEngine.getRefinementEngineStatistics();
+		generateProof();
+	}
+
```
maul-esel: Please avoid code duplication between the constructors

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
```diff
@@ -139,7 +142,7 @@ public abstract class AbstractCegarLoop<L extends IIcfgTransition<?>, A extends
 	/**
 	 * Current Iteration of this CEGAR loop.
 	 */
-	private int mIteration;
+	protected int mIteration;
```
maul-esel, line 142: Is it necessary to make this protected? AFAIR there is a getter method anyway.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
```diff
@@ -140,23 +140,51 @@
...
+
+		public String getLongString() {
+			return mLongString;
+		}
+
+		public String getShortString() {
+			return mShortString;
+		}
+
+	}
```
maul-esel: I think these additions are the result of merge issues.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
```diff
@@ -599,20 +624,17 @@ private NondeterministicInterpolantAutomaton<L> constructInterpolantAutomatonFor
...
-					final NestedWordAutomaton<L, IPredicate> inputInterpolantAutomaton,
-					final IPredicateUnifier predicateUnifier, final IHoareTripleChecker htc,
-					final InterpolantAutomatonEnhancement enhanceMode) {
-		final boolean conservativeSuccessorCandidateSelection =
-				enhanceMode == InterpolantAutomatonEnhancement.PREDICATE_ABSTRACTION_CONSERVATIVE;
+	private DeterministicInterpolantAutomaton<L> constructInterpolantAutomatonForOnDemandEnhancementPredicateAbstraction(
+			final NestedWordAutomaton<L, IPredicate> inputInterpolantAutomaton,
+			final IPredicateUnifier predicateUnifier, final IHoareTripleChecker htc,
+			final InterpolantAutomatonEnhancement enhanceMode) {
+		final boolean conservativeSuccessorCandidateSelection = enhanceMode == InterpolantAutomatonEnhancement.PREDICATE_ABSTRACTION_CONSERVATIVE;
```
maul-esel: There are many formatting-only changes here, making it hard to review. Please ensure you use the official formatting settings.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/CegarLoopFactory.java
```diff
@@ -221,16 +221,25 @@ private NwaCegarLoop<L> createFiniteAutomataCegarLoop(final IUltimateServiceProv
...
+					stateFactoryForRefinement);
+		case NONE:
+			if (mPrefs.isParallelCegarLoop()) {
+				return new ParallelNwaCegarLoop<>(name, abstraction, root, csToolkit, predicateFactory, mPrefs, errorLocs,
+						proofProducer, services, mTransitionClazz, stateFactoryForRefinement);
+			}
+			return new NwaCegarLoop<>(name, abstraction, root, csToolkit, predicateFactory, mPrefs, errorLocs,
+					proofProducer, services, mTransitionClazz, stateFactoryForRefinement);
+		default:
+			throw new AssertionError("Unknown Setting: " + mPrefs.getFloydHoareAutomataReuse());
```
maul-esel: Please do not add a `default` case here.

The previous version of the switch was exhaustive, and the exhaustiveness was checked by the Java compiler. This ensures that if in the future, new enum values are added, we get a compiler error and see immediately which code needs to be adapted. The `default` case undermines this.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/CegarNwaWorkerThread.java
```diff
@@ -0,0 +1,569 @@
+package de.uni_freiburg.informatik.ultimate.plugins.generator.traceabstraction;
```
maul-esel, line 28: Remember to add license headers to all new files.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/NwaCegarLoop.java
```diff
@@ -159,13 +159,19 @@ public NwaCegarLoop(final DebugIdentifier name, final INestedWordAutomaton<L, IP
...
 		mScoringMethod = taPrefs.getHeuristicEmptinessCheckScoringMethod();
 		mAStarHeuristic = taPrefs.getHeuristicEmptinessCheckAStarHeuristic();
 		mAStarRandomHeuristicSeed = taPrefs.getHeuristicEmptinessCheckAStarHeuristicRandomSeed();
 	}
 
+	protected NestedRun<L, IPredicate> runWithModifiedGoalSet(final INestedWordAutomaton<L, IPredicate> abstraction,
+			final Set<IPredicate> possibleEndPoints) throws AutomataOperationCanceledException {
+		return new IsEmpty<>(new AutomataLibraryServices(mServices), abstraction, abstraction.getInitialStates(),
+				Collections.emptySet(), possibleEndPoints).getNestedRun();
+
```
maul-esel: Unless I missed a call, this method seems not to be called within this class. If only your subclass needs it, why not declare it there?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/ParallelNwaCegarLoop.java
```diff
@@ -0,0 +1,894 @@
...
+			e.printStackTrace();
+		}
+	}
+
+	public ManagedScript getManagedScript() {
+		return mCsToolkit.getManagedScript();
+	}
+}
+
+final class WorkerThreadResult<L extends IIcfgTransition<?>, A extends IAutomaton<L, IPredicate>> {
```
maul-esel: Please do not declare multiple classes in the same file (except for nested classes).

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/PathProgramCache.java
```diff
@@ -48,8 +48,8 @@
 public class PathProgramCache<LETTER> {
 
 	private final ILogger mLogger;
-	private final Map<Set<LETTER>, Integer> mKnownPathPrograms;
-	private final List<Integer> mTraceHashes;
+	private  Map<Set<LETTER>, Integer> mKnownPathPrograms;
+	private  List<Integer> mTraceHashes;
```
maul-esel: Why not keep these as `final`?

trunk/source/RCFGBuilder/src/de/uni_freiburg/informatik/ultimate/plugins/generator/rcfgbuilder/RCFGBacktranslator.java
```diff
@@ -242,11 +242,13 @@ private void addCodeBlock(final IIcfgTransition<IcfgLocation> cb, final IRelevan
 
 	@Override
 	public IProgramExecution<BoogieASTNode, Expression>
-
 			translateProgramExecution(final IProgramExecution<IIcfgTransition<IcfgLocation>, Term> programExecution) {
 		if (!(programExecution instanceof final IcfgProgramExecution<IIcfgTransition<IcfgLocation>> rcfgProgramExecution)) {
 			throw new IllegalArgumentException();
 		}
+		
+		mTerm2Expression.getTypeSortTranslator().setCfgScriptFromWorker(rcfgProgramExecution.getOriginCfgScript());
```
maul-esel: Why is this call necessary here?

maul-esel (commented):
I left a number of comments. I focused mainly on changes to existing code (I do not have enough overview to review the newly-added parallel CEGAR itself).

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/TransferBetweenMainAndWorker.java
```diff
@@ -0,0 +1,547 @@
...
+							result.addState(automaton.isInitial(hier), automaton.isFinal(hier), hier);
+						}
+						final LETTER transferredLetter = transferEdge(returnTransition.getLetter());
+						returnAlphabet.add(transferredLetter);
+						result.addReturnTransition(potentialhier, hier, transferredLetter, succesor);
+						dequeue.add(succesor);
+					}
+				}
+			}
+			// TODO transfer summaries, so far they are ignored because i dont know how to detect a summary in the input
```
MaxBarth95: WIP

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmpty.java
maul-esel: Ok, now the extra parameter _is_ used. Nevertheless, please avoid duplicating all the other constructor code. I.e., just make the old constructor call this constructor with `initializeOnly = false`.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmpty.java
```diff
@@ -345,6 +382,7 @@ private DoubleDecker<STATE> dequeue() {
 		// If available, take a state pair that has been discovered by taking an internal or a return transition or a
 		// summary. If not, take a state pair that has been discovered by taking a call transition.
 		case DFS -> dequeueGivenQueues(mQueue, mQueueCall);
+		default -> throw new IllegalArgumentException("Unexpected value: " + mStrategy);
```
maul-esel: Please avoid adding a `default` case to this previously-exhaustive switch.

Ideally, `IsEmpty` should not need special handling for its subclass `IsEmptyParallel`. Do you really need the additional enum value? (it's also a bit strange since it couples a search mode (BFS) with the fact that the search is parallelized -- can they be split?)

If there is no way to avoid having the additional enum value, use `case PARALLEL ->` here.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyHeuristic.java
```diff
@@ -116,7 +118,25 @@ public IsEmptyHeuristic(final AutomataLibraryServices services,
 			final INwaOutgoingLetterAndTransitionProvider<LETTER, STATE> operand,
 			final IHeuristic<STATE, LETTER> heuristic) throws AutomataOperationCanceledException {
 		this(services, operand, CoreUtil.constructHashSet(operand.getInitialStates()), a -> false, operand::isFinal,
-				heuristic);
+				heuristic, new ArrayList<>());
+	}
+
+	/**
+	 * Default constructor. Here we search a run from the initial states of the automaton to the final states of the
+	 * automaton and use the zero heuristic.
```
maul-esel: This documentation seems incorrect (the caller actually passes a heuristic) and does not mention the meaning of the `wayPoints` parameter.

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyHeuristic.java
```diff
@@ -128,15 +148,15 @@ public IsEmptyHeuristic(final AutomataLibraryServices services, final INestedWor
 			final Set<STATE> startStates, final Predicate<STATE> funIsForbiddenState,
 			final Predicate<STATE> funIsGoalState, final IHeuristic<STATE, LETTER> heuristic)
 			throws AutomataOperationCanceledException {
-		this(services, (INwaOutgoingLetterAndTransitionProvider<LETTER, STATE>) operand, startStates,
-				funIsForbiddenState, funIsGoalState, heuristic);
+		this(services, operand, startStates, funIsForbiddenState, funIsGoalState, heuristic, new ArrayList<>());
```
maul-esel: minor: use `List.of()` or `Collections.emptyList()` to make it clear the list is unmodifiable

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyParallel.java
```diff
@@ -0,0 +1,711 @@
...
+import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.Pair;
+
+/**
+ * Get an accepting run of a nested word automaton. Based on IsEmpty but adapted for Parallel CEGAR Loop.
+ *
+ * The idea is to have a preprocessing that finds a path (prefix) to an arbitrary state not taken by any other
+ * counterexample. Then we call BFS in this arbitrary state to find a path (suffix) to a goal state. At the end both
+ * paths are merged.
+ *
+ * This class uses recursion. It recursively explores the successor of a state. TODO: non recursive
```
maul-esel, line 73: Are you planning to address this TODO?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyParallel.java
```diff
@@ -0,0 +1,711 @@
...
+	 * Constructor for parallel search strategy. Gets as additional argument the list of all counterexamples currently
+	 * investigated. Tries to find a new counterexample as much different as possible from the once considered.
+	 *
+	 * @param services
+	 *            Ultimate services
+	 * @param operand
+	 *            input NWA
+	 * @param strategy
+	 *            search strategy
+	 * @see #IsEmptyParallel(AutomataLibraryServices, INwaOutgoingLetterAndTransitionProvider)
```
maul-esel: Does the constructor you are referencing here actually exist?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyParallel.java
maul-esel: no?

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/boogie/TypeSortTranslator.java
maul-esel: Transferring between scripts does not seem like a primary task for `TypeSortTranslator`. Could `RCFGBacktranslator` not do the transfer itself, before calling `getType()` ?

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/cfg/IcfgProgramExecution.java
maul-esel: If `mScript` is used "on creation", as your comment above states, could you not set this in the constructor? (and make it a final field?)

Or, alternatively, pass the script alongside (but not as part of) the execution?

trunk/source/Library-SmtLibUtils/src/de/uni_freiburg/informatik/ultimate/lib/smtlibutils/ManagedScript.java
maul-esel: ping

trunk/source/Library-SmtLibUtils/src/de/uni_freiburg/informatik/ultimate/lib/smtlibutils/ManagedScript.java
maul-esel: ping

maul-esel: Often such warnings lead to bugs later on, or make debugging extremely difficult (because there are distinct variables that have the same name)

trunk/source/Library-TraceCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/tracecheckerutils/singletracecheck/InterpolatingTraceCheckCraig.java
maul-esel: ping

trunk/source/RCFGBuilder/src/de/uni_freiburg/informatik/ultimate/plugins/generator/rcfgbuilder/cfg/Call.java
maul-esel: ping

trunk/source/RCFGBuilder/src/de/uni_freiburg/informatik/ultimate/plugins/generator/rcfgbuilder/cfg/Call.java
maul-esel: ping

trunk/source/RCFGBuilder/src/de/uni_freiburg/informatik/ultimate/plugins/generator/rcfgbuilder/RCFGBacktranslator.java
maul-esel: (see my comment in `TypeSortTranslator`)

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/preferences/TraceAbstractionPreferenceInitializer.java
```diff
@@ -512,6 +512,24 @@ public enum CoinflipMode {
...
+
+	public static final String LABEL_THREADLIMIT = "Threadlimit for Parallel CEGAR";
+	private static final Integer DEF_THREADLIMIT = 1;
+
+	public static final String LABEL_MINIMIZE_ABSTRACTION_PER_WORKER =
+			"Minimize Abstraction every time a worker is done";
+	private static final boolean DEF_MINIMIZE_ABSTRACTION_PER_WORKER = true;
+	// Parallel CEGAR counterexample search stragies
+	// ========================================================================
+	public static final String LABEL_PARALLELSEARCH_ACTIVE_CEX_ONLY = "Consider only active in Search Strategy";
```
maul-esel, line 545: Could you add a more detailed description for this setting?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/preferences/TraceAbstractionPreferenceInitializer.java
```diff
@@ -512,6 +512,24 @@ public enum CoinflipMode {
...
+	private static final boolean DEF_PARALLELSEARCH_ACTIVE_CEX_ONLY = false;
+	public static final String LABEL_SEARCH_LOOP_BOUND = "search loop bound";
```
maul-esel, line 549: Could you add a more detailed description for this setting?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
maul-esel: ping

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
```diff
@@ -156,6 +156,8 @@ public abstract class BasicCegarLoop<L extends IIcfgTransition<?>, A extends IAu
 
 	protected boolean mFallbackToFpIfInterprocedural = false;
 	protected IRefinementEngineResult<L, NestedWordAutomaton<L, IPredicate>> mRefinementResult;
+	protected InterpolationTechnique mInterpolationTechnique;
+	protected Class<L> mTransitionClazz;
```
maul-esel: These two fields seem unused within this class. If they are only useful for a subclass, could you not declare them in the subclass directly?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/CegarLoopFactory.java
```diff
@@ -221,16 +221,24 @@ private NwaCegarLoop<L> createFiniteAutomataCegarLoop(final IUltimateServiceProv
...
+					stateFactoryForRefinement);
+		case NONE:
+			if (mPrefs.isParallelCegarLoop()) {
+				return new ParallelNwaCegarLoop<>(name, abstraction, root, csToolkit, predicateFactory, mPrefs,
+						errorLocs, proofProducer, services, mTransitionClazz, stateFactoryForRefinement);
+			}
+			return new NwaCegarLoop<>(name, abstraction, root, csToolkit, predicateFactory, mPrefs, errorLocs,
+					proofProducer, services, mTransitionClazz, stateFactoryForRefinement);
+		}
+		throw new AssertionError("Unknown Setting: " + mPrefs.getFloydHoareAutomataReuse());
```
maul-esel: As for other switches, please rewrite this such that it is an exhaustive switch again.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/CegarNwaWorkerThread.java
```diff
@@ -0,0 +1,578 @@
...
+
+		final Thread.UncaughtExceptionHandler exhandler = (th, ex) -> {
+			// TODO seems to work not sure if it is usefull
+			mThreadResult = new WorkerThreadResult<>(null, null, null, false, null, false, AutomatonType.ERROR, null,
+					mCounterexample, null, true);
+			try {
+				mBlockingQueueForResults.put(mThreadResult);
+			} catch (final InterruptedException e) {
+				// TODO Auto-generated catch block
+				e.printStackTrace();
```
maul-esel: Please replace this auto-generated block.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/CegarNwaWorkerThread.java
```diff
@@ -0,0 +1,578 @@
...
+	}
+
+	private List<IcfgLocation> getIcfgLocationsFromRun(final IRun<L, ?> run) {
+		return run.getStateSequence().stream().map(p -> ((ISLPredicate) p).getProgramPoint())
+				.collect(Collectors.toList());
+	}
+
+	/*
+	 * The worker creates for each counterexample a new strategy module. We can switch between different
+	 * RefinementStrateies here. For example we can use @RefinementStrategy.ACCELERATED_TRACE_CHECK if we have seen the
```
maul-esel: nitpick: the javadoc syntax is `{@link RefinementStrategy.ACCELERATED_TRACE_CHECK}`, not `@RefinementStrategy.ACCELERATED_TRACE_CHECK`

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/PathProgramCache.java
```diff
@@ -53,8 +53,8 @@ public class PathProgramCache<LETTER> {
 
 	public PathProgramCache(final ILogger logger) {
 		mLogger = logger;
-		mKnownPathPrograms = new HashMap<>();
-		mTraceHashes = new ArrayList<>();
+		mKnownPathPrograms = new ConcurrentHashMap<>();
+		mTraceHashes = Collections.synchronizedList(new ArrayList<>());
```
maul-esel, line 58: This seems intended to make this class thread-safe, which it wasn't intended to be by the original author.

I don't have a particular objection to this change (though it might have a bit of a performance impact, it should not be critical).

But it leads me to wonder which other classes are out there with similar problems, that will only be detected by hard-to-reproduce concurrency bugs. Do you have somewhere a documentation of which mutable objects are shared between threads, and how thread safety is ensured for them?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/TransferBetweenMainAndWorker.java
```diff
@@ -0,0 +1,547 @@
...
+		case WORKER2MAIN: {
+			return transferTransformula(inTF, mWorker2main, mMainScript);
+		}
+		default: {
+			throw new AssertionError("Unexpected transferrer mode: " + mMode);
+		}
+		}
+	}
+
+	private UnmodifiableTransFormula transferTransformula(final UnmodifiableTransFormula inTF,
```
maul-esel, line 216: We already have methods `TransferrerWithVariableCache::transferTransFormula` resp. `TransFormulaBuilder::transferTransformula`. Is your implementation doing something fundamentally different, or could you reuse those?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/TransferBetweenMainAndWorker.java
```diff
@@ -0,0 +1,547 @@
...
+		return result;
+
+	}
+
+	public CfgSmtToolkit constructWorkerCfgSmtToolkit() {
+		final HashRelation<String, IProgramNonOldVar> proc2globals =
+				constructNewProc2Globals(mMainCsToolkit.getModifiableGlobalsTable().getProcToGlobals(), mVarTransfer);
+		final ModifiableGlobalsTable modifiableGlobalsTable = new ModifiableGlobalsTable(proc2globals);
+		final IIcfgSymbolTable symbolTable =
+				constructNewSymbolTable(mMainCsToolkit.getSymbolTable(), mMainCsToolkit.getProcedures(), mVarTransfer);
```
maul-esel, line 398: `TransferrerWithVariableCache::transferSymbolTable` could help here.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/TransferBetweenMainAndWorker.java
```diff
@@ -0,0 +1,547 @@
...
+				final IProgramNonOldVar newVar = variableTranslation.getOrConstruct(old);
+				result.addPair(entry.getKey(), newVar);
+			}
+		}
+		return result;
+	}
+
+}
+
+class VariableTransferrer {
```
maul-esel: Please don't add multiple top-level classes in the same file.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/WorkerThreadResult.java
```diff
@@ -0,0 +1,99 @@
+package de.uni_freiburg.informatik.ultimate.plugins.generator.traceabstraction;
```
maul-esel, line 28: Please add a license header

maul-esel (commented):
Thanks for continuing the work on this 👍 

I've gone through the changes again and left some comments. I very much appreciate the much more decoupled approach compared to the first implementation 👍

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/nestedword/operations/IsEmptyParallel.java
MaxBarth95: Not in the near future

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/boogie/TypeSortTranslator.java
MaxBarth95: RCFGBacktranslator is definetly the wrong place to do this. It doesnt even have a script.
I moved it now to Term2Expression, even tho i found TypeSortTranslator very fitting.

PR #735, opened 2025-07-11
Show error if entry function does not exist

Heizmann:


schuessf:
Thanks for this work! I think this setting is a bit odd and it is good to rethink it.
However, I see a potential issue with the proposed change. Previously, if the entry function specified in the settings (usually `main`) was not defined, we would fall back to library mode. With your change, however, we crash instead. I think a few of our regression tests expect this fallback, as they don't define a main-function, and they therefore the tests would fail.
Also, I'm wondering whether we actually ever need to specify an entry function other than `main`. If this is not the case, we could consider removing this setting instead and show a warning if we fall back to library mode.

bahnwaerter (approved):

Heizmann:
>  therefore the tests would fail. 

Shortcomings in our test infrastructure should not be an argument for a confusing setting. But I agree that we should e.g., put these benchmarks in different folders.

> Also, I'm wondering whether we actually ever need to specify an entry function other than `main`. 

Yes, this was sometimes helpful. Especially if we want to understand a complex program better, but want to keep error-prone modifications of the input low.

Heizmann:
Results of discussion on Thu 2025-07-17:
We postpone the decision.
Questions that we had:
* What is the library mode: that every function is an entry or that global variables are initialized deterministically? Or both?
* Do we want a separate setting for both? Are all combinations relevant in practice?

PR #739, opened 2025-07-25
Add packaging support for Docker images

bahnwaerter:
The Docker packaging support allows to build Docker images for all Ultimate tools. The images are based on Alpine Linux to ensure a minimal footprint. The build is based on a multi-stage process where temporary build artifacts are discarded and not included in the final Docker images. Each Ultimate tool now has its own dedicated Docker image. Additionally, the WebBackend can be automatically configured through docker-compose environment files to simplify the configuration and deployment of the Web service.


bahnwaerter:
Currently, the Ultimate tools are built with individual build calls, and the appropriate configurations (toolchain and settings files) are still missing in the respective Docker images.

It would be advantageous if the complete build of the Docker images would be based on the `makeFresh.sh` script. However, this also includes the `makeZip.sh` functionality, which is not necessarily required here. Should we split the functionality of the two scripts with regard to the Docker packaging?

A new separation of functionality could look like this:
- `makeBuild.sh`: Responsible for the Maven build and to collect the correct configurations for each tool
- `makeZip.sh`: Compresses the corresponding built artifacts
- `makeFresh.sh`: Calls `makeBuild.sh` and then `makeZip.sh`

The separation introduces the `makeBuild.sh` script that could be used to build the Docker images.

danieldietsch:
That's a very good idea. I also like your Dockerfile, in particular the use of dockerize!

packaging/docker/Dockerfile
```diff
@@ -44,6 +46,40 @@ RUN mkdir -p "${DIR_BUILD}" && \
...
+                       "ruby-dev"
+
+# build the Ultimate web frontend
+WORKDIR "${REPO_FRONTEND_ROOT}"
+
+# patch the Jekyll installation to work under Ruby 3.4+
+RUN echo -e "gem 'csv'\n" \
+            "gem 'base64'\n" \
+            "gem 'bigdecimal'\n" \
+            "gem 'logger'" >> Gemfile
```
bahnwaerter: As soon as the reported issue #740 is fixed, this patch can be removed.

packaging/docker/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+
+services:
+  ultimate-webbackend:
+    build:
+      dockerfile: "Dockerfile"
+      target: "ultimate-webbackend"
+    image: "ultimate-webbackend"
+    env_file: "ultimate-webbackend.env"
+    ports:
+      - "8080:8080"
```
bahnwaerter: The port of the backend should not be hard-coded here.
Instead, the value from the Docker Compose environment file should be used.

packaging/docker/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    restart: "unless-stopped"
+
+  ultimate-webfrontend:
+    build:
+      dockerfile: "Dockerfile"
+      target: "ultimate-webfrontend"
+    image: "ultimate-webfrontend"
+    env_file: "ultimate-webfrontend.env"
+    ports:
+      - "80:80"
```
bahnwaerter: The port of the frontend should not be hard-coded here.
Instead, the value from the Docker Compose environment file should be used.

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,388 @@
...
+
+ENV PATH "${DIR_OUTPUT}/CLI-E4/linux/gtk/x86_64:${PATH}"
+RUN python3 scripts/build.py
```
bahnwaerter: The build cannot be configured by passing appropriate values, such as setting an arbitrary `baseurl` for Jekyll.
The used Python scripts must be adapted accordingly to make the Jekyll build configurable.

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,388 @@
...
+                       "dbus-x11" \
+                       "ttf-freefont" \
+                       "xauth"
+
+# install already built Ultimate Debug UI product
+COPY --chown="${USER_NAME}:${USER_GROUP}" --from=build "${DIR_OUTPUT}/Debug-E4/linux/gtk/x86_64" "${USER_HOME}"
+RUN ln -s "${USER_HOME}/UltimateDebug" "/usr/bin/UltimateDebug"
+
+# run the Ultimate Debug UI product as non-root user
+#USER "${USER_NAME}"
```
bahnwaerter: Try running the debug UI in a foreign X session as a regular user within the container instead of using the root user.

packaging/docker/Dockerfile
maul-esel: There is a `--production` parameter that, among other things, sets the `baseurl`. But if you need more flexibility, then the script can be made more flexible, yes.

bahnwaerter: I've already discovered the production parameter. This paraemter sets the `baseurl` to the fixed value `/`. Therefore, we should expand the script slightly to allow any `baseurl`  for the Docker image build.

packaging/docker/Dockerfile
bahnwaerter: The issue reported in #740 is fixed and the patch can be removed!

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,516 @@
...
+                       "maven" \
+                       "openjdk21"
+
+# clone fresh Ultimate Git repository
+RUN mkdir -p "${DIR_BUILD}" && \
+    git clone --depth=1 --branch="${REPO_BRANCH}" "${REPO_URL}" "${DIR_ULTIMATE}"
+
+# build all Ultimate products
+WORKDIR "${DIR_ULTIMATE}"
+RUN cd "${DIR_OUTPUT_BUILD}" && bash makeBuild.sh -m "${ULTIMATE_MEMORY_HEAP_MAX}" -s "${ULTIMATE_MEMORY_STACK_MAX}" -p linux
```
danieldietsch: Would it make sense to also support the other platforms? In particular Mac might be interesting to some, but I guess we still don't have anybody who could test that. 
Instead, can you add a line in the README that says why it is limited to Linux?

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,516 @@
...
+                       "g++" \
+                       "make" \
+                       "libc-dev" \
+                       "libc++-dev" \
+                       "linux-headers" \
+                       "ruby-full" \
+                       "ruby-dev"
+
+# install 'dockerize' for automatic configuration file creation
+ENV DOCKERIZE_VERSION="v0.9.5"
```
danieldietsch, line 156: I would place this ENV together with the others at the beginning

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,516 @@
...
+
+ARG REPO_DIR_BUILD
+ARG REPO_DIR_TOOLS
+ARG DIR_BUILD
+ARG DIR_ULTIMATE
+ARG DIR_OUTPUT_BUILD
+ARG DIR_TOOLS
+ARG USER_NAME
+ARG USER_GROUP
+ARG USER_HOME
```
danieldietsch, line 203: Are all of these args necessary for the packaging stages? I dont quite remember if you need to redeclare them for the dependent images...

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,516 @@
...
+USER "${USER_NAME}"
+
+# define health check to supervise the Ultimate web frontend's availability
+HEALTHCHECK --interval=1m --timeout=10s --start-period=20s \
+    CMD curl -f "${ULTIMATE_FRONTEND_PROTOCOL}://${ULTIMATE_FRONTEND_HOST}:${ULTIMATE_FRONTEND_PORT}${ULTIMATE_FRONTEND_ROUTE}/" || exit 1
+
+# default entry point to serve the Ultimate web frontend
+ENTRYPOINT dockerize -template "config.js.tpl:js/webinterface/config.js" \
+                     bundle exec jekyll serve --host "${ULTIMATE_FRONTEND_HOST}" \
+                                              --port "${ULTIMATE_FRONTEND_PORT}"
```
danieldietsch, line 188: Hm, why not build the frontend with jekyll and serve it with a "real" webserver? I was at least under the impression that ``jekyll serve`` is only used for debugging/designing a website, not really for production use.

packaging/docker/Dockerfile
```diff
@@ -0,0 +1,516 @@
...
+ADD "WebBackend.ini.tpl" "${USER_HOME}/WebBackend.ini.tpl"
+RUN rm -f "${USER_HOME}/WebBackend.ini"
+
+ENV ULTIMATE_BACKEND_SETTINGS_WHITELIST="${USER_HOME}/settings_whitelist.json"
+ENV ULTIMATE_BACKEND_SETTINGS_FILE="${USER_HOME}/web.config.properties"
+
+# Serving the frontend from the backend is not supported by this Docker image
+ENV ULTIMATE_FRONTEND_SERVE="False"
+ENV ULTIMATE_FRONTEND_PATH="${USER_HOME}"
+ENV ULTIMATE_FRONTEND_ROUTE="/website"
```
danieldietsch, line 503: Again, I would move all ENVs to the front if they are meant to be editable by someone.

packaging/docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+
+  - `ultimate-cli`
+  - `ultimate-debug`
+
+shipped with the Ultimate program analysis framework.
+
+> [!NOTE]
+> Building the Ultimate product images is currently limited to the Docker target platform `linux/amd64` (Linux containers for the 64-bit x86 architecture).
+> However, these images can still be used on a Windows system with [Docker Desktop](https://docs.docker.com/desktop/setup/install/windows-install/) configured with the WSL2 or Hyper-V backend in order to create and run Linux containers on a Windows system.
+
```
danieldietsch, line 34: Thats the place for the reason :)

packaging/docker/README.md
```diff
@@ -0,0 +1,79 @@
...
+If you want to run Ultimate interactively for any verification input, you can spwan a bash in a created and started Ultimate `PRODUCT` container as follows.
+```shell
+docker run -it <PRODUCT>
+<PRODUCT> -tc <TOOLCHAIN> -s <SETTINGS> -i <PROGRAM>
+```
+Calling the Ultimate `PRODUCT` within the container then follows as usual, where a `TOOLCHAIN`, `SETTINGS`, and `PROGRAM` file should be specified for a verification run.
+
+> [!NOTE]
+> The pre-configured products are already provided with the appropriate configuration (toolchain and setting files).
```
danieldietsch, line 52: Confusingly worded: does this mean I dont have to specify them in the run command above?

packaging/docker/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
+version: "2"
```
danieldietsch, line 1: Recent docker-compose files do not have the version anymore

packaging/docker/welcome.sh
```diff
@@ -0,0 +1,34 @@
...
+        echo "You can access the configuration directory via the environment variable 'ULTIMATE_CONFIG_PATH'."
+    else
+        echo "Product-specific toolchain and setting files for Ultimate are not part of this installation."
+    fi
+}
+
+echo "▗▖ ▗▖▗▖ ▗▄▄▄▖▗▄▄▄▖▗▖  ▗▖ ▗▄▖▗▄▄▄▖▗▄▄▄▖"
+echo "▐▌ ▐▌▐▌   █    █  ▐▛▚▞▜▌▐▌ ▐▌ █  ▐▌   "
+echo "▐▌ ▐▌▐▌   █    █  ▐▌  ▐▌▐▛▀▜▌ █  ▐▛▀▀▘"
+echo "▝▚▄▞▘▐▙▄▄▖█  ▗▄█▄▖▐▌  ▐▌▐▌ ▐▌ █  ▐▙▄▄▖"
```
danieldietsch, line 27: 🤣

packaging/docker/welcome.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+ULTIMATE_CONFIG_PATH="/home/ultimate/config"
+
+check_and_print_version() {
+    for cmd in "${@}"; do
+        if command -v "${cmd}" >/dev/null 2>&1; then
+            "${cmd}" --version
+            return
+        fi
```
danieldietsch, line 10: else?

releaseScripts/default/adds/Ultimate.ini
danieldietsch, line 1: Is this not needed anymore?

releaseScripts/default/makeBuild.sh
```diff
@@ -0,0 +1,193 @@
...
+# Default platforms for the build of Ultimate
+PLATFORMS=("linux" "win32")
+# Default maximum heap memory size for Ultimate products
+MEM_HEAP_MAX_SIZE="12G"
+# Default maximum stack memory size for Ultimate products
+MEM_STACK_MAX_SIZE="512M"
+
+_print_help() {
+  printf 'Usage: %s [-m <size>] [-s <size>] [-p all|linux|win32] [-h]\n' "${0}"
+  print_newline
```
danieldietsch, line 25: feels overengineered ;) -- why not `echo`?

releaseScripts/default/makeZip.sh
danieldietsch, line 1: I think this could now become a function in makeFresh -- which could also be somewhat simpler then. In particular, now that packaging is done before and this only names the zip and actually creates it.

trunk/source/BA_SiteRepository/pom.xml
```diff
@@ -54,7 +54,7 @@
 									<attachId>UltimateDeltaDebugger</attachId>
 								</product>
 								<product>
-									<id>UltimateEliminator</id>
+									<id>Eliminator</id>
```
danieldietsch, line 57: Why remove the Ultimate prefix here?

trunk/source/CoreRCP/plugin.xml
```diff
@@ -37,8 +37,8 @@
       </product>
    </extension>
    
-   <extension point="org.eclipse.core.runtime.products" id="UltimateEliminator">
-      <product application="de.uni_freiburg.informatik.ultimate.core.CoreRCP" name="UltimateEliminator">
+   <extension point="org.eclipse.core.runtime.products" id="Eliminator">
```
danieldietsch, line 40: see above

trunk/source/WebsiteStatic/scripts/build.py
```diff
@@ -28,16 +28,24 @@ def parse_args() -> argparse.Namespace:
         parser.add_argument(
             "--skip-settings", action="store_true", help="skip rebuilding settings (useful during debugging)"
         )
+        parser.add_argument(
+            "--baseurl", action="store", type=str, help="set base URL for site (necessary when hosting the site in a sub-directory)"
```
danieldietsch, line 32: default is `website` right?

danieldietsch (commented):
Very cool changes! 

I added some minor comments, and have two things:

#### Use "real" webserver
I would like to avoid using the Jekyll webserver for the frontend. Could you just use nginx as base?

Our current frontend at https://ultimate-pa.org does the following.
```
services:
  frontend:
    image: nginx:latest
    volumes:
      - /mnt/shared/ultimate-pa/config/nginx/conf.d:/etc/nginx/conf.d
      - /mnt/shared/ultimate-pa/config/nginx/nginx.conf:/etc/nginx/nginx.conf
      - /mnt/shared/ultimate-pa/frontend:/www/frontend
      - /mnt/shared/ultimate-pa/config/frontend/config.js:/www/frontend/config/config.j
```
where ``/mnt/shared/ultimate-pa/frontend`` is the static website, ``config.js`` the config you would expect, ``/mnt/shared/ultimate-pa/config/nginx/nginx.conf`` is default, and `/mnt/shared/ultimate-pa/config/nginx/conf.d/default.conf` is default except specifying the frontend folder:
```
server {
    listen       80;
    listen  [::]:80;
    server_name  localhost;

    location / {
        root   /www/frontend;
        index  index.html index.htm;
    }

    error_page   500 502 503 504  /50x.html;
    location = /50x.html {
        root   /usr/share/nginx/html;
    }

    error_page  404 /404.html;
    location /404.html {
        root /www/frontend;
        internal;
    }
}
```

#### Check windows build
I tried running ``./makeFresh.sh`` as usual, and got at the end:
```
Archive Ultimate Debug UI [linux]
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Not enough arguments supplied -- use arguments in the following order
1. the toolname
2. the launcher name
3. the maximum heap memory size
4. the maximum stack memory size
5. 'linux' or 'win32' for the target platform
```
But everything else worked as expected

packaging/docker/Dockerfile
bahnwaerter: > Would it make sense to also support the other platforms?
> Instead, can you add a line in the README that says why it is limited to Linux?

Sure, that would be a great idea. We could especially add packaging support for the already built MacOS version.
Note that the restriction in the `makeBuild.sh` script is only used for building the Docker images and only affects the final packaging, not the Maven build. I've already added information to the `README.md` that we currently only support Linux containers for Docker. These can also be run on Windows using Docker Desktop with the WSL2 backend. Building native Windows containers for Docker is pretty annoying, but not impossible. Furthermore, I doubt users will want to deal with the licensing issue of Windows containers.

packaging/docker/Dockerfile
bahnwaerter: Yes, the re-declaration of shared build `ARG`s are necessary in a multi-stage Docker file since they only last for the build phase of a single Docker image.

packaging/docker/Dockerfile
bahnwaerter: I created the Docker image of the Web frontend for development purposes. Here, `jekyll serve` offers an automatic rebuild and reload mechanism, which makes it super easy to debug and extend the frontend.
This is obviously not suitable for a productive Docker setup. In this case, you would probably be better off using a decent Web server, like your proposed nginx, and a powerful proxy server, like Traefik.

We could create a second Docker image of the Web frontend for production purposes. What do you think?

packaging/docker/Dockerfile
bahnwaerter: These settings are actually hidden because they are not meant to be editable, but rather to deactivate the Web frontend serving of the Jetty web service. Unfortunately, these variables cannot be static Docker build `ARG`s, as otherwise `dockerize` can no longer use them.

releaseScripts/default/makeBuild.sh
bahnwaerter: That's probably in the nature of a C programmer :wink:

releaseScripts/default/makeZip.sh
bahnwaerter: Good idea! The script has become very small now. It's really not worth a script anymore.

trunk/source/BA_SiteRepository/pom.xml
bahnwaerter: This will finally give us a consistent naming scheme. The actual product name simplifies the folder naming for the packaging process and keeps the packaging logic generic. Until now, this product name has always been out of place. Not even the Delta Debugger beat that :wink:

trunk/source/WebsiteStatic/scripts/build.py
bahnwaerter: Yes. This value is everywhere (Docker file, Docker Compose) the default value.

packaging/docker/Dockerfile
danieldietsch: Ah, that makes sense. Perhaps a comment in the README would be nice so that we remember why it is like it is :)

And of course,  a second variant with nginx would be appreciated, then I would switch our current setup over to that.

releaseScripts/default/makeBuild.sh
danieldietsch: :+1:

trunk/source/BA_SiteRepository/pom.xml
danieldietsch: Fair

releaseScripts/default/semver2.sh
```diff
@@ -0,0 +1,354 @@
...
+# MUST NOT contain leading zeroes
+normalizeZero() {
+  next=$(printf %s "${1#0}")
+  if [ -z "$next" ]; then
+    printf %s "$1"
+  fi
+  printf %s "$next"
+}
+
+semver_compare() {
```
schuessf, line 141: I just tried running `makeFresh.sh`, but it just showed the build splashscreen and exited silently.
I checked and it turned out that `javac --version` produced `javac 24`, which should be >= 21.
After I used `update-alternatives` to change javac to 21, the build worked as expected.
Therefore, I think there is some issue here.

bahnwaerter: You're right. The version comparison should support that.

releaseScripts/default/adds/Ultimate.ini
bahnwaerter: Yes, this configuration file is no longer needed. It is automatically generated from the arguments specified in the `*.product` files. At the end, the package script adjusts the memory sizes for the Java Runtime as needed.

bahnwaerter: One advantage of this approach is that the launcher settings are always adjusted automatically and do not have to be changed after framework migrations.

packaging/docker/README.md
bahnwaerter: You are right. This is very badly written. I will rephrase the note.

> [...] does this mean I dont have to specify them in the run command above?

No, that is not what the note means :rofl: What is meant here is that the images for the containers already ship the product-specific configuration files needed to call the Ultimate `PRODUCT`.

packaging/docker/README.md
bahnwaerter: Is the explanation sufficient for you?

packaging/docker/welcome.sh
bahnwaerter: The `else` branch was intentionally omitted because the script is very general and intended to print version information for each Ultimate product. Therefore, the script checks all known launcher names and prints the version information for the first one that exists in the Docker container.

danieldietsch: Ah, that makes sense -- I was afraid that this leads to silent failures when no command matches

packaging/docker/README.md
danieldietsch: No :) -- that does not contain the reason you gave in the other discussion, and it would be nice to have it here.

PR #743, opened 2025-08-13
Use Ant build task to generate versioning information

### LINKED ISSUE #742: Bug: Eclipse build does not generate Ultimate versioning information
In the Eclipse build process, the Ultimate versioning information is not generated from the Git repository. This prevents the version information from being displayed in the built Ultimate products. The Maven build process is unaffected and generates the version information correctly.

bahnwaerter:
The Ultimate versioning information is now generated in a standalone project from the Git repository. This ensures that the Ultimate version is correctly generated during both an Eclipse build and a Maven build. This change fixes #742.


danieldietsch:
Why is it necessary to have a separate project for version info? Wouldnt it be enough to move the Maven version code to Core?

bahnwaerter:
> Why is it necessary to have a separate project for version info? Wouldnt it be enough to move the Maven version code to Core?

Currently, we have to support two different build processes. So far, we built Eclipse projects without any Maven integration (e.g., via Eclipse M2E). This allows us to use Eclipse's native incremental build system for development purposes, resulting in faster builds. However, this is no longer possible for generating version information. Instead, we have to integrate the Maven `git-commit-id-maven-plugin` into Eclipse. To keep the build as lean as possible, the generation of version information is now outsourced to a separate Eclipse project. Otherwise, we would have to build the entire `CoreRCP` with Maven via Eclipse M2E. This slows down the build because many Maven goals and Tycho plugin phases must be run through, and the Ant build of our XML-based run configurations can no longer be done natively within Eclipse.

danieldietsch:
Ah, and an alternative, i.e., using some Ant process to generate version information in the Eclipse build, is not as nice? I mean, it would be good if we could keep Maven out of the Eclipse build process altogether.

bahnwaerter:
Yeah, I forgot to mention an Ant task as an alternative. This would be possible but requires some shell logic to extract the necessary Git versioning information. Maybe there is already a customized And task that does this automatically?!?

schuessf:
Thanks for your work!
I don't really understand this Ant magic, but I just tried it out 😉
After the Ant build was executed, the correct `version.properties` is generated and therefore the correct version is displayed, also when running it from Eclipse. 👍 
However, for me the build is only executed, when I trigger a clean build (at least for `CoreRCP`) manually. Ideally, the build should be triggered after every commit etc. (I guess if `.git` is modified). Is this somehow possible, or am I missing something here?

danieldietsch:
It would probably also be ok if it would be triggered for every build.

danieldietsch (approved):

bahnwaerter:
> However, for me the build is only executed, when I trigger a clean build (at least for `CoreRCP`) manually. Ideally, the build should be triggered after every commit etc. (I guess if `.git` is modified). Is this somehow possible, or am I missing something here?

Thanks for testing. The version information should be generated during each build. I have not specified any dependent sources, which means the Ant target should be triggered for every build.

bahnwaerter:
The version information should now also be generated correctly in a Git Worktree. Previously, the branch name was incorrectly resolved in such cases.

bahnwaerter:
Note that the Ultimate versioning information will only be regenerated correctly if the Eclipse IDE option `Project` &rarr; `Build Automatically` is disabled. If the option is checked, the Eclipse IDE will automatically compile the Java code whenever a Java file is modified. This does not mean that the Eclipse IDE will completely build and package all Ultimate projects. It basically really just compiles the code. Therefore, it would be better to keep the Eclipse IDE option in our default Eclipse settings disabled and leave the Eclipse IDE option `Window` &rarr; `Preferences` &rarr; `Run/Debug` &rarr; `Launching` &rarr; `[General Options]` &rarr; `Build (if required) before launching` enabled.

trunk/source/CoreRCP/.externalToolBuilders/CoreRCP-Generator.launch
```diff
@@ -0,0 +1,20 @@
...
...uot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/SmtParser&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/SpaceExParser&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/SpaceExParserTest&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/SyntaxChecker&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/TraceAbstraction&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/TraceAbstractionWithAFAs&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/TreeAutomizer&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/UltimateCLI&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/UltimateDeltaDebugger&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/UltimateEliminatorController&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/UltimateRegressionTest&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/UltimateTest&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/WebBackend&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/WebsiteStatic&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/WitnessParser&quot; type=&quot;4&quot;/&gt;&#10;&lt;item path=&quot;/WitnessPrinter&quot; type=&quot;4&quot;/&gt;&#10;&lt;/resources&gt;}"/>
```
maul-esel: Will new projects have to be added here (manually)?

bahnwaerter: Yes, new projects would have to be added here. However, I've found that this doesn't really have any effect. I thought this explicit dependency would help Eclipse if it loses its internal build state. But that doesn't seem to be the case. Only a clean and rebuild of the plugin projects will help. So I'm going to undo this little change.

Issue #440, opened 2019-08-14
Represent ambivalent floats as bitvectors


enkejill:
We want to rewrite the current translation of C floating point types to Boogie to fix various issues, in particular the support of different kinds of ``NaN``s.

Our solution is -- at a very high level -- to wrap all FP operations with our own bitvector encoding, which should be changeable (see also @Heizmann's comment on #228). 

The following parts are needed. 
- [x] Represent floating point types as bitvectors instead of floats (i.e., change the representation in PostProcessor) 
- [x] Create wrapper function similar to:
````
procedure write~intFLOATTYPE4(#value : C_FLOAT, #ptr : $Pointer$, #sizeOfWrittenType : C_INT) returns ();
ensures (exists #valueAsBitvector : bv32 :: #memory_int == old(#memory_int)[#ptr := #valueAsBitvector] && ~fp~FLOAT(#valueAsBitvector[32:31], #valueAsBitvector[31:23], #valueAsBitvector[23:0]) == #value);
modifies #memory_int;
````
- [x] Add wrapper function as procedure in `PostProcessor.postProcess()`
- [x] Convert float function input/output from/to bitvec
- [ ] Adjust rounding mode handling accordingly.
- [x] Custom implementation of copysign
- [ ] Research which other functions that will need a custom implementation according to standard.
   (functions that can return specific NaNs, i.e. sNaN, qNaN or specific +NaN / -NaN)

This feature is related to #228, #268, #381, #398, #412, and #423 . 

It might also be relevant for #233 and #54

**Comment by Heizmann** [comment]
Let us assume we have a floating point operation fp.op that occurs in 
``z := fp.op(x,y)``

An obvious(?) translation would be
```
havoc z;
assume to_fp_unsigned(z) ==  fp.op(to_fp_unsigned(x),to_fp_unsigned((y))
```
Unfortunately, this will in general lead to quantified formulas but I cannot see a better solution.
However, the following could be an optimization.
```
if (fp.isNan(fp.op(to_fp_unsigned(x),to_fp_unsigned((y))) {
    havoc z;
    assume to_fp_unsigned(z) == fp.op(to_fp_unsigned(x),to_fp_unsigned((y))
} else {
    z := fp.to_ubv(fp.op(to_fp_unsigned(x),to_fp_unsigned((y))) )
}
```

danieldietsch:
[...]

Issue #398, opened 2018-11-08
Support the modification of the rounding mode for IEEE 754 floating point numbers


Heizmann:
### Current state
* We support basic float operations and functions by translating them via our [`:builtin` directive](https://github.com/ultimate-pa/ultimate/wiki/Boogie#directives) into SMT functions.
* The [Floating Point theory](http://smtlib.cs.uiowa.edu/theories-FloatingPoint.shtml) of the SMT-LIB standard (and the solvers) supports all rounding modes. For each SMT function the rounding mode is passed as an argument.
* We always pass the default rounding mode and crash if the C program uses e.g., the `fesetround` function which changes the current rounding mode.

### Desired state
* Introduce (in the Boogie program) a global variable that stores the current rounding mode.
* Initialize this variables with the default rounding mode.
* Modify this variable if the rounding mode of the original C program is changed.
* Pass the value of this variable as an argument to each SMT FloatingPoint function (instead of the hardcoded default rounding mode).
* This requires that we add to the Boogie program a rounding mode type and five values. This should be done in a way that we can use our [`:builtin` directive](https://github.com/ultimate-pa/ultimate/wiki/Boogie#directives) for obtaining the the corresponding SMT sorts and constants.

The "current state" of our translation has also some benefits. For C programs that do not change the rounding mode, the verification is less resource-consuming than the "desired state" of our translation.
Hence we would prefer to that the "desired state" is available _in addition_ to the "current state" and we can use a setting to switch between boths.

Heizmann:
This should be discussed with @danieldietsch before it is implemented.

danieldietsch:
I had to remove various parts of the current implementation to restore soundness in time for the SVCOMP release: 
 * 635dfa2a78e594483db942e07fa9a12410e3a48e removed the complete rounding mode implementation. I am sorry that I did not notice this earlier: The rounding mode identifier expression should point to a global variable _in the Boogie program_. Therefore. ``fesetround`` has to be a procedure that changes a global variable in the Boogie program if the argument is correct. All C aspects that use the rounding mode then need to reference the identifier expression  declared in ``BitvectorExpressionTranslation``, which then should point to a global variable.
[...]

PR #750, opened 2025-09-29
Move complete handling of functions to `MathLibraryModel`, support additional functions

schuessf:
PR #714 refactored the way standard functions in C are modeled in Boogie. As a result, separate classes are now used to handle functions from specific C libraries. In particular, the class `MathLibraryModel` is responsible for modeling functions from `math.h`. However, in case of `MathLibraryModel` the actual translation is delegated to [`BitvectorTranslation::constructOtherUnaryFloatOperation`](https://github.com/ultimate-pa/ultimate/blob/a98ac16fc834ddb35deff87f94df70f46d1ad04e/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/BitvectorTranslation.java#L1035) and [`BitvectorTranslation::constructOtherBinaryFloatOperation`](https://github.com/ultimate-pa/ultimate/blob/a98ac16fc834ddb35deff87f94df70f46d1ad04e/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/BitvectorTranslation.java#L1253) (the other implementations of those methods in `IntegerTranslation` simply throw an exception, since these float operations are not supported), while `FloatFunction` serves as some kind of connection (with some hacky string operations to determine the C type). Consequently, whenever support for new math library functions is added, three classes had to be changed (`MathLibraryModel`, `BitvectorTranslation`, `FloatFunction`), which made the process cumbersome.

Therefore, this PR removed all the handling of standard functions from `BitvectorTranslation`, as well as the (rather hacky) class `FloatFunction` and moved the translation of functions from `math.h` entirely to `MathLibraryModel` (similar as done in other library models). To facilitate this, new methods were added to `ExpressionTranslation` to serve as a proper API. In `BitvectorTranslation` these methods implement calls to SMT-defined floating-point theory functions, while in `IntegerTranslation`, they throw exceptions.

Additionally, this PR added support for more functions from `math.h` (with the help of overapproximations):
* `signbit`, `remainder`, `copysign`: They were removed in 042d866 (see [this comment](https://github.com/ultimate-pa/ultimate/issues/398#issuecomment-442757689)), and since then only an exception was thrown. This was due to an unsoundness, since we were not able to handle negative NaNs in SMT, but they can occur in C (with the help of `copysign`). While some efforts were made to handle these cases precisely (see #440), they were left incomplete, so this PR uses overapproximations instead. Specifically, when NaN values appear as arguments to `copysign` or `signbit`, the result is safely overapproximated, which remains precise enough for most of the cases.
* Several numeric functions (e.g., `sin`, `exp`, ...) were added, with an error handling (i.e., ±0, ±∞ and NaN as arguments) according to the C standard. For all other return values, a very coarse overapproximation is used, capturing simple properties - for example `sin(x)` is always between -1 and 1, and `exp(x)` is always at least `x+1`. Despite this coarse overapproximation, the translation should still be precise enough to verify some (very simple) tasks of the [neural-networks](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/tree/main/c/neural-networks) benchmarks.


danieldietsch:
@schuessf @bahnwaerter does this lead to conflicts with the refactored memory model?

schuessf:
> @schuessf @bahnwaerter does this lead to conflicts with the refactored memory model?

I don't think so, this PR basically just changes `MathLibraryModel` and removes two methods from `BitvectorTranslation`, whereas the memory model refactoring mostly changes `MemoryHandler`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/MathLibraryModel.java
```diff
@@ -532,6 +593,467 @@ private Result handleIsLessGreater(final IDispatcher main, final IASTFunctionCal
 		return rtr;
 	}
 
+	private ExpressionResult handleSqrt(final IDispatcher main, final IASTFunctionCallExpression node,
+			final ILocation loc, final String name, final CPrimitive type) {
```
maul-esel, line 600: minor refactoring suggestion: why not let these kinds of methods take a `CPrimitives` and have one `new CPrimitive(...)` call in the method, rather than one in every usage? This will improve readability of `getFunctionModels`.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
```diff
@@ -605,6 +543,40 @@ public abstract Expression checkInRangeInfinitePrecision(ILocation loc, Expressi
...
+
+	public abstract Expression createPlusZero(ILocation loc, CPrimitive type);
+
+	public abstract Expression min(ILocation loc, Expression firstArgument, Expression secondArgument, CPrimitive type);
+
+	public abstract Expression max(ILocation loc, Expression firstArgument, Expression secondArgument, CPrimitive type);
+
+	public abstract Expression remainder(ILocation loc, Expression firstArgument, Expression secondArgument,
+			CPrimitive type);
+
```
maul-esel: Please add some documentation to all the methods above, at least the cppreference URLs / C standard references.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
```diff
@@ -605,6 +543,40 @@ public abstract Expression checkInRangeInfinitePrecision(ILocation loc, Expressi
 	 */
 	public abstract Expression convertInfinitePrecisionExpression(ILocation loc, Expression exp, CPrimitive type);
 
+	public abstract Expression roundToIntegral(ILocation loc, Expression argument, CPrimitive type,
+			Expression roundingMode);
```
maul-esel: `ExpressionTranslation` has many responsibilities (already before, they were just grouped in two methods). I am wondering a bit if it would be possible to separate some of this functionality. E.g. could we separate float support, by giving `ExpressionTranslation` a method `IFloatingPointHandler getFloatingPointHandler()`, and then moving float support to different `IFloatingPointHandler` implementations?

I don't have a good overview of the internal dependencies though, so no idea if this is a viable suggestion.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
```diff
@@ -605,6 +543,40 @@ public abstract Expression checkInRangeInfinitePrecision(ILocation loc, Expressi
 	 */
 	public abstract Expression convertInfinitePrecisionExpression(ILocation loc, Expression exp, CPrimitive type);
 
+	public abstract Expression roundToIntegral(ILocation loc, Expression argument, CPrimitive type,
+			Expression roundingMode);
+
+	public abstract Expression sqrt(ILocation loc, Expression argument, CPrimitive type);
+
+	public abstract Expression abs(ILocation loc, Expression argument, CPrimitive type);
```
maul-esel: For `abs` and many of the other methods here , it is not so clear whether they relate only to floating-point types, or also integral types. Please clarify this (the above refactoring of moving them to an `IFloatingPointHandler` or similar would of course take care of this).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
```diff
@@ -605,6 +543,40 @@ public abstract Expression checkInRangeInfinitePrecision(ILocation loc, Expressi
...
+
+	public abstract Expression createInfinity(ILocation loc, CPrimitive type);
+
+	public abstract Expression createMinusInfinity(ILocation loc, CPrimitive type);
+
+	public abstract Expression createPlusZero(ILocation loc, CPrimitive type);
+
+	public abstract Expression min(ILocation loc, Expression firstArgument, Expression secondArgument, CPrimitive type);
+
+	public abstract Expression max(ILocation loc, Expression firstArgument, Expression secondArgument, CPrimitive type);
```
maul-esel: Is it necessary to have `min` and `max` in `ExpressionTranslation`, or is it possible to just have comparison and then implement the two methods in `MathLibraryModel`?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/IntegerTranslation.java
```diff
@@ -802,4 +783,88 @@ protected Pair<Expression, Expression> constructOverflowCheckForLeftShift(final
 			final ExpressionResult exprResult) {
 		return constructOverflowCheck(loc, resultType, exprResult.getLrValue().getValue());
 	}
+
+	@Override
+	public Expression roundToIntegral(final ILocation loc, final Expression argument, final CPrimitive type,
+			final Expression roundingMode) {
+		throw new UnsupportedOperationException(NOT_IMPLEMENTED);
```
maul-esel: Would it (in the future, not this PR) make sense to implement some of these functions with overapproximations as well? Or do we prefer to fail early if these are used, e.g. because we assume only bitvector mode will have a chance to verify such programs? (CC @danieldietsch ?)

maul-esel (commented):
:+1: Thanks for making this effort to further improve our library handling, and even extending it!

I don't know much about floating-point arithmetic, so I couldn't review those aspects in detail. I have left some comments/questions mostly on the architecture.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
schuessf: I think this is a good idea!
I was a bit unsure what to do with those methods, therefore I also did not add any documentation yet.
As you already pointed out, these methods are only for floats, but this is currently not clear from their name (e.g. `remainder`, `abs`, ...). I was already thinking about giving these functions some prefix, but moving them to another interface seems even better.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
schuessf: We could also implement them in `MathLibraryModel`, but there exist corresponding SMTlib functions (`fp.min` and `fp.max`), which makes it probably easier for the SMT-solver.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
schuessf: I will add documentation when I refactored them again. But just to make clear: These functions are not directly related to the C standard, they just correspond to SMTlib functions from the floating point library (for the `BitvectorTranslation`) that are then used in `MathLibraryModel` to actually model C functions.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/IntegerTranslation.java
danieldietsch: Oh I think in particular for float we could have various different implementations, also with overapproximations. I don't think we want to fail early (or: not unconditionally, it should be an option), but rather always have the option of, e.g., havoc + assume inside type range + overapprox annotation.

schuessf: In principle, it could be helpful, e.g. if a call to such a function does not really influence the correctness of the program or if a violation already occurs before a call to such a function (which seems both not too likely), since we already crash here before running the actual verification.
This PR makes this easier, only a relatively small refactoring should be required to return an `ExpressionResult` that contains the overapproximation instead of an `Expression` for those methods.

maul-esel: >  I don't think we want to fail early (or: not unconditionally, it should be an option)

Maybe another argument for some splitting: `IntegerExpressionTranslation` could select a different `IFloatingPointHandler` implementation based on a setting :)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
maul-esel: :+1: My main point was that we should clarify the contract that these methods must satisfy (as it is not the C standard, apparently it was confusing at least to me).

The SMT-LIB functions seem to be an implementation detail of the `BitvectorTranslation`. Maybe IEEE floating points is the right reference then? And the contract is that they must produce an `Expression(Result)` overapproximating IEEE semantics?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
schuessf: Unfortunately, I don't think separating this functionality is possible. This float handling is closely connected to the `ExpressionTranslation`. In the `BitvectorTranslation`, these methods return calls to boogie procesdures with specific names to `BitvectorTranslation` and declares them if necessary with the help of SMT-functions.

Therefore, I think the best way is to keep this functionality as it is in these methods. But we should definitely rename them and give them some documentation to clarify their functionality. Do you have any good idea for such method names?

maul-esel: How tightly is this coupled? Do other `ExpressionTranslation` methods also call these float operations? (I didn't see any changes to this effect, so I guess not).

I don't think there is any problem in giving an `IFloatingPointHandler` implementation a reference to the containing `BitvectorTranslation` / `IntegerTranslation`.

schuessf: > Do other `ExpressionTranslation` methods also call these float operations?

No, `IntegerTranslation` does not call any of the float operations, it either overapproximates or throws an exception.

> I don't think there is any problem in giving an `IFloatingPointHandler` implementation a reference to the containing `BitvectorTranslation` / `IntegerTranslation`.

The implementations of those methods in `BitvectorTranslation` call methods that are not part of `ExpressionTranslation`, in particular `getBoogieFunctionName` and `declareFloatingPointFunction`, because only the `BitvectorTranslation` declares procedures for floating point operations and their names are also specific to the sizes of the bitvectors.
Of course, I could just implement the interface inside `BitvectorTranslation` to still call `getBoogieFunctionName` and `declareFloatingPointFunction`, but this does not really separate the functionality.

Also, I think moving these methods to an interface, makes the usage quite ugly, `exprTranslation.getFloatingPointHandler().roundToIntegral(...)` instead of `exprTranslation.roundToIntegral(...)` has to be used (`exprTranslation.getFloatingPointHandler()` could be moved to a separate field, but only if it does not throw an exception like for `IntegerTranslation`).

maul-esel: > The implementations of those methods in BitvectorTranslation call methods that are not part of ExpressionTranslation, in particular getBoogieFunctionName and declareFloatingPointFunction, because only the BitvectorTranslation declares procedures for floating point operations and their names are also specific to the sizes of the bitvectors.

If these are at least package-visible methods, the handler could still call them, right?

> Of course, I could just implement the interface inside BitvectorTranslation [...]  but this does not really separate the functionality.

True, that would not really improve the situation.

> `getBoogieFunctionName` and `declareFloatingPointFunction`

At least the latter of these sounds as if it also belongs into a floating-point handler, doesn't it?

> exprTranslation.getFloatingPointHandler() could be moved to a separate field, but only if it does not throw an exception like for IntegerTranslation

I don't understand this part

schuessf: > If these are at least package-visible methods, the handler could still call them, right?

But they only exist in `BitvectorTranslation` and not in `IntegerTranslation` .

I still don't quite understand your idea of `IFloatingPointHandler`.
`ExpressionTranslation` already serves as an abstraction to be used when e.g., translating arithmetic operations or standard functions. For this reason, it has various methods (e.g., `constructBinaryComparisonIntegerExpression`, `extractBits`, `concatBits`) with a specific implementation, already before this refactoring. Why are float operations so different that they deserve an interface as an extra level?

maul-esel: > But they only exist in BitvectorTranslation and not in IntegerTranslation

```java
class BitvectorFloatingPointHandler implements IFloatingPointHandler {
  private final BitvectorTranslation mExprTranslation;
  public BitvectorFloatingPointHandler(BitvectorTranslation exprTranslation) {
    mExprTranslation = exprTranslation;
    mExprTranslation.getBoogieFunctionName("why not like this?");
  }
}
class BitvectorTranslation extends ExpressionTranslation {
  private final IFloatingPointHandler mFloatHandler;
  public BitvectorTranslation() {
    mFloatHandler = new BitvectorFloatingPointHandler(this);
  }
  public IFloatingPointHandler getFloatingPointHandler() {
    return mFloatHandler;
  }
}
```

> Why are float operations so different that they deserve an interface as an extra level?

Mainly because it appears float handling is quite complicated (e.g. we have to directly support many math library functions, which can be handled generically for integers). We should try to not overcomplicate individual classes. This is a neat package of functionality that can be encapsulated separately.

Also, we already discussed that in the future, we may want to have options for different float handling (throw exceptions, vs different overapproximations) with the same ExpressionTranslation (integer). If we encapsulate float handling in a separate type, we only have to change which constructor is called by `IntegerTranslation` (depending on a setting), and we are done.

schuessf: > Mainly because it appears float handling is quite complicated

I am not sure, if floats are really _that_ special in `BitvectorTranslation`, e.g., the method `signExtend` does look a lot similar to the methods I added for floats.

> we have to directly support many math library functions, which can be handled generically for integers

But this is part of `MathLibraryModel` and not `ExpressionTranslation`. And how can these be handled generically for integers?

maul-esel: > But this is part of MathLibraryModel and not ExpressionTranslation

I am referring to the 16 methods added to `ExpressionTranslation` in this PR. As far as I can see, there are no corresponding methods for integer types in `ExpressionTranslation` for any of these methods.

(The fact that they need to be supported here is not new, they were just hidden -- so all of this is not a criticism of your refactoring, just a suggestion to make it even better :smiley: )

> I am not sure, if floats are really that special

Maybe floats are not that special. My point is simply that the float methods seem to form a coherent package of functionality that can be encapsulated separately. If there are similarly sensible packages of integer-related functionality, it might be a good idea to also encapsulate them.

The underlying reason is simply the complexity of this class. If I count correctly, ExpressionTranslation would now have 79 methods, and each subclass must implement at least the 53 abstract methods. That's a lot! (Our sonar creates a warning once a class has more than 35 methods, and I believe that rule is if anything a bit lax.)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/BitvectorFloatingPointHandler.java
```diff
@@ -0,0 +1,126 @@
+package de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.base.expressiontranslation;
```
maul-esel, line 28: don't forget to add license headers to the new files

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/BitvectorFloatingPointHandler.java
```diff
@@ -0,0 +1,126 @@
...
+
+	public BitvectorFloatingPointHandler(final BitvectorTranslation translation) {
+		mTranslation = translation;
+	}
+
+	@Override
+	public Expression roundToIntegral(final ILocation loc, final Expression argument, final CPrimitive type,
+			final Expression roundingMode) {
+		final String smtFunctionName = "fp.roundToIntegral";
+		mTranslation.declareFloatingPointFunction(loc, smtFunctionName, false, true, type, type);
```
maul-esel, line 53: possibility for further refactoring (but we don't need to push everything into one PR): move this method and its other users also to `BitvectorFloatingPointHandler`.

I guess this is also used by some unary / binary operations on floating points. We shouldn't move the general methods for unary / binary operations to `IFloatingPointHandler` of course, but perhaps their implementations can call the necessary methods which reside either in the `IFloatingPointHandler` interface, or just in `BitvectorFloatingPointHandler` (as e.g. the integer translation already does some overapproximation for these, it should not call methods on `UnsupportedFloatingPointHandler`).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/BitvectorTranslation.java
```diff
@@ -1633,4 +1242,8 @@ protected Pair<Expression, Expression> constructOverflowCheckForLeftShift(final
 		return new Pair<>(biggerMinInt, smallerMaxInt);
 	}
 
+	@Override
+	public IFloatingPointHandler getFloatingPointHandler() {
+		return new BitvectorFloatingPointHandler(this);
```
maul-esel: Consider storing the handler in a final field? (Semantically it doesn't seem to be a problem, but it is perhaps a bit unnecessary overhead to recreate new objects each time)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/ExpressionTranslation.java
```diff
@@ -612,4 +550,40 @@ public static Statement modelUnsupportedFeature(final ILocation loc, final Strin
 		return new WhileStatement(loc, ExpressionFactory.createBooleanLiteral(loc, true),
 				new LoopInvariantSpecification[0], new Statement[] { assertFalse });
 	}
+
+	public abstract IFloatingPointHandler getFloatingPointHandler();
+
+	public interface IFloatingPointHandler {
```
maul-esel: move this interface to its own file

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/BitvectorFloatingPointHandler.java
```diff
@@ -0,0 +1,126 @@
+package de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.base.expressiontranslation;
+
+import de.uni_freiburg.informatik.ultimate.boogie.ast.Expression;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.base.expressiontranslation.ExpressionTranslation.IFloatingPointHandler;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.container.c.CPrimitive;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.container.c.CPrimitive.CPrimitives;
+import de.uni_freiburg.informatik.ultimate.core.model.models.ILocation;
+
+public class BitvectorFloatingPointHandler implements IFloatingPointHandler {
```
maul-esel, line 42: add documentation, e.g. along the lines of _Implementation of IFloatingPointHandler that constructs calls to SMT-defined Boogie functions, which refer to built-in functions in the theory of floating-point bit vectors (FPBV)_ or similar

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/IntegerTranslation.java
```diff
@@ -611,30 +611,11 @@ private static Expression constructEquality(final ILocation loc, final int nodeO
 		throw new IllegalArgumentException("operator is neither equals nor not equals");
 	}
 
-	@Override
-	public ExpressionResult createNanOrInfinity(final ILocation loc, final String name) {
-		throw new UnsupportedOperationException("createNanOrInfinity is unsupported in non-bitprecise translation");
-	}
-
 	@Override
 	public Expression getCurrentRoundingMode() {
```
maul-esel, line 616: Another candidate that could potentially be moved to IFloatingPointHandler

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/UnsupportedFloatingPointHandler.java
```diff
@@ -0,0 +1,94 @@
+package de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.base.expressiontranslation;
```
maul-esel, line 28: remember to add a license header

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/UnsupportedFloatingPointHandler.java
```diff
@@ -0,0 +1,94 @@
...
+
+import de.uni_freiburg.informatik.ultimate.boogie.ast.Expression;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.base.expressiontranslation.ExpressionTranslation.IFloatingPointHandler;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.container.c.CPrimitive;
+import de.uni_freiburg.informatik.ultimate.core.model.models.ILocation;
+
+public class UnsupportedFloatingPointHandler implements IFloatingPointHandler {
```
maul-esel, line 40: add a documentation comment

maul-esel (commented):
Just a couple more small comments.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/BitvectorFloatingPointHandler.java
schuessf: Declaring these functions is done through the field `mFunctionDeclarations` and it is quite deeply nested in `BitvectorTranslation`. In particular, there exist also many other methods to declare non-float functions (e.g., `declareBitvectorFunctionForComparisonOperation`). Therefore, this would require quite a bit more of refactoring and will be left for future PRs.

bahnwaerter:
> > @schuessf @bahnwaerter does this lead to conflicts with the refactored memory model?
> 
> I don't think so, this PR basically just changes `MathLibraryModel` and removes two methods from `BitvectorTranslation`, whereas the memory model refactoring mostly changes `MemoryHandler`.

You're right. The changes only affect the memory model, which particularly affects the memory structure and addressing. There may be a small merge conflict, but this should be very easy to resolve. I would suggest that we merge the memory model first (as soon as the last unsoundness issues have been resolved) and then merge this pull request.

trunk/examples/programs/FloatingPoint/regression/c/all/ctrans-float-copysign-2.c
```diff
@@ -20,5 +20,5 @@ int main(void)
 
   x = copysign(NAN, -2.0);
   __VERIFIER_assert(isnan(x));
-  __VERIFIER_assert(signbit(x));
+  //__VERIFIER_assert(signbit(x));
```
maul-esel: Is this commented-out because, due to the overapproximation, the program cannot be proven safe any longer?

If so, please add a short comment to document this (and other commented-out test code below).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/expressiontranslation/UnsupportedFloatingPointHandler.java
```diff
@@ -0,0 +1,126 @@
...
+ */
+
+package de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.base.expressiontranslation;
+
+import de.uni_freiburg.informatik.ultimate.boogie.ast.Expression;
+import de.uni_freiburg.informatik.ultimate.cdt.translation.implementation.container.c.CPrimitive;
+import de.uni_freiburg.informatik.ultimate.core.model.models.ILocation;
+
+/**
+ * Implementation of {@link IFloatingPointHandler} that crashes whenever any method is invoked, which can be used when
```
maul-esel:
```suggestion
 * Implementation of {@link IFloatingPointHandler} that throws {@link UnsupportedOperationException} whenever any method is invoked, which can be used when
```

maul-esel (approved):

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/MathLibraryModel.java
```diff
@@ -532,6 +596,464 @@ private Result handleIsLessGreater(final IDispatcher main, final IASTFunctionCal
...
+		final CPrimitive type = (CPrimitive) argumentResult.getCType();
+		final ExpressionResultBuilder builder = new ExpressionResultBuilder().addAllExceptLrValue(argumentResult);
+		final AuxVarInfo auxVar = mAuxVarInfoBuilder.constructAuxVarInfo(loc, intType, AUXVAR.RETURNED);
+		builder.addAuxVarWithDeclaration(auxVar);
+		final Statement nondet = new HavocStatement(loc, new VariableLHS[] { auxVar.getLhs() });
+		// TODO: Handle negative NaN correctly, only overapproximated until then
+		// signbit(x) := isNegative(x) ? 1 : 0;
+		new OverapproxVariable("sign of NaN", loc).annotate(nondet);
+		final Expression zero = mExpressionTranslation.constructLiteralForIntegerType(loc, intType, BigInteger.ZERO);
+		builder.addStatement(StatementFactory.constructIfStatement(loc, mFloatHandler.isNan(loc, argument, type),
```
schuessf, line 979: @danieldietsch @Heizmann As you have worked with floats in the past, does this handling of `signbit` (incl. the overapproximation) seem reasonable?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/MathLibraryModel.java
```diff
@@ -532,6 +596,464 @@ private Result handleIsLessGreater(final IDispatcher main, final IASTFunctionCal
...
+				ExpressionFactory.or(loc,
+						mExpressionTranslation.constructBinaryComparisonExpression(loc, IASTBinaryExpression.op_equals,
+								auxVar.getExp(), type, first, type),
+						mExpressionTranslation.constructBinaryComparisonExpression(
+								loc, IASTBinaryExpression.op_equals, auxVar.getExp(), type, mExpressionTranslation
+										.constructUnaryExpression(loc, IASTUnaryExpression.op_minus, first, type),
+								type)));
+		// TODO: Overapproximate if second is NaN
+		new OverapproxVariable("sign of NaN", loc).annotate(nondet);
+		builder.addStatement(StatementFactory.constructIfStatement(loc, mFloatHandler.isNan(loc, first, type),
```
schuessf, line 1013: @danieldietsch @Heizmann As you have worked with floats in the past, does this handling of `copysign` (incl. the overapproximation) seem reasonable?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/library/MathLibraryModel.java
```diff
@@ -532,6 +596,464 @@ private Result handleIsLessGreater(final IDispatcher main, final IASTFunctionCal
...
+												IASTUnaryExpression.op_minus, abs, type))))))));
+		return builder.setLrValue(new RValue(auxVar.getExp(), type)).build();
+	}
+
+	private ExpressionResult handleRemainder(final IDispatcher main, final IASTFunctionCallExpression node,
+			final ILocation loc, final String name, final CPrimitive type) {
+		final List<ExpressionResult> arguments = handleFloatArguments(main, node, loc, name, 2, type);
+		final Expression first = arguments.get(0).getLrValue().getValue();
+		final Expression second = arguments.get(1).getLrValue().getValue();
+		return new ExpressionResultBuilder().addAllExceptLrValue(arguments)
```
schuessf, line 1033: @danieldietsch In https://github.com/ultimate-pa/ultimate/commit/042d866d19cb7f43a77b7bd04199d4770735fc6f you removed also the support for `remainder`, claiming in [this comment](https://github.com/ultimate-pa/ultimate/issues/398#issuecomment-442757689) that our handling was also unsound. In this PR, however, I added support again using the SMT function `fp.rem`. The documentations for [`fp.rem`](https://smt-lib.org/theories-FloatingPoint.shtml) and [`remainder`](https://en.cppreference.com/w/c/numeric/math/remainder.html) seemed to match and I could not find any unsoundness. Do you remember the issue with `remainder`?

Issue #749, opened 2025-09-19
Is it possible to compile on arm?


paolo-di-biase:
I need the tool compiled for arm for some experiments using benchexec, so I'd need the compiled file as you release it for sv-comp but for the arm architecture. I was wondering if I can have some compilation instructions since I can't find them, or if you already have the compiled versions for arm. Thanks.

maul-esel:
You can find information on building Ultimate [in our wiki](https://github.com/ultimate-pa/ultimate/wiki/Usage#building-ultimate).

Generally, I'd expect Ultimate itself to run on ARM processors without changes, given a suitable Java runtime. However, you will likely have to create/download ARM builds of the SMT solvers (z3, cvc4/5, mathsat, ...) to replace the binaries that are shipped with Ultimate (located in `releaseScripts/default/adds` in the repo, resp. in the ZIP files created by the build script).

Maybe our build system experts @danieldietsch @bahnwaerter can point out if I missed something?

bahnwaerter:
We currently provide Ultimate for the `x86-64` architecture on Windows, Linux, and MacOS. As @maul-esel already pointed out, ARM support is conceptually possible. However, this would require not only building the solvers for ARM, but also replacing platform-specific components of our used framework (Equinox). We don't have support for this, but we may add it in the future.

@paolo-di-biase do you need Ultimate for the 64-bit ARM (`aarch64`) architecture? And what operating system are you using?

paolo-di-biase:
oKk thanks for the reply.
Yes i need it for aarch64 architecture and Linux operating system, specifically Linux raspi

maul-esel:
@bahnwaerter are these the same steps we would also have to take if we want ARM-based docker images that run on MacOS host machines?

bahnwaerter:
> [@bahnwaerter](https://github.com/bahnwaerter) are these the same steps we would also have to take if we want ARM-based docker images that run on MacOS host machines?

Yes, we could use that setup for the building of ARM-based Docker images, too. All those images are based on a simple Linux distribution (Alpine Linux) and are build to match the specific target architecture (`x86-64` or `aarch64`). First experiments for the `x86-64` architecture have shown, that this approach runs smooth on Windows using Docker Desktop. I'm sure this also works on MacOS.

bahnwaerter:
> oKk thanks for the reply. Yes i need it for aarch64 architecture and Linux operating system, specifically Linux raspi

[...]

PR #757, opened 2025-11-07
Migrate framework to Eclipse E4.37 and update Maven build plugins

bahnwaerter:
The changes migrate the Equinox framework with all its components from Eclipse E4.32 to E4.37. The Java version remains unchanged at version 21. In addition, the Maven build plugins are updated to their latest stable version. Architecture and operating system-specific platform targets are obsolete and are no longer needed, since the specific framework dependencies (Eclipse plugins) are now listed in the feature files of Ultimate. From now on, only the general `Ultimate.target` needs to be used as the platform target and all required framework dependencies will then be resolved automatically according to the content of the feature files.


bahnwaerter:
These changes also facilitate the implementation of the desired feature in #749. However, the implementation will be done in a separate pull request.

bahnwaerter:
The changes have already been tested manually. It is possible to build and run Ultimate based on Eclipse E4.37 under
 - [x] Linux (Arch Linux)
 - [x] macOS Ventura
 - [x] Windows 10

while using Eclipse 2025-09 as IDE.

It is also possible to build Ultimate using Maven on Linux (Arch Linux) where a Maven version greater than 3.9.9 is required for this.

trunk/source/BA_SiteRepository/Ultimate.target
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
schuessf, line 1: Is there any reason you renamed this file? I think, this makes migration a bit less convenient and requires a small change in our wiki for the installation instructions.

trunk/source/BA_SiteRepository/Ultimate.target
```diff
@@ -0,0 +1,113 @@
...
+            <unit id="org.eclipse.cdt.feature.group" version="0.0.0"/>
+            <unit id="org.eclipse.cdt.sdk.feature.group" version="0.0.0"/>
+            <unit id="org.eclipse.equinox.sdk.feature.group" version="0.0.0"/>
+            <unit id="org.eclipse.sdk.feature.group" version="0.0.0"/>
+            <unit id="org.eclipse.jdt.feature.group" version="0.0.0"/>
+            <unit id="org.eclipse.equinox.server.jetty.feature.group" version="0.0.0"/>
+        </location>
+        <location includeAllPlatforms="true" includeConfigurePhase="true" includeMode="planner" includeSource="false" type="InstallableUnit">
+            <repository id="orbit-aggregation" location="https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2025-09"/>
+            <unit id="org.eclipse.orbit.legacy.feature.group" version="0.0.0"/>
```
schuessf, line 19: You replaced all the versions by 0.0.0. Does this mean that any version that is found in the repo is accepted and thus makes future framework updates even simpler?

trunk/source/BA_SiteRepository/Ultimate.target
bahnwaerter: I did that on purpose, because it's a bad idea to include version numbers in a filename when using source code versioning. With every framework migration, we would have to rename the file and lose its versioning history. After this one-time renaming, that should no longer be the case.

trunk/source/BA_SiteRepository/Ultimate.target
bahnwaerter: > [...] makes future framework updates even simpler?

You recognized that correctly. The version information for all dependencies will now be resolved automatically (according to the version constraints of the plugins). This applies to both the bundles and each individual plugin. This approach for RCP development is now set as the new default in the Eclipse IDE.

schuessf (approved):
So, then this looks good to me. The question is, should we merge this before SV-COMP or is this somehow "too risky"?

bahnwaerter:
> [...] The question is, should we merge this before SV-COMP or is this somehow "too risky"?

I would tend to merge these changes after SV-COMP. We also don't have any performance comparison yet. Also, it's unclear whether the Eclipse compiler update will cause further problems (as it was last year the case).

schuessf:
With this framework migration, the version of `org.eclipse.equinox.launcher` has also changed, right? If this is the case, several hardcoded version strings (e.g., in [`Ultimate.py`](https://github.com/ultimate-pa/ultimate/blob/23b7c4e0426ba1abdc7a3c014efbfda56e6fdb47/releaseScripts/default/adds/Ultimate.py#L295)) still have to be adapted.

bahnwaerter:
> With this framework migration, the version of `org.eclipse.equinox.launcher` has also changed, right? [...]

That's correct, thanks for the hint. I just adjusted the filename of the launcher Jar in the startup scripts.

schuessf:
I just rebased this MR and started the nightly to check if it succeeds (required 7826bda as a small fix).
Since Eclipse 2025-12 is out now, can you check if we can directly upgrade to this version (not urgent though)?

schuessf:
Unfortunately, there was still some issue that lead to all tests failing, but I hopefully fixed that in ~~26d3096~~ f39e5c0.
Also, I just tried updating to 2025-123 (and cleaning up the target file): https://github.com/ultimate-pa/ultimate/tree/wip/fs/update-target
Let's see, if it is working 🙂

bahnwaerter:
The preparation for this pull request is now complete. @danieldietsch and @maul-esel, would you like to review the changes to ensure everything is okay before merging?

danieldietsch (approved):
Looks good to me

maul-esel:
I can try to take a look this evening. But if I don't manage or you want to deal with this today, I'm fine if you merge directly.

PR #758, opened 2025-11-13
Add C translation support for anonymous structs and unions

bahnwaerter:


trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/TypeHandler.java
```diff
@@ -430,6 +431,16 @@ public Result visit(final IDispatcher main, final IASTCompositeTypeSpecifier nod
 					bitFieldWidths.add(declaration.getBitfieldSize());
 				}
 			} else if (r instanceof SkipResult) { // skip ;)
+			} else if (r instanceof final TypesResult tr && tr.getCType() instanceof final CStructOrUnion su) {
+				if (su.isAnonymous()) {
+					// Flat unnamed (anonymous) struct or union by adding the field(s) to to the parent struct or union.
+					fNames.addAll(su.getFieldNames());
+					fTypes.addAll(Arrays.asList(su.getFieldTypes()));
+					bitFieldWidths.addAll(su.getBitFieldWidths());
```
maul-esel, line 439: I understand the idea to flatten anonymous structs into a parent struct; and I guess, similarly, to flatten anonymous unions into a parent union.

But does it also make sense to flatten an anonymous struct into a parent union, or vice versa? A struct is a kind of product type (basically a tuple, but the fields are named), whereas a union is a sum type (only one of the values is present, not all). It seems to me that this information is lost by flattening one into the other, isn't it?

maul-esel:
Thanks for looking into this issue :+1:

I don't fully understand the conceptual approach yet, see comment above.

Also, does your fix work if there are multiple anonymous structs / unions inside a parent struct / union? Please add some tests for this.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/TypeHandler.java
schuessf: Yes, after I thought more about this, I think that the should not be flattened for this reason. At least the `sizeof` computation could be fixed for that reason. I guess the better solution would be to just have a recursive lookup in `CStructOrUnion` instead.

maul-esel: :+1: I was thinking in a similar direction: structs and unions could have special handling for anonymous fields (let's be careful to avoid "name clashes" because they all have the field name `""`), and then the field lookup should recurse into these anonymous fields.

Relatedly, is it also possible in C to have an anonymous field but where the struct type is named? (and perhaps declared separately?)

bahnwaerter: You're right, @maul-esel. So far, anonymous unions are not supported correctly.

> I was thinking in a similar direction: structs and unions could have special handling for anonymous fields (let's be careful to avoid "name clashes" because they all have the field name ""), and then the field lookup should recurse into these anonymous fields.

That would be indeed a great conceptual idea. Unfortunately, our representation of anonymous structs and unions is not created exactly like you described it. Declaration nodes with empty names are not stored as valid declarations. This means that anonymous structs and unions cannot be members of a parent type. We could add more hacks but I wanted to keep the changes as simple as possible.

schuessf: I would just suggest to have a separate member in `CStructOrUnion` for anonymous fields and to adapt the methods (recursive lookup in anonymous fields, somehow return the fields in getters).

PR #764, opened 2025-12-05
C translation: Handle incomplete arrays properly

schuessf:
In C, an incomplete array refers to an array type whose size is not yet specified. A common special case are incomplete array members in struct (called flexible arrays) are ignored in the `sizeof` computation. While this behavior is already supported in Ultimate (see commit 2a975d2), the check for whether an array is incomplete is currently quite hacky. It relies on the magic number [`INCOMPLETE_ARRY_MAGIC_NUMBER`](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/container/c/CArray.java#L57). We mark a `CArray` type as incomplete when its size equals this magic number.

This approach works for integer translation, since array sizes can't be negative. However, it fails in the bitvector translation, where arithmetic is always unsigned. In that case, a wraparound occurs, and the result never matches `INCOMPLETE_ARRY_MAGIC_NUMBER`. Therefore, some of our Nightly tests (e.g., [`flexible_array.c`](https://github.com/ultimate-pa/ultimate/blob/dev/trunk/examples/programs/regression/c/flexible_array.c)) currently fail. The same issue also leads to incorrect behavior when allocating memory for incomplete arrays.

This PR removes the use of the magic number `INCOMPLETE_ARRY_MAGIC_NUMBER`. Instead, `CArray` now includes an explicit field for the bound (an expression) that may be `null`. An incomplete array becomes complete when it is initialized, based on the length of its initializer (see C11 6.7.9.22). This should fix the known issues with incomplete arrays, including those in the bitvector translation.

### DIFF
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
@@ -1993,7 +1993,6 @@ public class CHandler {
 		 * declarator: (e.g. (int [])): then the size depends on the initializer - otherwise: the size is given by the
 		 * CType
 		 */
-		mTypeSizeComputer.constructBytesizeExpression(loc, cType);
 
 		final ExpressionResultBuilder builder = new ExpressionResultBuilder();
 
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
@@ -1342,7 +1342,7 @@ public class CHandler {
 			// the innermost type is the value type.
 			ICType arrayType = resType.getCType();
 
-			final CPrimitive arrayIndexCtype = mTypeSizes.getSizeT();
+			final CPrimitive arrayIndexCtype = mExpressionTranslation.getCTypeOfPointerComponents();
 
 			// expression results of from array modifiers
 			final ArrayList<ExpressionResult> expressionResults = new ArrayList<>();
@@ -1359,7 +1359,7 @@ public class CHandler {
 					final ExpressionResult dispatched = (ExpressionResult) main.dispatch(am.getConstantExpression());
 					final ExpressionResult switched = mExprResultTransformer.switchToRValue(dispatched, loc, node);
 					final ExpressionResult converted =
-							mExpressionTranslation.convertIntToInt(loc, switched, mTypeSizes.getSizeT());
+							mExpressionTranslation.convertIntToInt(loc, switched, arrayIndexCtype);
 					expressionResults.add(converted);
 					final RValue rValue = (RValue) converted.getLrValue();
 					bound = rValue.getValue();
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
@@ -1341,8 +1341,7 @@ public class CHandler {
 		if (node instanceof final IASTArrayDeclarator arrDecl) {
 			// the innermost type is the value type.
 			ICType arrayType = resType.getCType();
-
-			final CPrimitive arrayIndexCtype = mExpressionTranslation.getCTypeOfPointerComponents();
+			final CPrimitive boundType = mExpressionTranslation.getCTypeOfPointerComponents();
 
 			// expression results of from array modifiers
 			final ArrayList<ExpressionResult> expressionResults = new ArrayList<>();
@@ -1357,8 +1356,7 @@ public class CHandler {
 					// e.g., a[23] or a[n+1]
 					final ExpressionResult dispatched = (ExpressionResult) main.dispatch(am.getConstantExpression());
 					final ExpressionResult switched = mExprResultTransformer.switchToRValue(dispatched, loc, node);
-					final ExpressionResult converted =
-							mExpressionTranslation.convertIntToInt(loc, switched, arrayIndexCtype);
+					final ExpressionResult converted = mExpressionTranslation.convertIntToInt(loc, switched, boundType);
 					expressionResults.add(converted);
 					final RValue rValue = (RValue) converted.getLrValue();
 					bound = rValue.getValue();
@@ -1372,7 +1370,7 @@ public class CHandler {
 							throw new UnsupportedOperationException("expected IASTEqualsInitializer");
 						}
 						intSizeFactor = computeSizeOfInitializer((IASTEqualsInitializer) arrDecl.getInitializer());
-						bound = mTypeSizes.constructLiteralForIntegerType(loc, arrayIndexCtype,
+						bound = mTypeSizes.constructLiteralForIntegerType(loc, boundType,
 								BigInteger.valueOf(intSizeFactor));
 					} else if (resType.getCType() instanceof CFunction) {
 						// if we have an array of function pointers,
@@ -1386,7 +1384,7 @@ public class CHandler {
 							throw new UnsupportedOperationException("expected initializer");
 						}
 						intSizeFactor = computeSizeOfInitializer((IASTEqualsInitializer) fundecl.getInitializer());
-						bound = mTypeSizes.constructLiteralForIntegerType(loc, arrayIndexCtype,
+						bound = mTypeSizes.constructLiteralForIntegerType(loc, boundType,
 								BigInteger.valueOf(intSizeFactor));
 					} else {
 						// we have an incomplete array type without an initializer --
@@ -1397,7 +1395,7 @@ public class CHandler {
 				} else {
 					throw new IncorrectSyntaxException(loc, "wrong array type in declaration");
 				}
-				arrayType = new CArray(bound, arrayIndexCtype, arrayType);
+				arrayType = new CArray(bound, boundType, arrayType);
 			}
 			final ExpressionResult allResults = new ExpressionResultBuilder()
 					.addAllExceptLrValue(expressionResults.toArray(new ExpressionResult[expressionResults.size()]))
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
@@ -1342,6 +1342,8 @@ public class CHandler {
 			// the innermost type is the value type.
 			ICType arrayType = resType.getCType();
 
+			final CPrimitive arrayIndexCtype = mTypeSizes.getSizeT();
+
 			// expression results of from array modifiers
 			final ArrayList<ExpressionResult> expressionResults = new ArrayList<>();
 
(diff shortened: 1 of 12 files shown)


schuessf:
Note: This issue in the bivector translation was also responsible for one unsound task in SV-COMP 2026 ([`aws_string_destroy_secure_harness_negated.yml`](https://sv-comp.sosy-lab.org/2026/results/results-verified/META_C.Overall_uautomizer.table.html#/table?filter=id(values(aws_string_destroy_secure_harness_negated.yml,,)))). Luckily, we found another issue in this task ([!1722](https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/merge_requests/1722)), so we could exclude it for this year anyway 😆

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/container/c/CArray.java
```diff
@@ -50,17 +48,11 @@
...
 public final class CArray implements ICType {
-
-	/**
-	 * Size that we use to indicate that an array has a variable length.
-	 */
-	public static final int INCOMPLETE_ARRY_MAGIC_NUMBER = -1234567;
-
 	/**
 	 * Array bound. Note that we use nesting of array types for multidimensional array types
 	 */
```
maul-esel, line 53: Please update the documentation comment.

Also, what is the meaning of the "Note that..." sentence? Does it apply to the bound, or to the value type?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/container/c/CArray.java
```diff
@@ -71,31 +63,43 @@ public final class CArray implements ICType {
...
 		return mValueType;
 	}
 
+	public void complete(final Expression bound) {
+		if (mBound != null) {
+			throw new AssertionError("only incomplete arrays can be completed");
+		}
+		mBound = bound;
+	}
+
```
maul-esel: Making the type mutable in this way seems risky. Types that were previously considered equal will be non-equal after a call to this method. The hashcode will change, which is dangerous if the type is stored in some hash set or hash map.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
```diff
@@ -2168,8 +2169,8 @@ private Result handleStringLiteralExpression(final ILocation loc, final IASTLite
 		final Expression sizeInBytesExpr = mTypeSizes.constructLiteralForIntegerType(actualLoc,
 				mExpressionTranslation.getCTypeOfPointerComponents(), BigInteger.valueOf(sizeInBytes));
 
-		final RValue dimension = new RValue(sizeInBytesExpr, mExpressionTranslation.getCTypeOfPointerComponents());
-		final CArray arrayType = new CArray(dimension, new CPrimitive(CPrimitives.CHAR));
+		final CArray arrayType = new CArray(sizeInBytesExpr, mExpressionTranslation.getCTypeOfPointerComponents(),
```
maul-esel, line 2173: Above, `mTypeSizes.getSizeT()` is seemingly used as "default" type for array bounds, here it is `mExpressionTranslation.getCTypeOfPointerComponents()`. I am not sure which one (if either) is the correct choice, but I guess we should look it up and use it consistently.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/ArrayHandler.java
```diff
@@ -260,8 +263,8 @@ private void addArrayBoundsCheckForCurrentIndex(final ILocation loc, final RValu
 		final Expression nonNegative = mExpressionTranslation.constructBinaryComparisonExpression(loc,
 				IASTBinaryExpression.op_lessEqual, zero, indexType, currentIndex.getValue(), indexType);
 		final Expression notTooBig = mExpressionTranslation.constructBinaryComparisonExpression(loc,
-				IASTBinaryExpression.op_lessThan, currentIndex.getValue(), indexType, currentDimension.getValue(),
-				(CPrimitive) currentDimension.getCType().getUnderlyingType());
+				IASTBinaryExpression.op_lessThan, currentIndex.getValue(), indexType, currentDimension,
+				(CPrimitive) currentBoundType.getUnderlyingType());
```
maul-esel: Do we always expect the bound type to be a primitive? (that seems reasonable to me). If so, why not give `CArray.mBound` directly the type `CPrimitive`? (and update the corresponding getter method, as well as the parameter type here)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/InitializationHandler.java
```diff
@@ -1219,9 +1219,15 @@ private InitializerInfo constructIndexToInitInfo(final ILocation loc,
...
 			cellType = array.getValueType();
-			bound = CTranslationUtil.getConstantFirstDimensionOfArray(array, mTypeSizes);
-			if (CTranslationUtil.isToplevelVarlengthArray(array, mTypeSizes)) {
-				throw new UnsupportedOperationException("varlength not yet supported here");
+			if (array.isIncomplete()) {
+				// C11 6.7.9.22:
+				// If an array of unknown size is initialized, its size is determined by the largest indexed element
+				// with an explicit initializer. The array type is completed at the end of its initializer list.
+				bound = initializerResults.size();
+				array.complete(mExpressionTranslation.constructLiteralForIntegerType(loc,
```
maul-esel: See comment in `CArray` about making the class mutable. Are there other ways of implementing this that avoid mutability?

maul-esel (commented):
Thanks @frank for taking care of this! I've left some comments; the main concern would be the mutability of `CArray` I guess.

Other than that, it looks good as far as I can see. It's a bit hard to judge the effects in our C translation, so should we perhaps to a `CTranslationOnly` run over sv-benchmarks to check for any unintended side effects?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/container/c/CArray.java
schuessf: > Making the type mutable in this way seems risky.

I agree, this is not optimal. I just tried to implement it the same way as for `CStructOrUnion` (which is of course also broken...). While there is also a `complete` method that returns a new type, rather than modifying an existing one, it is actually currently never used. But I just found that for `CEnum` this immutable `complete` is used, together with the field `TypeHandler::mIncompleteType`. I will try, if a similar handling is also possible for `CArray`.

> Types that were previously considered equal will be non-equal after a call to this method.

This is somehow expected. If you have the two definitions of arrays:
```c
int x[] = {1,2,3};
int y[] = {1,2,3,4};
```
the types of `x` and `y` are actually different after their completion (because of the different size). However, I am not quite sure, if we actually need this equals comparison at all.

maul-esel: > I will try, if a similar handling is also possible for CArray.

:+1: 

> the types of x and y are actually different after their completion (because of the different size).

I agree. It just seems risky then to represent them using the same instances that were previously used for the "equal" incomplete types.

> However, I am not quite sure, if we actually need this equals comparison at all.

Yes, if there is no way to avoid mutability, it may be more sensible to just implement `equals` as reference-equality (and ensure that the hashcode is not mutable).

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/ArrayHandler.java
schuessf: Indeed, the bound is always a `CPrimitive`, therefore I changed it. The only case, where the type was not directly primitive was when an explicit bound was given (e.g., `int x[5]`), but there it was converted to a `size_t` anyways.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java
schuessf: The C standard states:
> The definition of the subscript operator [] is that E1[E2] is identical to (*((E1)+(E2))).

Therefore, I think `mExpressionTranslation.getCTypeOfPointerComponents()` is actually a better solution, as arrays should be handled exactly the same as pointers.

Also, I am wondering why the bound type is included in `CArray` at all. It should be always the same type `mExpressionTranslation.getCTypeOfPointerComponents()`.

maul-esel: >  I think mExpressionTranslation.getCTypeOfPointerComponents() is actually a better solution, as arrays should be handled exactly the same as pointers.

Sounds good. Is that solution compatible with all the different memory models we now support?

> Also, I am wondering why the bound type is included in CArray at all. It should be always the same type mExpressionTranslation.getCTypeOfPointerComponents().

I was wondering the same thing...

schuessf: > Also, I am wondering why the bound type is included in `CArray` at all. It should be always the same type `mExpressionTranslation.getCTypeOfPointerComponents()`.

Currently, it still has to be included, since otherwise `CArray` is no more self-contained and `TypeHandler::cType2AstType` has no access to `ExpressionTranslation`. Therefore, I think we should just keep it as it is.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/InitializationHandler.java
schuessf: I just debugged the code and it turns out that this part is actually never reached, when we have just "normal" array initializations (like `int x[] = {1,2,3};`). Such array initializations are handled in [`CHandler::visit(IASTDeclarator)`](https://github.com/ultimate-pa/ultimate/blob/5fbc7936f8e6aa4adef141e5fc50485fc248bd53/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java#L1325), including an [on-the-fly completion](https://github.com/ultimate-pa/ultimate/blob/5fbc7936f8e6aa4adef141e5fc50485fc248bd53/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java#L1371) of the array type based on the initializer.

This completion was only necessary for compound literals (such as `(int []) { 1, 2 }`), therefore it seems like it should be working, if we just complete the array type before calling this method (in  [`CHandler::visit(IASTTypeIdInitializerExpression)`](https://github.com/ultimate-pa/ultimate/blob/5fbc7936f8e6aa4adef141e5fc50485fc248bd53/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/CHandler.java#L1926)) rather than modifying it here. But we will see in the nightly, if there are any more issues with that 😉

schuessf:
@maul-esel I tried to adress the issues pointed out in your previous review, most notably, I made `CArray` immutable again. Also, all tests in our nightly seem to pass. Can you please review this PR again with the newest changes?

maul-esel (approved):
Looks good, thanks :+1:

It might still be worth doing a CTranslation-run on a larger benchmark set, but I leave that up to you.

PR #765, opened 2025-12-16
ACSLPrettyPrinter: omit parentheses when allowed by operator precedence

maul-esel:
Currently, our generated invariants are often very difficult to read due to excessive use of parentheses around almost all sub-expressions.

With these changes, we take operator precedence into account to omit parentheses where they are not needed. This should lead to much more readable invariants.


trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ACSLPrettyPrinter.java
```diff
@@ -75,6 +75,8 @@ public class ACSLPrettyPrinter {
 	private static final String STRING_TIMES = "*";
 	private static final String STRING_PLUS = "+";
 
+	private static final int UNARY_PREFIX_PRECEDENCE = 200;
```
schuessf: Why  this magic number 200?

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ACSLPrettyPrinter.java
```diff
@@ -212,4 +245,121 @@ public static String binaryOperatorToString(final BinaryExpression.Operator oper
 		case COMPPO, BITVECCONCAT -> throw new AssertionError("Unhandled operator " + operator);
 		};
 	}
+
+	// https://en.cppreference.com/w/c/language/operator_precedence.html
```
schuessf: We should refer to the ACSL standard instead, even though I don't expect much of a difference. In section 2.2.1 there is the following table:
grafik

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ACSLPrettyPrinter.java
```diff
@@ -212,4 +245,121 @@ public static String binaryOperatorToString(final BinaryExpression.Operator oper
...
+		// use numbers in the range 100-200 that are higher than any precedence given for binary operators
+		return switch (operator) {
+		case PLUS, MINUS -> 101;
+		case LTLGLOBALLY, LTLFINALLY, LTLNEXT -> 102;
+		case LOGICNEG, LOGICCOMPLEMENT, ADDROF, POINTER -> UNARY_PREFIX_PRECEDENCE;
+		};
+	}
+
+	// Follows the precedence defined in our parser (see GlobalLocalParser.cup).
+	// (See also: https://en.cppreference.com/w/c/language/operator_precedence.html)
```
schuessf: Also refer to the ACSL standard here.

maul-esel:
@schuessf I tested this with `WitnessRegressionTestSuite` and it seems fine. Do you think there are other ways we should test it before merging?

@Heizmann and @schuessf: How stingy do we want to be with parentheses? Do you know of any cases where it might be better to have parentheses even though they are not strictly necessary? What are your opinions on the following cases:
- I implemented it so that `&&` and `||` have the same precedence, and therefore parentheses are added.

  I find the invariant
  ```
  (x == 3 && i == 1 && y == 0) || (x == 3 && i == 0 && y == 0) || (x == 3 && i == 3 && y == 3) || (x == 3 && y == 1 && i == 2)
  ```
  (generated with these changes) much more readable than if we omitted all parentheses from it.
- Perhaps casts are another relevant case.

  I am always unsure how to read the invariant
  ```
  k == ((long long) i * 3) % 4294967296 && j + (i / 2147483648) * 4294967296 == (long long) i * 2 && (long long) i * 3 <= k && (long long) i * 2 <= j
  ```
  (generated with these changes) -- I guess the cast to `long long` applies only to `i`, not the product `i * 3`? (would make sense, as the intention of these casts is to prevent overflow.)

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ACSLPrettyPrinter.java
```diff
@@ -212,4 +245,121 @@ public static String binaryOperatorToString(final BinaryExpression.Operator oper
...
+		case final CastExpression castExp -> UNARY_PREFIX_PRECEDENCE;
+
+		case final FieldAccessExpression faExp -> 300;
+
+		case final IfThenElseExpression iteExp -> 3;
+		case final QuantifierExpression quantExp -> 2;
+		case final SyntacticNamingExpression synExp -> 0;
+
+		// unambiguous / highest precedence
+		case final ACSLResultExpression resExp -> Integer.MAX_VALUE;
```
schuessf: Is there any nicer solution than using magic numbers and `Integer.MAX_VALUE`?

schuessf (commented):
Thank you for taking care of this! This should indeed improve the readability of our invariants 🙂 
I have left a few comments.

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ACSLPrettyPrinter.java
maul-esel: I have been wondering that as well. Perhaps we could define a `Precedence` enum? I wasn't sure, but a quick search suggests that enums in Java are always ordered as they are declared (in terms of their `Comparable` implementation).

schuessf:
> I tested this with `WitnessRegressionTestSuite` and it seems fine. Do you think there are other ways we should test it before merging?

In our regression tests this is the only way to test this. Maybe a quick SV-COMP with witness validation would be helpful, just to be sure everything is still working as expected.

> How stingy do we want to be with parentheses? Do you know of any cases where it might be better to have parentheses even though they are not strictly necessary? What are your opinions on the following cases:
> 
>   *  I implemented it so that `&&` and `||` have the same precedence, and therefore parentheses are added.
>       I find the invariant
>       ```
>       (x == 3 && i == 1 && y == 0) || (x == 3 && i == 0 && y == 0) || (x == 3 && i == 3 && y == 3) || (x == 3 && y == 1 && i == 2)
>       ```

This sound reasonable to me. I see, you already added a comment for this precedence, maybe you can make the comment a bit more clear.

>    * Perhaps casts are another relevant case.
>       I am always unsure how to read the invariant
>       ```
>       k == ((long long) i * 3) % 4294967296 && j + (i / 2147483648) * 4294967296 == (long long) i * 2 && (long long) i * 3 <= k && (long long) i * 2 <= j
>       ```     
>       (generated with these changes) -- I guess the cast to `long long` applies only to `i`, not the product `i * 3`? (would make sense, as the intention of these casts is to prevent overflow.)

Here, I would still prefer `(long long) i * 3` over `((long long) i) * 3`, it seems more readable to me. The cast to `long long` is of course only applied to `i`, but because of that cast the result has also type `long long` -- the cast is only done before the multiplication to avoid overflows.

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ACSLPrettyPrinter.java
maul-esel: I have changed the code to use an enum.

maul-esel:
> Maybe a quick SV-COMP with witness validation would be helpful, just to be sure everything is still working as expected.

I could try and do a run of the inductive-witness validation experiments later.

> I see, you already added a comment for this precedence, maybe you can make the comment a bit more clear.

Done.

> Here, I would still prefer (long long) i * 3 over ((long long) i) * 3, it seems more readable to me. 

Ok, I have no strong opinion on this. :+1:

trunk/source/ACSLParser/src/de/uni_freiburg/informatik/ultimate/model/acsl/ACSLPrettyPrinter.java
maul-esel: fixed

schuessf (approved):

Heizmann:
* I think it is a very good idea to improve the readability of the expressions.
* Putting parenthesis around disjunctions if they occur in a conjunction seems a good choice to me.
* Regarding other choices I do not have preferences.
* I am not very familiar with the topic. Maybe there is some literature on readability of expressions or best practices that are applied by other tools.
* However, I am a bit worried about the terminology of the implementation. Same *precedence* for `&&` and `||`. That sounds very wrong to me. The ACSL documentation says something different. Can we maybe use a different word here?

maul-esel:
> However, I am a bit worried about the terminology of the implementation. Same precedence for && and ||. That sounds very wrong to me. The ACSL documentation says something different. Can we maybe use a different word here?

My instinct was to stick to the standard terminology. The deviation from the ACSL documentation is explicitly explained in a comment (`We pretend &&, ^^ and || have the same precedence`). Do you have a suggestion for an alternative terminology?

maul-esel:
I've merged the changes for now. We can easily rename the enum if there is a better suggestion than `Precedence`.

Heizmann:
> Do you have a suggestion for an alternative terminology?

* PrettyPrintPrecedence
* PrintPrecedence
* PPrecedence
* PpPrecedence
* OutputPrecedence
* OPrecedence

But if the enum stays private and stays in a class that is about prettyprinting my preference would be rather to keep the enum's name as it is.

PR #766, opened 2025-12-16
Boogie2ACSL backtranslation improvements

maul-esel:
The main point of these changes is to support backtranslation of if-then-else expressions. These are needed for ghost updates in our Owicki-Gries annotations (not yet merged to `dev`).

In order to support these expressions properly, I first did some small refactorings in `Boogie2ACSL`.

@schuessf I think you are most familiar with this code. Could you take a look and see if it seems sensible?


schuessf (approved):
Great! These changes seem reasonable to me.
So just to clarify: The only functional difference is that in some contexts conjuncts or disjuncts that cannot be backtranslated are not omitted anymore, but the backtranslation yields `null`?

maul-esel:
> in some contexts conjuncts or disjuncts that cannot be backtranslated are not omitted anymore, but the backtranslation yields null?

Yes, but that should only happen for conjunctions / disjunctions that appear inside a `FunctionApplication`, an `ArrayAccessExpression`, or a `BinaryOperation` that is not itself a disjunction / conjunction. These cases should be extremely rare, if they ever occur at all.

The other functional change is that we now support `IfThenElseExpression` instead of just returning `null`.

PR #756, opened 2025-11-06
Add 1-Dimensional memory model for memory safe tasks

jankoerner:
This pull request introduces a 1-dimensional memory model which can be used for ReachSafety tasks.
The 1-dimensional memory model comes with some major refactorings of the source code.
These refactoring can be seen in the attached image and allow an easy extension for new Addressings, Structures, ManagementStrategies and Metadatas (not visible in the image). 
2025-11-06T20:16:22,830512837+01:00

PR #768, opened 2025-12-21
Some memory model simplifications

maul-esel:
Since the dedicated work by @jankoerner in #756, our C translation uses general interfaces to encapsulate different memory model implementations, rather than hardcoding a single memory model all throughout the translation (thanks again for this work!).

This PR collects some simplifications and refactorings of the code for which there was not enough time in the project / before SV-COMP; with the goal to improve maintainability and make it easier to implement further memory models in the future.

_To review the changes in this PR, it might be a good idea to view them commit-by-commit._

In addition to the changes already here, the following items could also be improved (some of them may simply require more documentation):

- [ ] It would probably make sense to move a lot of this code into a dedicated sub-package, to bundle it in one place. Then, we should also see which parts need to be public and what can be package-private.
- [x] what is the meaning of the lists of tuples returned by
  - [x] `IMemoryAddressing::constructMallocSpecificationExpressions`
  - [x] `IMemoryAddressing::constructDeallocSpecificationExpressions`
  - [x] `IMemoryAddressing::constructAllocInitSpecificationExpressions`
- [ ] What is the meaning of these methods? (Improve documentation or reconsider if they should be implemented differently.)
  - [ ] `IMemoryAddressing::constructUltimateInitStatements` (what do these statements do?)
  - [ ] `IMemoryAddressing::getFixedAddressCounterCountingStep`
  - [ ] `IMemoryAddressing::constructRhsAssignmentStatementHda`
- [ ] what is the relation between the methods `::doPointerArithmetic`, `::addExpressionToPointer`, `addIntegerConstantToPointer` and `::doPointerSubtraction` on `IMemoryAdressing`? Can some of them be implemented in terms of the others?
- [ ] These methods seem overly specific. Can they be implemented by a caller, using the other existing methods?
  - [x] `IMemoryAddressing::constructAddressForStructField`  (perhaps implementable using `::addIntegerConstantToPointer`?)
  - [x] `IMemoryAddressing::getLastCharOfString`
  - [ ] `IMemoryAddressing::constructStrChrAssumeStatement` (if not, improve documentation and let it return an expression)
  - [ ] `IMemoryAddressing::checksForStringCopyOverlapping` (otherwise, document parameters and return an expression)
  - [ ] `IMemoryAddressing::constructReallocBodyStatements`
- [ ] Having two methods `IMemoryAddressing::constructPointerValidityCheckExpr` and `::constructPointerTargetFullyAllocatedCheckExpr` seems specific to the 2-dimensional model. Can they be merged into a single method? (that perhaps returns a list of expressions ~~; and callers would need to take care of the `size` parameter separately e.g. through a call to `::addExpressionToPointer`~~)
- [x] `IMemoryAddressing::getChecksForFreeCall` should return expressions (or `ExpressionResult`s) rather than assert statements
- [x] Can `IMemoryAddressing::constructMemSafeStatementsForPointerExpression` be implemented in terms of `::constructPointerValidityCheckExpr` and `::constructPointerTargetFullyAllocatedCheckExpr`?
- [ ] `IMemoryAddressing::constructInitialPointerFromPointer` only makes sense for the 2-dimensional model. It is used for variadic arguments; they must be implemented in another way.
- [x] `IMemoryAddressing::getValidArray` only makes sense for the 2-dimensional model. It is used only for memory neutrality checks; these should be implemented like `::constructPointerValidityCheckExpr` etc.
- [ ] An instance of `IMemoryPointer` is not actually a pointer. Can we rename the interface?
- [ ] (where) is `LocalLValueILocationPair` needed? (if we keep it, make it a record, possibly private, and improve field names)
- [ ] `MemoryAddressingFactory` violates the open-closed principle (for every new `IMemoryPointer` implementation, we would need to change this class). Can this functionality become a method on `IMemoryPointer`?
- [ ] The construction of a suitable pointer-integer conversion in `MemoryAddressingBase` violates the open-closed principle (for every new `IMemoryPointer` implementation, we would need to change this class). Can this functionality become a method on `IMemoryPointer`?
- [ ] update the documentation (in particular, of parameters) for `MemoryHandler::getWriteCall`
- [ ] update the documentation for `MemoryHandler::constructMemsafetyChecksForPointerExpression`
- [ ] document `MemoryStructureMultiBitprecise`
- [ ] documentation for `SimpleIncreasingStrategy` mentions a type parameter that does not exist
- [ ] The documentation for `NonBijectiveMapping1D` talks about base and offset (copy-pasted from 2D case). Is the mapping even non-bijective in the 1-dimensional case?
- [x] The fields `SFO.MEMCPY` and `SFO.MEMSET` are not used and can be deleted
- [ ] the documentation of `FakePointer1D` talks about base and offset

### DIFF
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/TypeHandler.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/TypeHandler.java
@@ -865,7 +865,6 @@ public class TypeHandler implements ITypeHandler {
 						|| cPrimitive.getType() == CPrimitives.SCHAR);
 	}
 
-	@Override
 	public IMemoryPointer getMemoryPointer() {
 		return mMemoryPointer;
 	}
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/IMemoryAdressing.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/IMemoryAdressing.java
@@ -253,6 +253,11 @@ public interface IMemoryAdressing {
 			final RequiredMemoryModelFeatures requiredFeatures,
 			final MemoryModelDeclarationsHandler memoryModelDeclarationsHandler);
 
-	Expression getValidArray(final ILocation loc, final RequiredMemoryModelFeatures requiredMemoryModelFeatures,
+	/**
+	 * Constructs a boolean expression that, when evaluated at the end of a procedure invocation, results in
+	 * {@code true} iff no memory is allocated that was not already allocated at the beginning of the invocation.
+	 */
+	Expression constructMemoryNeutralityCheckExpr(final ILocation loc,
+			final RequiredMemoryModelFeatures requiredMemoryModelFeatures,
 			final MemoryModelDeclarationsHandler memoryModelDeclarationsHandler);
 }
--- a/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/IMemoryAdressing.java
+++ b/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/IMemoryAdressing.java
@@ -35,7 +35,6 @@ import de.uni_freiburg.informatik.ultimate.boogie.ast.AssumeStatement;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.Declaration;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.Expression;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression;
-import de.uni_freiburg.informatik.ultimate.boogie.ast.Specification;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.Statement;
 import de.uni_freiburg.informatik.ultimate.boogie.ast.VariableLHS;
 import de.uni_freiburg.informatik.ultimate.boogie.type.BoogieType;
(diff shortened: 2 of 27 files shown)


schuessf:
Thank you for your improvements! The changes you did so far look good to me.

I have just a few comments to some of the open issues:

> It would probably make sense to move a lot of this code into a dedicated sub-package, to bundle it in one place. Then, we should also see which parts need to be public and what can be package-private.

👍

>  An instance of `IMemoryPointer` is not actually a pointer. Can we rename the interface?

Yes, it is not a pointer, but it is responsible for the pointer representation. So myabe something like `IPointerRepresentation` is suitable?

>  (where) is `LocalLValueILocationPair` needed? (if we keep it, make it a record, possibly private, and improve field names)

It is used in [`MemoryHandler`](https://github.com/ultimate-pa/ultimate/blob/7f5e455baa6dd87862ea3ed5389f22f4c5147569/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/MemoryHandler.java#L158), mostly internally, but there is another usage in [`FunctionHandler`](https://github.com/ultimate-pa/ultimate/blob/7f5e455baa6dd87862ea3ed5389f22f4c5147569/trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/base/chandler/FunctionHandler.java#L627). But I think, converting it to a record is a good idea.

> The documentation for `NonBijectiveMapping1D` talks about base and offset (copy-pasted from 2D case). Is the mapping even non-bijective in the 1-dimensional case?

A pointer in the 1D case is just represented using the base (`{base: x}`) and there is obviously a bijective mapping to integers (just `x`). Therefore I am also wondering, whether we even need another option for the 1D case to convert integers.

> `MemoryAddressingFactory` violates the open-closed principle (for every new `IMemoryPointer` implementation, we would need to change this class). Can this functionality become a method on `IMemoryPointer`?

> The construction of a suitable pointer-integer conversion in `MemoryAddressingBase` violates the open-closed principle (for every new `IMemoryPointer` implementation, we would need to change this class). Can this functionality become a method on `IMemoryPointer`?

I think having these functionalities as methods in `IMemoryPointer` would improve the situation. I wonder however, if we could also have a factory that creates the whole memory model. There exists already `MemoryPointerFactory` to create `IMemoryPointer`, maybe it can be extended to create also the rest of the memory model.

> the documentation of `FakePointer1D` talks about base and offset

Maybe we can just merge `FakePointer1D` and `FakePointer2D` into a single class instead, as it is just responsible for the string representation anyway.

PR #770, opened 2026-01-27
Use `IcfgBuilder` instead of `RCFGBuilder` as default

schuessf:
Currently, Ultimate still offers two ways to build control-flow graphs:
* `RCFGBuilder`: the legacy implementation, which only works for programs without `while`/`if` and therefore requires `UnstructureCode`
* `IcfgBuilder`: the newer implementation, which supports arbitrary programs and fixed existing bugs

Maintaining both implementations is undesirable because any change requires keeping multiple (partly duplicated) files in sync. For SV-COMP 2026 we already used `IcfgBuilder` and observed almost identical results as with `RCFGBuilder`. However, this was achieved via changes in the settings and toolchains as the quickest solution. This PR applies the switch properly ~~and removes `RCFGBuilder` entirely~~. Therefore, the following changes were performed:
* Switch all settings and toolchains that previously used `RCFGBuilder` to `IcfgBuilder` (incl. those used for regression tests)
* ~~Remove the `RCFGBuilder` plugin~~ and move required files into `IcfgBuilder`
* Disable `UnstructureCode` by default (globally, not just via settings)


trunk/examples/Backtranslation/regression/c/standard/Globals_GenericKojakCInline_Generic.errorpath
```diff
@@ -1,18 +1,15 @@
 [L3]               int a = 5;
 [L4]               int x = 0;
-       VAL         [a=5, x=0]
```
schuessf, line 3: @Heizmann `IcfgBuilder` seems to behave slightly different than `RCFGBuilder` on these states in the error path. Is this expected?

trunk/examples/Backtranslation/regression/c/standard/Loops.errorpath
```diff
@@ -42,10 +42,12 @@
        VAL         [a=0, c=0, d=0, e=0, f=2]
 [L23]  COND FALSE  !(d>0)
        VAL         [a=0, c=0, e=0, f=2]
+[L29]  COND TRUE   f>0
```
schuessf, line 45: @Heizmann `IcfgBuilder` seems to behave slightly different than `RCFGBuilder` on these states in the error path. Is this expected?

schuessf:
After some discussion with @Heizmann we agreed to the following:
* We use `IcfgBuilder` as default in all tollchains (instead of `RCFGBuilder`)
* `IcfgBuilder` does not depend anymore on `RCFGBuilder`, i.e., all commonly used classes are moved to `IcfgBuilder`
* But we keep `RCFGBuilder` for now as a reference, but because of the previous point just a handfull of classes are left that can be easily removed in the future

Therefore, I just adapted the PR to reflect these changes.

Heizmann:
- I haven't had a chance to look into the questions above yet, and that's unlikely to change in the coming days.
- In the past we had many differences like this. Some of these differences pointed out bugs in the ICFG. Some of these differences are due to the fact that the ICFG builder has slightly less program points or program points at different positions and hence sometimes fewer values in counterexamples.
- If these are the only differences, then let's just ignore them and hope that there are good reasons for not showing values there.

schuessf:
Thanks for the comments!

> * In the past we had many differences like this. Some of these differences pointed out bugs in the ICFG. Some of these differences are due to the fact that the ICFG builder has slightly less program points or program points at different positions and hence sometimes fewer values in counterexamples.
> 
> * If these are the only differences, then let's just ignore them and hope that there are good reasons for not showing values there.

After taking another look at these backtranslation tests, the change to `Loops.errorpath` has nothing to do with the program states. There, a new action for the loop condition was added, which is a positive change that can be explained by the changes to `IcfgBuilder`.

The change to `Globals*.errorpath` is, however, somewhat concerning. Several states are missing: the absence of states after the initialization of global variables could be considered acceptable, but more problematic is that states are also missing after some return statements:

```
[L8]               return x;
       VAL         [\result=7, a=6, x=7]
[L13]  RET, EXPR   f(f(a++))
```

With `IcfgBuilder`, the state after the `return` statement `[\result=7, a=6, x=7]` is missing.

I just checked the produced CFG, `IcfgBuilder` also applies block encoding to assume statements carrying attributes. This means that in this example, assume statements with the `begin_inline` and `end_inline` attributes are merged into a single large block that incldues the two calls in line 13 of the C program.
@Heizmann: This differs from the behavior in `RCFGBuilder`, so is this really intended?

Heizmann:
We might need a meeting to discuss this.
It seems to me a trade-off between fewer program points and the possibility to get values.
At a first glance, I would say: If a user is too greedy to spend CPU time on our interprocedural analysis he or she voided all their rights to get procedure contracts or values at procedure exists.
But right now, I don't know if merging the inline attributes has also other negative consequences.

schuessf:
I checked again: the issue isn't related to the attributes, since the `RCFGBuilder` doesn't ensure that statements with attributes get their own block either. Instead, the issue is because of the changed handling of labels. 26cbad3 introduced an optimization that omits labels in cases where either no goto exists or the goto is in the line immediately before the label. The second case often occurs as a result of inlining, for returns at the end of a function, e.g., in `Globals.c`:
```
  f_$Ultimate##0#1 { :auxiliary_label } :
    [...]
    f_#res#1 := ~x~0;
    goto f_returnLabel#1;
  f_returnLabel#1 { :auxiliary_label } :
    [...]
```
>  It seems to me a trade-off between fewer program points and the possibility to get values.

I agree. Now we can at least see the reason behind this change...

> We might need a meeting to discuss this.

...and can discuss whether this is desired or not.

schuessf:
After the discussion today, we decided to keep the optimization that results potentially less program states in the output if inlining is used. In the future, we may introduce a setting for this.

Also, I just moved all the files of the ICFG datastructure to its own library to reduce the dependencies on `IcfgBuilder`. Some of these datastructure classes still heavily depnded on IcfgBuilder (used e.g., Activator.PLUGIN_ID or its preferences), but I managed to decouple them by introducing new parameters.

I would like to merge this PR in the next week to finally close this chapter, so if you want to review it, please do so soon.

bahnwaerter (approved):
@schuessf thank you very much for your extensive work! Based on the decisions from today's meeting, all your changes look good to me.

PR #771, opened 2026-01-27
Report `NotAnalyzedResult` instead of `UnprovableResult` when analysis stops early

schuessf:
Currently, if a specification is not analyzed (e.g., because an error was found earlier and `Stop after first violation was found` is enabled), Ultimate reports a `UnprovableResult` for the remaining specifications. This is misleading, since it can look like we failed to verify those specifications, when in fact they were never checked.

This PR introduces a dedicated `NotAnalyzedResult` for such cases (as already proposed by @danieldietsch [here](https://github.com/ultimate-pa/ultimate/pull/730#issuecomment-2996205689) 😉), which should make the results easier to understand (e.g., in the web interface). As a result, the output changes from
```
- UnprovableResult [Line: ...]: Unable to prove that a call to reach_error is unreachable
  Unable to prove that a call to reach_error is unreachable
  Reason: Not analyzed. 
```
to
```
- NotAnalyzedResult [Line: ...]: Not analyzed if a call to reach_error is unreachable
  Not analyzed if a call to reach_error is unreachable
```

As a potential follow-up (not yet done in this PR), we could also add a setting whether such results should be output at all.

### DIFF
--- a/trunk/source/Library-UltimateCore/src/de/uni_freiburg/informatik/ultimate/core/lib/results/NotAnalyzedResult.java
+++ b/trunk/source/Library-UltimateCore/src/de/uni_freiburg/informatik/ultimate/core/lib/results/NotAnalyzedResult.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2026 Frank Schüssele (schuessf@informatik.uni-freiburg.de)
+ * Copyright (C) 2026 University of Freiburg
+ *
+ * This file is part of the ULTIMATE Core.
+ *
+ * The ULTIMATE Core is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published
+ * by the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


danieldietsch (approved):

maul-esel:
:+1: good, I've been meaning to do something about that.

Beyond the confusing message, the sheer number of these results often makes it difficult to find the actually meaningful information. Would it be an option to summarize them in one result? (either now, or with a future PR that introduces a setting)

schuessf:
> Beyond the confusing message, the sheer number of these results often makes it difficult to find the actually meaningful information. Would it be an option to summarize them in one result? (either now, or with a future PR that introduces a setting)

That would likely require additional changes, since we currently support only a single location (`IResultWithLocation`). I can see some alternative options that should be easier to integrate:
* Group these into a single result without a location (e.g., "Some specifications were not analyzed"), controlled by a setting  (whether we want results for every specification)
* Use the same setting to omit `NotAnalyzedResult`s entirely
* Keep emitting them, but sort results so that `NotAnalyzedResult`s come first and the "important" results appear last

bahnwaerter (approved):

PR #772, opened 2026-02-13
Add ARM64 build support for Linux and macOS

### LINKED ISSUE #749: Is it possible to compile on arm?

bahnwaerter:
The framework update introduces build support for the 64-bit ARM (aarch64) target architecture. The Ultimate framework and its components can now be built and executed on ARM64 platforms running Linux and macOS. This feature resolves #749.

### DIFF
--- a/trunk/source/BA_FeatureDependenciesCommon/feature.xml
+++ b/trunk/source/BA_FeatureDependenciesCommon/feature.xml
@@ -4,7 +4,7 @@
       label="BA_FeatureDependenciesCommon"
       version="0.3.1"
       provider-name="Ultimate Program Analysis Team"
-      arch="x86_64">
+      arch="aarch64,x86_64">
 
    <plugin
          id="com.google.gson"
@@ -72,6 +72,12 @@
          version="0.0.0"/>
 
    <plugin
+         id="org.eclipse.cdt.core.linux.aarch64"
+         os="linux"
+         arch="aarch64"
+         version="0.0.0"/>
+
+   <plugin
          id="org.eclipse.cdt.core.linux.x86_64"
          os="linux"
          arch="x86_64"
@@ -111,6 +117,12 @@
 
    <plugin
          id="org.eclipse.core.filesystem"
+         version="0.0.0"/>
+
+   <plugin
+         id="org.eclipse.core.filesystem.linux.aarch64"
+         os="linux"
+         arch="aarch64"
          version="0.0.0"/>
 
    <plugin
--- a/trunk/source/BA_FeatureDependenciesDebugE4/feature.xml
+++ b/trunk/source/BA_FeatureDependenciesDebugE4/feature.xml
@@ -4,7 +4,7 @@
       label="BA_FeatureDependenciesDebugE4"
       version="0.3.1"
       provider-name="Ultimate Program Analysis Team"
-      arch="x86_64">
+      arch="aarch64,x86_64">
 
    <requires>
       <import feature="BA_FeatureDependenciesCommon"/>
@@ -229,6 +229,13 @@
          os="macosx"
          ws="cocoa"
          arch="x86_64"
+         version="0.0.0"/>
+
+   <plugin
+         id="org.eclipse.swt.gtk.linux.aarch64"
+         os="linux"
+         ws="gtk"
+         arch="aarch64"
          version="0.0.0"/>
 
    <plugin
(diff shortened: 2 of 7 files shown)


schuessf:
Thank you for your work! I tried running the build using `makeFresh.sh` (also on my Raspberry Pi) and it suceeded. However, I found a few issues:
* [`makeZip.sh`](https://github.com/ultimate-pa/ultimate/blob/dev/releaseScripts/default/makeZip.sh#L54) currently always copies the `x86_64` folder. Instead, the target architecture (e.g. `aarch64`) should be passed as a parameter.
* Currently, the repository only contains x86 solver binaries, which are therefore the only ones packaged. As a result, the build cannot be executed on aarch. Although aarch builds of [Z3](https://github.com/Z3Prover/z3/releases), [MathSAT](https://mathsat.fbk.eu/download.html), [bitwuzla](https://github.com/bitwuzla/bitwuzla/releases), and [CVC5](https://github.com/cvc5/cvc5/releases/) are available and could be integrated (for Z3, we could also adapt our [build script](https://github.com/ultimate-pa/ultimate/blob/dev/releaseScripts/default/makeZ3Generic.sh#L20) to build for aarch), there is no aarch version of CVC4. Maybe we can also use this as an opportunity to replace CVC4 with bitwuzla (which I already stated [here](https://github.com/ultimate-pa/ultimate/tree/wip/fs/bitwuzla2)). We mainly use CVC4 for bitvectors, and I made some experiments that showed that bitwuzla is slightly better for that use case.
* I also tried running the resulting build on my Raspberry Pi, but I always got a `ClassNotFoundException: org.eclipse.swt.SWTError cannot be found by org.eclipse.ui.workbench_3.136.0.v20250814-1558`, but I did not figure out why. Is there still some issue with the build or is there some missing requirement on my system?

bahnwaerter:
Thank you @schuessf for testing the new ARM64 support.

> [`makeZip.sh`](https://github.com/ultimate-pa/ultimate/blob/dev/releaseScripts/default/makeZip.sh#L54) currently always copies the `x86_64` folder. Instead, the target architecture (e.g. `aarch64`) should be passed as a parameter.

I haven't added support to the `makeFresh.sh` script yet. I will do this together with the new feature for automatically building Docker images.

> Currently, the repository only contains x86 solver binaries, which are therefore the only ones packaged. As a result, the build cannot be executed on aarch. [...]

That's a great idea and we should check that out!

> I also tried running the resulting build on my Raspberry Pi, but I always got a `ClassNotFoundException: org.eclipse.swt.SWTError cannot be found by org.eclipse.ui.workbench_3.136.0.v20250814-1558`, but I did not figure out why. Is there still some issue with the build or is there some missing requirement on my system?

It seems like there is a component missing on your system that is required by the Java virtual runtime in order to avoid runtime issues. Which version of Raspberry Pi OS have you used for testing?

schuessf:
> > [`makeZip.sh`](https://github.com/ultimate-pa/ultimate/blob/dev/releaseScripts/default/makeZip.sh#L54) currently always copies the `x86_64` folder. Instead, the target architecture (e.g. `aarch64`) should be passed as a parameter.
> 
> I haven't added support to the `makeFresh.sh` script yet. I will do this together with the new feature for automatically building Docker images.

👍 
Sure I was just checking what is missing to make the build enviroment fully ready for aarch -- also as a reminder 😉 

> > I also tried running the resulting build on my Raspberry Pi, but I always got a `ClassNotFoundException: org.eclipse.swt.SWTError cannot be found by org.eclipse.ui.workbench_3.136.0.v20250814-1558`, but I did not figure out why. Is there still some issue with the build or is there some missing requirement on my system?
> 
> It seems like there is a component missing on your system that is required by the Java virtual runtime in order to avoid runtime issues. Which version of Raspberry Pi OS have you used for testing?

Oh, it looks like the problem is resolved. I already installed GTK3 before, which seems to fix the issue, but it required a reboot first (which I didn't expect).
I am using Raspberry Pi OS Lite (Trixie).

schuessf:
Okay, since my issue was fixed, I was now able to actually run Ultimate on my Raspberry Pi, by performing the additional steps:
* run `makeFresh.sh`
* overwite the content of `UAutomizer-linux` with `CLI-E4/linux/gtk/aarch64`
* download the aarch solvers and pretend bitwuzla to be CVC4 😉

PR #778, opened 2026-03-24
Migrate framework to Eclipse E4.41 and update other dependencies

schuessf:
#757 previously updated the Eclipse dependencies to E4.37 and consolidated the target files. Building on that work, this update updates the Eclipse dependencies further to E4.41.

To achieve this, several unused feature groups (including some that no longer exist) have been removed from the target file, and the `feature.xml` files were cleaned up as well. Both changes should simplify future dependency updates.

Additionally, the Maven plugins, covering both the build process and regular dependencies, have been updated.

~~**Note:** This PR updates the version number of `org.eclipse.equinox.launcher` in several files within the Ultimate repo (e.g., `Ultimate.py`). However, the tool info module in benchexec still needs to be updated, as it also depends on the exact version number (see [here](https://github.com/sosy-lab/benchexec/blob/main/benchexec/tools/ultimate.py#L38)), which was also forgotten for the changes in #757. To avoid having to make such changes for every future dependency update, I already implemented a wildcard-based solution: https://github.com/ultimate-pa/benchexec/commit/88b9aa5b03b5cbcfd6c1aa5ef812811e493a618f. If you agree with that approach, I can also open a PR for benchexec.~~

### DIFF
--- a/trunk/source/BA_FeatureDependenciesDebugE4/feature.xml
+++ b/trunk/source/BA_FeatureDependenciesDebugE4/feature.xml
@@ -63,14 +63,6 @@
          version="0.0.0"/>
 
    <plugin
-         id="org.apache.xalan"
-         version="0.0.0"/>
-
-   <plugin
-         id="org.apache.xmlgraphics"
-         version="0.0.0"/>
-
-   <plugin
          id="org.eclipse.core.databinding"
          version="0.0.0"/>
 
--- a/trunk/source/BA_FeatureDependenciesDebugE4/feature.xml
+++ b/trunk/source/BA_FeatureDependenciesDebugE4/feature.xml
@@ -7,7 +7,7 @@
       arch="x86_64">
 
    <requires>
-      <import feature="BA_FeatureDependenciesCommon"/>
+      <import feature="BA_FeatureUltimateCommon"/>
    </requires>
 
    <plugin
--- a/releaseScripts/default/adds/UltDelta.py
+++ b/releaseScripts/default/adds/UltDelta.py
@@ -52,6 +52,16 @@ def check_string_contains(strings, words):
     return False
 
 
+@lru_cache(maxsize=1)
+def get_launcher_jar():
+    candidates = glob.glob(os.path.join(ultimatedir, "plugins", "org.eclipse.equinox.launcher_*.jar"))
+    if not candidates:
+        raise RuntimeError("Could not find any launcher JAR.")
+    if len(candidates) > 1:
+        raise RuntimeError("Found multiple launcher JARs: " + ", ".join(candidates))
+    return candidates[0]
+
+
 def get_binary():
     ultimate_bin = [
         "java",
@@ -65,10 +75,7 @@ def get_binary():
 
     ultimate_bin = ultimate_bin + [
         "-jar",
-        os.path.join(
-            ultimatedir,
-            "plugins/org.eclipse.equinox.launcher_1.7.100.v20251111-0406.jar",
-        ),
+        get_launcher_jar(),
         "-data",
         datadir,
     ]
--- a/releaseScripts/default/adds/UltDelta.py
+++ b/releaseScripts/default/adds/UltDelta.py
@@ -67,7 +67,7 @@ def get_binary():
         "-jar",
         os.path.join(
             ultimatedir,
-            "plugins/org.eclipse.equinox.launcher_1.7.0.v20250519-0528.jar",
+            "plugins/org.eclipse.equinox.launcher_1.7.100.v20251111-0406.jar",
         ),
         "-data",
         datadir,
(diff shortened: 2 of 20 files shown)


trunk/source/BA_FeatureDependenciesDebugE4/feature.xml
```diff
@@ -7,7 +7,7 @@
       arch="x86_64">
 
    <requires>
-      <import feature="BA_FeatureDependenciesCommon"/>
+      <import feature="BA_FeatureUltimateCommon"/>
```
danieldietsch, line 10: Why all these changes to the feature dependencies (also in other features)?

danieldietsch (commented):
Wrt. to benchexec just using any launcher: I think that is ok, but I would not take the highest launcher if it finds multiple ones, but rather crash. It might be that someone copied a newer Ultimate over an older one, and by some miracle it launches. I would prefer then that the wrapper script crashes and tells us that there are multiple launchers.

trunk/source/BA_FeatureDependenciesDebugE4/feature.xml
schuessf: @bahnwaerter changed this in b5bba479eaab4618b11b2d2aecfa5ff52cae81a2, but I don't quite remember why 😉

schuessf:
> Wrt. to benchexec just using any launcher: I think that is ok, but I would not take the highest launcher if it finds multiple ones, but rather crash. It might be that someone copied a newer Ultimate over an older one, and by some miracle it launches.

Sure, I changed it like that (https://github.com/ultimate-pa/benchexec/commit/43634d70be0ede073112eb2b8f463657f3369e70).

> I would prefer then that the wrapper script crashes and tells us that there are multiple launchers.

Currently, the wrapper script simply runs the specified version of `org.eclipse.equinox.launcher` without looking for other potentially present versions. I ~~could also adapt~~ adapted it to use wildcards (similar to the change in benchexec), but the version number would still need to be updated in several other files (incl. `WebBackend.ini`, where a wildcard-based solution is not possible).

schuessf:
For reference, I just opened a benchexec PR already: https://github.com/sosy-lab/benchexec/pull/1263
(this allows us to use the current version of Ultimate, and also for this PR)

schuessf:
I just tried to upgrade to Eclipse 2026-09 (E4.41), which was just released. If everything works as expected, I will merge this PR soon.

trunk/source/BA_FeatureDependenciesDebugE4/feature.xml
bahnwaerter: I've simplified the feature imports. Transitive dependencies are now supported. This means that an Ultimate product primarily depends on its Ultimate features (e.g., Ultimate Libraries). These, in turn, may depend on further external dependencies. These should then be imported transitively to keep the imports modular. Explicitly listing all transitively dependent imports is no longer necessary.

bahnwaerter:
The migration to Eclipse 2026-09 (E4.41) should go smoothly, as this release is still based on Java 21. Therefore, there is no need to adjust any configurations of the Eclipse projects (e.g. classpaths).

schuessf:
The nightly tests looked fine. As @bahnwaerter already suggested, on `dev` there are some regression tests writing to standard error (`ClassNotFoundException: org.glassfish.jaxb.runtime.v2.JAXBContextFactory cannot be found by com.sun.xml.bind.jaxb-osgi.source_4.0.5`, see e.g., [here](https://jenkins.sopranium.de/job/Ultimate/job/Ultimate%20Nightly/job/dev/1777/testReport/de.uni_freiburg.informatik.ultimate.regressiontest.generic/ChcRegressionTestSuite/I_smtlib_horn_regression_simple_test_1traceOfLength4_sat_smt2_S_smtlib_horn_regression_simple_ChcSolver_Golem_epf_T_smtlib_horn_regression_simple_ChcSolver_xml/)) and this is no more the case for this PR (see [here](https://jenkins.sopranium.de/job/Ultimate/job/Ultimate%20Nightly/job/PR-778/1/testReport/de.uni_freiburg.informatik.ultimate.regressiontest.generic/ChcRegressionTestSuite/I_smtlib_horn_regression_simple_test_1traceOfLength4_sat_smt2_S_smtlib_horn_regression_simple_ChcSolver_Golem_epf_T_smtlib_horn_regression_simple_ChcSolver_xml/)). I just rebased the PR again, and I will merge it after another nightly.

bahnwaerter:
That's great to hear that this JAXB exception no longer occurs. The JAXB 4.0.5 package is indeed incorrectly packaged. It is incorrectly referencing the source package (even though this should only be optional). It's great that this problem has now been fixed.

bahnwaerter (approved):

PR #783, opened 2026-05-29
Add Owicki-Gries annotations

maul-esel:
This PR adds Owicki-Gries annotations as proofs for concurrent programs, as well as algorithms to compute such annotations after verification.

---

:construction: **This is work in progress** :construction:

---

Specifically, we add:

- basic classes for describing and validating Owicki-Gries annotations.
- 2 algorithms for the construction of Owicki-Gries annotations:
  - a _naive_ algorithm (which creates very large annotations, but can be useful as a baseline for comparison)
  - the _empire automaton_-based algorithm described in our [POPL'26 paper](https://dl.acm.org/doi/10.1145/3776684)

  We previously implemented some other algorithm variants, but as they have been superseeded by the POPL'26 algorithm, they are no longer part of this PR. With one exception:
- a partial implementation of a refined algorithm based on _directed empires_: while not yet complete, this may be a promising base for future improvements to our Owicki-Gries computation.
- changes in backtranslation and correctness witness (v2.1) generation that allow us to output correctness witnesses for concurrent programs from Owicki-Gries proofs.


trunk/examples/settings/default/gemcutter/svcomp-DataRace-32bit-GemCutter_Bitvector.epf
```diff
@@ -61,7 +61,6 @@ file_export_version=3.0
 /instance/de.uni_freiburg.informatik.ultimate.plugins.generator.traceabstraction/Logic\ for\ external\ solver=ALL
 /instance/de.uni_freiburg.informatik.ultimate.plugins.generator.traceabstraction/SMT\ solver=External_ModelsAndUnsatCoreMode
 /instance/de.uni_freiburg.informatik.ultimate.plugins.generator.traceabstraction/Command\ for\ external\ solver=cvc4 --incremental --print-success --lang smt
-/instance/de.uni_freiburg.informatik.ultimate.plugins.generator.traceabstraction/Positions\ where\ we\ compute\ the\ Hoare\ Annotation=LoopHeads
```
schuessf, line 64: Why were these settings for GemCutter changed? GemCutter should not be affected by the Owicki-Gries annotations (at least for now), right?

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/Condition.java
```diff
@@ -47,7 +47,7 @@ public class Condition<LETTER, PLACE> implements Serializable {
 	private static final long serialVersionUID = -497620137647502376L;
 
 	// See https://github.com/ultimate-pa/ultimate/pull/595 for discussion
-	private static final boolean USE_HASH_JENKINS = false;
+	private static final boolean USE_HASH_JENKINS = true;
```
schuessf, line 50: Why was this changed?

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
```diff
@@ -872,6 +879,26 @@ private BacktranslatedExpression translateArrayAccess(
...
+	public BacktranslatedExpression declareAndTranslateAuxiliaryVariable(
+			final de.uni_freiburg.informatik.ultimate.boogie.ast.IdentifierExpression variable) {
+		final String identifier = variable.getIdentifier();
+		if (mAuxVariables.containsKey(identifier)) {
+			throw new UnsupportedOperationException("Auxiliary variable already declared");
+		}
+
+		final String ghostPrefix = "~ghost~";
+		final String newIdentifier;
+		if (Pattern.matches(ghostPrefix + "([a-zA-Z0-9]*)", identifier)) {
```
schuessf: Is there any reason why the backtranslated variable has a different naming scheme, based on the name of the Boogie variable?

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/EmpireAutomaton.java
```diff
@@ -0,0 +1,324 @@
...
+	private Set<Region<P>> replaceRegions(final Transition<L, P> transition,
+			final Set<Region<P>> replacementBystanders) {
+		final var regions = new HashSet<>(replacementBystanders);
+		for (final var succ : transition.getSuccessors()) {
+			regions.add(Region.singleton(succ));
+		}
+		return regions;
+	}
+
+	// TODO use ImmutableSet for bystanders
```
schuessf: Bystanders are not part of the state anymore, so I guess this TODO is outdated.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/SaturatedEmpire.java
```diff
@@ -0,0 +1,324 @@
...
+	// TODO use ImmutableSet for bystanders
+	public record State<L, P>(Territory<P, Region<P>> territory, IPredicate law, int hash) {
+		// Convenience constructor that computes the correct hash code. Always use this constructor.
+		public State(final Territory<P, Region<P>> territory, final IPredicate law) {
+			this(territory, law, Objects.hash(territory, law));
+		}
+
+		@Override
+		public int hashCode() {
+			// Hash code is cached for performance.
```
schuessf, line 289: And therefore, is it still beneficial to cache the hashcode? For the `Territory`  it should be already cached via `Region`.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/naive/OwickiGriesConstruction.java
```diff
@@ -0,0 +1,351 @@
...
+ *
+ * @author Dominik Klumpp (klumpp@informatik.uni-freiburg.de)
+ * @author Miriam Lagunes (miriam.lagunes@students.uni-freiburg.de)
+ *
+ * @param <P>
+ *            The type of places in the Petri program
+ * @param <L>
+ *            The type of statements in the Petri program
+ */
+public class OwickiGriesConstruction<L, P> {
```
schuessf: Maybe this class can be renamed to sth. like `NaiveOwickiGriesConstruction`, as it is just specific to the naive approach and not a general construction approach.

trunk/source/Library-TraceCheckerUtilsTest/META-INF/MANIFEST.MF
```diff
@@ -9,8 +9,11 @@ Bundle-RequiredExecutionEnvironment: JavaSE-21
 Require-Bundle: de.uni_freiburg.informatik.ultimate.lib.core,
  de.uni_freiburg.informatik.ultimate.lib.test,
  org.junit,
+ org.yaml.snakeyaml,
```
schuessf: Why are there new dependencies in this plugin?

trunk/source/Library-UltimateUtil/src/de/uni_freiburg/informatik/ultimate/util/datastructures/DataStructureUtils.java
```diff
@@ -112,12 +112,19 @@ public static <T> Optional<T> getSomeCommonElement(final Set<T> set1, final Set<
...
+	 * The returned set is not guaranteed to be mutable.
+	 *
+	 * Future changes to the sets a and b will not be reflected in the returned set.
 	 */
 	public static <T> Set<T> difference(final Set<T> a, final Set<T> b) {
 		if (a.isEmpty()) {
 			return Collections.emptySet();
 		}
 		if (b.isEmpty()) {
+			if (a instanceof ImmutableSet<?>) {
```
schuessf, line 125: I guess this branch was added to improve the performance by caching the hashcode.
Maybe another comment to this: Would it make sense for this method to always return an `ImmutableSet`?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/TraceAbstractionStarter.java
```diff
@@ -119,6 +132,10 @@ public enum CegarRestartBehaviour {
 	private final CegarLoopResultReporter<L> mResultReporter;
 	private final IWitnessTransformer<L> mWitnessTransformer;
 
+	// TODO #proofRefactor This is only supposed to be a temporary workaround.
```
schuessf, line 123: What would be the nice solution here? Also, it looks like `mUnpetrifyVariable` is not even used.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/TraceAbstractionStarter.java
```diff
@@ -201,6 +217,83 @@ private void runCegarLoops(final IIcfg<IcfgLocation> icfg) {
 					FloydHoareUtils.createProcedureContractResults(mServices, Activator.PLUGIN_NAME, icfg, annotation,
 							backTranslatorService, mResultReporter::reportResult);
 				}
+			} else if (proof instanceof OwickiGriesAnnotation<?, ?, ?>) {
```
schuessf, line 208: Maybe move this code to a method

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/DirectedLegalFocusOwickiGries.java
```diff
@@ -0,0 +1,261 @@
...
+import de.uni_freiburg.informatik.ultimate.lib.proofs.owickigries.empire.directed.DirectedEmpireProduct.ProductState;
+import de.uni_freiburg.informatik.ultimate.lib.proofs.owickigries.empire.directed.DirectedLegalEmpireToOG;
+import de.uni_freiburg.informatik.ultimate.lib.proofs.owickigries.empire.directed.DirectedLegalFocus;
+import de.uni_freiburg.informatik.ultimate.lib.smtlibutils.IncrementalPlicationChecker.Validity;
+import de.uni_freiburg.informatik.ultimate.lib.smtlibutils.ManagedScript;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.DataStructureUtils;
+import de.uni_freiburg.informatik.ultimate.util.statistics.IStatisticsDataProvider;
+import de.uni_freiburg.informatik.ultimate.util.statistics.TimeTracker;
+
+public class DirectedLegalFocusOwickiGries<L extends IAction, P> implements IPetriNetProofProducer<L, P> {
```
schuessf, line 69: I don't quite remember the status of this implemented approach. Does it work and is it sound? If not, please mark it somehow.

schuessf (commented):
Thank you for preparing this PR! Overall this looks good.
I just quickly passed over all these changes and made some comments.

trunk/examples/settings/default/gemcutter/svcomp-DataRace-32bit-GemCutter_Bitvector.epf
maul-esel: The intention here was exactly to avoid any effect on GemCutter. Previously this setting was completely ignored for concurrent programs; now we sometimes have to compute annotations.

So generally I think it's best to remove these settings from GemCutter files, because they don't make sense anyway. But let's leave this discussion open until I have re-checked how exactly the setting is handled now :+1:

trunk/source/Library-Automata/src/de/uni_freiburg/informatik/ultimate/automata/petrinet/unfolding/Condition.java
maul-esel: :+1: I have to investigate this, and possibly revert it.

trunk/source/Library-TraceCheckerUtilsTest/META-INF/MANIFEST.MF
maul-esel: Good catch; this is a leftover from when we tested O/G proof construction in this project. We have since added a separate `Library-ProofsTest` project (but even that no longer needs `snakeyaml`).

trunk/source/Library-UltimateUtil/src/de/uni_freiburg/informatik/ultimate/util/datastructures/DataStructureUtils.java
maul-esel: The point is not only to cache the hashcode, but also avoid copying the set (which takes time and memory).

Always returning an immutable set:
- It is difficult to get an overview over all the usages throughout Ultimate and be certain nothing would be broken.
- It is not clear that creating a fresh immutable set would always be a good idea, as they do come with some memory overhead.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/concurrency/CegarLoopForPetriNet.java
```diff
@@ -106,7 +111,12 @@ public enum SizeReduction {
 		REMOVE_DEAD, REMOVE_REDUNDANT_FLOW
 	}
 
-	private static final boolean USE_ON_DEMAND_RESULT = false;
+	private static final boolean DUMP_OWICKI_GRIES_TEST = false;
+	private final List<INwaOutgoingLetterAndTransitionProvider<L, IPredicate>> mProofAutomata = new ArrayList<>();
+	private final IPetriNet<L, IPredicate> mInitialNet;
+	private HashRelation<IPredicate, Transition<L, IPredicate>> mPossibleInterferences;
+
+	private static final boolean USE_ON_DEMAND_RESULT = true;
```
maul-esel, line 119: We need to be careful about this change, it affects verification performance.
- Do our algorithms still rely on changing this flag? This was probably only relevant for the unfolding-based approaches (but let's check).
- If we do rely on it, we either need to adapt our implementation, or at least only modify this flag if proof production is enabled.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/cdt/translation/implementation/CLocation.java
```diff
@@ -223,8 +223,14 @@ private static boolean hasParent(final IASTNode node, final Set<IASTNode> otherN
...
-		return getUniqueElementOrNull(
-				mNodes.stream().map(x -> CdtASTUtils.findScope(x).getDeclarator().getName().toString()));
+		return getUniqueElementOrNull(mNodes.stream().map(x -> {
+			final var scope = CdtASTUtils.findScope(x);
+			if (scope == null) {
+				// global scope
+				return null;
+			}
+			return scope.getDeclarator().getName().toString();
+		}));
```
maul-esel, line 233: This was a hack to get witness production working. I vaguely recall that I either fixed the NPE that occurred here in another way, or at least realised it _should_ be fixed in another way ;-) I think this was related to invariants / ghost updates at global variable initializations (which we probably want to avoid).

Let's check this, and if it's not strictly needed, re-assess whether this change is a good idea or not.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/witness/ExtractedGhostUpdate.java
```diff
@@ -157,6 +158,23 @@ private static List<Statement> annotateFork(final ILocation loc, final List<Stat
...
+				result.add(StatementFactory.constructAtomicStatement(loc, ghostUpdate));
+			}
+			result.add(st);
+		}
+		if (!isAnnotated) {
+			throw new UnsupportedOperationException("No statement found to annotate with the expected ghost update");
+		}
+		return result;
+	}
+
```
maul-esel, line 177: I think this duplicates some logic from `annotateFork`; see if they can be combined.

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
maul-esel: The idea was to avoid that every backtranslator adds yet another `ghost`-prefix, and to preserve meaningful names where possible, to make the proofs easier to understand.

I should at least move that string constant to `SFO` or some other static declaration though.

trunk/source/Library-ModelCheckerUtils/src/de/uni_freiburg/informatik/ultimate/lib/modelcheckerutils/boogie/Term2Expression.java
```diff
@@ -643,6 +643,22 @@ private static Expression xor(final Expression[] params) {
...
+		if (pv != null) {
+			throw new UnsupportedOperationException("Cannot redeclare program variable as auxiliary: " + pv);
+		}
+		final IBoogieType type = mTypeSortTranslator.getType(variable.getSort());
+		final String newIdentifier = "~ghost~" + mAuxiliaryVariables.size();
+		final var id = new IdentifierExpression(null, type, newIdentifier,
+				new DeclarationInformation(StorageClass.QUANTIFIED, null));
+		mAuxiliaryVariables.put(variable, id);
+		return id;
+	}
```
maul-esel, line 661: Should we here also try to preserve meaningful ghost variable names?

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/EmpireStatistics.java
```diff
@@ -0,0 +1,91 @@
...
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.IPredicate;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.PredicateWithConjuncts;
+import de.uni_freiburg.informatik.ultimate.smtinterpol.util.DAGSize;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.Pair;
+import de.uni_freiburg.informatik.ultimate.util.statistics.AbstractStatisticsDataProvider;
+import de.uni_freiburg.informatik.ultimate.util.statistics.MinMaxMed;
+
+public class EmpireAutomataStatistics extends AbstractStatisticsDataProvider {
+	public static final String AUTOMATON_SIZE = "automaton size";
+	public static final String UNIQUE_PAIRS = "number of unique pairs";
```
maul-esel, line 41: This statistic is likely obsolete.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/EmpireAutomaton.java
```diff
@@ -0,0 +1,323 @@
...
+import de.uni_freiburg.informatik.ultimate.automata.petrinet.IPetriNet;
+import de.uni_freiburg.informatik.ultimate.automata.petrinet.netdatastructures.Transition;
+import de.uni_freiburg.informatik.ultimate.core.model.services.ILogger;
+import de.uni_freiburg.informatik.ultimate.core.model.services.IUltimateServiceProvider;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.IPredicate;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.DataStructureUtils;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.ImmutableSet;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.Pair;
+
+// TODO Give this class a more descriptive name
```
maul-esel: ... and document the class

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/SaturatedEmpire.java
```diff
@@ -0,0 +1,323 @@
...
+	public boolean isInitial(final State<L, P> state) {
+		return mInitialState.equals(state);
+	}
+
+	/**
+	 * Determines a state s as final, if it contains an error place and the law is false, OR if there exists at least
+	 * one transition in enabled(territory(s)), for which there is no successor in the automaton. In this case, the
+	 * successor law must be false.
+	 */
+	public boolean isFinal2(final State<L, P> state) {
```
maul-esel: Re-examine the usages of final states in this class (and the corresponding interfaces). Delete redundant methods.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/EmpireToOG.java
```diff
@@ -0,0 +1,264 @@
...
+import de.uni_freiburg.informatik.ultimate.lib.smtlibutils.SmtSortUtils;
+import de.uni_freiburg.informatik.ultimate.lib.smtlibutils.SmtUtils;
+import de.uni_freiburg.informatik.ultimate.logic.SMTLIBConstants;
+import de.uni_freiburg.informatik.ultimate.logic.Script;
+import de.uni_freiburg.informatik.ultimate.logic.Term;
+import de.uni_freiburg.informatik.ultimate.logic.TermVariable;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.DataStructureUtils;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.Pair;
+
+public class EmpireToOG<S, L, P> {
```
maul-esel: Document and rename

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/IEmpireAutomaton.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.List;
+
+import de.uni_freiburg.informatik.ultimate.automata.nestedword.INwaOutgoingLetterAndTransitionProvider;
+import de.uni_freiburg.informatik.ultimate.automata.nestedword.transitions.OutgoingCallTransition;
+import de.uni_freiburg.informatik.ultimate.automata.nestedword.transitions.OutgoingReturnTransition;
+import de.uni_freiburg.informatik.ultimate.automata.petrinet.netdatastructures.Transition;
+import de.uni_freiburg.informatik.ultimate.automata.statefactory.IStateFactory;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.IPredicate;
+
+// TODO Possibly rename to IEmpire
```
maul-esel: ... and document

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/ILegalFocusFunction.java
```diff
@@ -0,0 +1,82 @@
...
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.PredicateWithConjuncts;
+
+/**
+ * Abstract interface for legal focus.
+ *
+ * A legal focus assigns potentially different laws to different regions in the territory belonging to an empire
+ * automaton. For each region, the assigned law must be weaker than the state's full law, and satisfy additional
+ * conditions.
+ *
+ * TODO Document additional conditions
```
maul-esel: (check in the paper)

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/LegalFocus.java
```diff
@@ -0,0 +1,283 @@
...
+import de.uni_freiburg.informatik.ultimate.automata.petrinet.IPetriNet;
+import de.uni_freiburg.informatik.ultimate.automata.petrinet.netdatastructures.Transition;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.IPredicate;
+import de.uni_freiburg.informatik.ultimate.lib.modelcheckerutils.smt.predicates.PredicateWithConjuncts;
+import de.uni_freiburg.informatik.ultimate.lib.smtlibutils.SmtUtils;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.DataStructureUtils;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.HashRelation;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.relation.Pair;
+
+public class LegalFocus<S, L, P> implements ILegalFocusFunction<S, P> {
```
maul-esel, line 70: document this class

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/naive/NaiveOwickiGriesConstruction.java
```diff
@@ -0,0 +1,351 @@
...
+		mManagedScript = mgdScript;
+		mScript = mManagedScript.getScript();
+
+		mNet = net;
+		mReachableMarkings = reachableMarkings;
+		mFloydHoareAnnotation = floydHoare;
+		mSymbolTable = new DefaultIcfgSymbolTable(symbolTable, procedures);
+
+		// TODO let callers pass predicate factory
+		mFactory = new BasicPredicateFactory(services, mManagedScript, mSymbolTable);
```
maul-esel, line 109: Is this TODO still relevant?

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/EmpireAutomaton.java
matthiaszumkeller: Maybe we can rename it to _Empire_ if we call the interface _IEmpire_?

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/SaturatedEmpire.java
matthiaszumkeller: After re-examination I did not find any usage of the final method of the EmpireAutomaton class (there are some relevant calls to the isFinal method of the InterpolantAutomaton though).

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/DirectedLegalFocusOwickiGries.java
matthiaszumkeller: This method belongs to our first attempts of a "geographical focus" and is not yet sound. In some cases the resulting OG is valid though. The approach is mainly kept for documentation (I don't think we have any other  notes on the idea behind the approach).

schuessf: Thanks for the clarification! I don't want to remove this approach, we may continue the developement of this approach. However, we should at least state in some comment that it is unsound, ~~potentially even remove the setting to enable it in `OwickiGriesSettings`~~.

EDIT: Okay, we already throw an exception for `DIR_LEGAL_FOCUS` in `PetriInitialAbstractionProvider::getProofProducer`, so this seems alright.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/IExplicitEmpire.java
```diff
@@ -0,0 +1,139 @@
...
+public interface IExplicitEmpireAutomaton<L, P, S>
+		extends IEmpireAutomaton<L, P, S>, INestedWordAutomaton<Transition<L, P>, S> {
+
+	/**
+	 * @deprecated We should not abuse the final states for empires, they do not represent any meaningful language.
+	 *             Instead introduce a suitably-named new method.
+	 */
+	@Override
+	@Deprecated
+	Collection<S> getFinalStates();
```
schuessf: This method is marked as deprecated, but it is used in `LegalFocus::computeLegalFocus`. Is this reason for deprecation (*"We should not abuse the final states for empires, they do not represent any meaningful language. Instead introduce a suitably-named new method."*) still valid, or should we remove it?

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/SaturatedEmpire.java
matthiaszumkeller: However as @schuessf pointed out [here](https://github.com/ultimate-pa/ultimate/pull/783#discussion_r3332725422), there is still implicit usage of the isFinal method for states where we pruned transitions to a state with law=false. This is required to find entry points for the legal focus computation! What would be a nice name for such a method?

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/EmpireAutomaton.java
maul-esel: That seems like a _less_ descriptive name to me :)

trunk/source/CACSL2BoogieTranslator/src/de/uni_freiburg/informatik/ultimate/plugins/generator/cacsl2boogietranslator/Boogie2ACSL.java
maul-esel: :heavy_check_mark: done

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/naive/OwickiGriesConstruction.java
maul-esel: :heavy_check_mark: done

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/TraceAbstractionStarter.java
maul-esel: :heavy_check_mark: done

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/EmpireToOG.java
maul-esel: :heavy_check_mark: done

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/EmpireStatistics.java
maul-esel: I thought this was specific to one of the removed implementations. But it is actually more general, and serves as an indicator of how much history information (_beyond_ the current control locations and the current predicate) are needed for the proof. So let's keep it.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/EmpireAutomaton.java
matthiaszumkeller: I think it would at least resolve any confusion, whether this class implements the empire as presented in the paper. But maybe something more descriptive would be _EmpireTransitionFunctionProvider_ or _EmpireTransitionProvider_?

maul-esel: I am ok with getting rid of the `Automaton` part (though I do not feel strongly about it).

But in regard to making the name more descriptive, I meant more of an indication of _which_ empire is computed. If I see an interface `IEmpire` and an implementation `Empire`, I would always ask why two types are needed for the same concept.

I now renamed the interface `IEmpire`, and this implementation `SaturatedEmpire` (as in the paper).

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/directed/DirectedEmpire.java
matthiaszumkeller, line 1: After inspection of this class and as far as I can recall, this class was a nearly 1-to-1 copy from the original (_SaturatedEmpire_) class at the time, which was only added so that we a able to quickly check test the approach. However if we would also parametrize the empire and _State_ record with the type of region, with some engeneering effort most of this class (maybe even the whole class) should be obsolete (besides the method _extendAll_ ).

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/directed/DirectedEmpireProduct.java
```diff
@@ -0,0 +1,156 @@
...
+import de.uni_freiburg.informatik.ultimate.automata.petrinet.IPetriNet;
+import de.uni_freiburg.informatik.ultimate.automata.petrinet.netdatastructures.Transition;
+import de.uni_freiburg.informatik.ultimate.core.model.services.IUltimateServiceProvider;
+import de.uni_freiburg.informatik.ultimate.lib.proofs.owickigries.empire.Region;
+import de.uni_freiburg.informatik.ultimate.lib.proofs.owickigries.empire.Territory;
+import de.uni_freiburg.informatik.ultimate.lib.proofs.owickigries.empire.directed.DirectedEmpire.State;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.DataStructureUtils;
+import de.uni_freiburg.informatik.ultimate.util.datastructures.ImmutableSet;
+
+public class DirectedEmpireProduct<L, P> {
```
matthiaszumkeller, line 53: Maybe this class should also just extend the _IEmpire_ interface? The structure is practically the same  and the empire product should again be an empire (I don't think we need to change this right now, but maybe a note for the future).

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/directed/DirectedEmpireProduct.java
```diff
@@ -0,0 +1,156 @@
...
+		final var interRegions = new HashSet<Region<P>>();
+		for (final P place : places) {
+			final var regions = territories.stream().map(t -> t.getPlaceRegion(place)).collect(Collectors.toSet());
+			final var inter = ConnectedRegion.intersectConnectedRegions(regions, place);
+			interRegions.add(inter);
+		}
+		return interRegions;
+	}
+
+	private INestedWordAutomaton<Transition<L, P>, ProductState<L, P>> constructProductAutomaton() {
```
matthiaszumkeller, line 92: If this approach gets resumed in the future, I think this could also be done on-the-fly.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/LegalFocus.java
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	public interface IFocusedRegionHeuristic<P> {
+		Comparator<Pair<Region<P>, IPredicate>> getPreference();
+
+		default Comparator<Region<P>> getPreference(final IPredicate law) {
+			final var comparator = getPreference();
+			return (r1, r2) -> comparator.compare(new Pair<>(r1, law), new Pair<>(r2, law));
+		}
+
```
matthiaszumkeller, line 353: Should we add some documentation here? As far as I remember, the heuristics for choosing the right region for the focus also left a lot of space for future work.

matthiaszumkeller (commented):
Thank you for preparing this merge. I added some additional comments.

maul-esel:
In another project building of this work (branch `wip/dk/civlized-og`) we've just discovered a bug (or multiple bugs) that should be investigated and fixed before this is merged:
- [ ] With the attached file `fork-join-01.bpl`, the generated **O/G proof seems to be insufficient**: the invariant between `fork` and `join` does not contain any information about `x`, and thus the proof should be insufficient to show the assert.
  - In the project, we ask a deductive verifier to confirm the proof; and indeed it fails to do so because of this issue.
- [ ] Yet, no asserts fail. Do we not internally validate the proof when `-ea` is passed? Or is our **validator defect?**
- [ ] Also, inserting some blank lines in the program (`fork-join-01.backup-bpl`) seems to "fix" this problem: we get a stronger invariant, and the proof is confirmed by the external validator.

  Apparently, there is a **nondeterminism bug** (likely, inserting blank lines changes the line numbers of some actions, which changes their hashcode, and affects some hashset iteration order).

[fork-join01.backup.bpl.txt](https://github.com/user-attachments/files/28728155/fork-join01.backup.bpl.txt)
[fork-join01.bpl.txt](https://github.com/user-attachments/files/28728157/fork-join01.bpl.txt)
[settings.epf.txt](https://github.com/user-attachments/files/28728168/settings.epf.txt)
[toolchain.xml](https://github.com/user-attachments/files/28728169/toolchain.xml)

schuessf:
> With the attached file `fork-join-01.bpl`, the generated **O/G proof seems to be insufficient**: the invariant between `fork` and `join` does not contain any information about `x`, and thus the proof should be insufficient to show the assert.

Is it really necessary to have an invariant stronger than `true` between `fork` and `join`? In this example, we already have `x == 1` as an invariant at the exit location of the thread, so I would expect this to be sufficient to show the assert in the main thread.

> In the project, we ask a deductive verifier to confirm the proof; and indeed it fails to do so because of this issue.

Maybe the validator (or your modelling) has a different understanding of the join that allows more behavior?

> Also, inserting some blank lines in the program (`fork-join-01.backup-bpl`) seems to "fix" this problem: we get a stronger invariant, and the proof is confirmed by the external validator.
> Apparently, there is a **nondeterminism bug** (likely, inserting blank lines changes the line numbers of some actions, which changes their hashcode, and affects some hashset iteration order).

Still, we should investigate this nondeterministic behavior!

matthiaszumkeller:
> * [ ]  With the attached file `fork-join-01.bpl`, the generated **O/G proof seems to be insufficient**: the invariant between `fork` and `join` does not contain any information about `x`, and thus the proof should be insufficient to show the assert.
>   
>   * In the project, we ask a deductive verifier to confirm the proof; and indeed it fails to do so because of this issue.
> * [ ]  Yet, no asserts fail. Do we not internally validate the proof when `-ea` is passed? Or is our **validator defect?**

We do check validity of the OG-proof internally and at least to me, the resulting proof seems valid **wrt. the Petri net representation** of the program. It is true, that no information about variable `x` is contained in the invariant between `fork` and `join` . However, for the transition representing the `join` in the Petri net, there are **three** predecessor places: `{6#L13true,4#threadEXITtrue,threadThread1of1ForFork0InUse}`. Two of them do not contain any information about `x` in their invariants, they ensure that the ghostvariable is set to the value `3` associated with the exit of  the thread. However the third place (`threadThread1of1ForFork0InUse`) does contain `x`in its invariant, more specifically if `ghost==3`, then it ensures that the formula `x==1` holds. Therefore, the transition corresponding to the `join` seems to satisfy inductivity and the postcondition contains the invariant `x==1`. Am I missing something, or are there any other validity concerns that I did not consider?

> * [ ]  Also, inserting some blank lines in the program (`fork-join-01.backup-bpl`) seems to "fix" this problem: we get a stronger invariant, and the proof is confirmed by the external validator.
>   Apparently, there is a **nondeterminism bug** (likely, inserting blank lines changes the line numbers of some actions, which changes their hashcode, and affects some hashset iteration order).

I also think, that the non-deterministic behavior should be investigated.

Update:
I again looked at the output of the Ultimate toolchain for the program and it seems to be the case that the invariant of `threadThread1of1ForFork0InUse` is missing in the result after backtranslation. Maybe this could cause the issue?

schuessf:
> However, for the transition representing the `join` in the Petri net, there are **three** predecessor places: `{6#L13true,4#threadEXITtrue,threadThread1of1ForFork0InUse}`. Two of them do not contain any information about `x` in their invariants, they ensure that the ghostvariable is set to the value `3` associated with the exit of the thread. However the third place (`threadThread1of1ForFork0InUse`) does contain `x`in its invariant, more specifically if `ghost==3`, then it ensures that the invariant `x==1` holds. Therefore, the transition corresponding to the `join` seems to satisfy inductivity and the postcondition contains the invariant `x==1`. Am I missing something, or are there any other validity concerns?

Oh, I see. So I guess we do not output the invariant for `threadThread1of1ForFork0InUse` if we consider the Boogie-program, since this location is just an auxiliary-location that does not belong to any location in the Boogie program. As a result, the annotation is valid for the Petri program but not for the Boogie program. I suppose we somehow still need to consider invariants at auxiliary locations during backtranslation.

This also seems to be the reason for non-determinism, it depends whether we put the invariant at `threadThread1of1ForFork0InUse` or `6#L13true`.

matthiaszumkeller:
> This also seems to be the reason for non-determinism, it depends whether we put the invariant at `threadThread1of1ForFork0InUse` or `6#L13true`.

Nice catch! I think that I found the location where the non-determinism occurs. In the legal focus, method `chooseBestRegion` chooses the smallest region (currently our only heuristics) that contains predecessor places of the transition to be focused. In this example, it can choose between those two regions `[threadThread1of1ForFork0InUse], [6#L13true]` (the third one contains two elements) and the regions are given as a set. If it is very inconvenient to also include the auxilliary places in the resulting backtranslation, we could maybe also filter regions with auxilliary places (that are predecessors of the transition) / adjust the comparator to choose another region if possible.

schuessf:
> Nice catch! I think that I found the location where the non-determinism occurs. In the legal focus, method `chooseBestRegion` chooses the smallest region (currently our only heuristics) that contains predecessor places of the transition to be focused. In this example, it can choose between those two regions `[threadThread1of1ForFork0InUse], [6#L13true]` (the third one contains two elements) and the regions are given as a set. If it is very inconvenient to also include the auxilliary places in the resulting backtranslation, we could maybe also filter regions with auxilliary places (that are predecessors of the transition) / adjust the comparator to choose another region if possible.

I guess we could check for the auxiliary places in the legal focus as a heuristics (even if it is not quite nice). But I am still wondering whether the loss of precision when omitting the invariants for auxiliary places a) only occurs in combination with the legal focus and b) could be always fixed with such a simple heuristics (which I am quite skeptical).

maul-esel:
Thanks to both of you for investigating this! This explains why we could get a seemingly invalid annotation after backtranslation but not trigger an error before.

Let's maybe continue the discussion how to fix this in another channel to not overload this PR.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/SaturatedEmpire.java
maul-esel: As this was only used by the `LegalFocus`, I have now (in 13f01cf) moved this computation to `LegalFocus` itself.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/IExplicitEmpire.java
maul-esel: See [here](https://github.com/ultimate-pa/ultimate/pull/783#discussion_r3766234192)

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/directed/DirectedEmpire.java
maul-esel: :+1: I've added a note to the code, in case we resume this in the future. Same for your next two comments.

trunk/source/Library-Proofs/src/de/uni_freiburg/informatik/ultimate/lib/proofs/owickigries/empire/LegalFocus.java
maul-esel: I have documented the heuristics a bit.

PR #787, opened 2026-06-19
Refine settings for improved dumping of automata to files

bahnwaerter:
This change refines the existing automata dumping settings to enable more detailed and configurable automata dump options. The enhanced settings align with the `WitnessPrinter` settings and simplify the dump configuration in BenchExec runs, particularly for reuse interpolant automata (incremental verification). The change also introduces a standardized naming convention for the dumped files (using `TaskIdentifier` for proper filename patterns).


trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
```diff
@@ -624,23 +634,60 @@ private CegarLoopResult<L> performLimitReachedActions(final IRunningTaskStackPro
...
-				"", automaton);
+				determineAutomatonName(automaton), fullAutomataPath, mDumpAutomataFormat, "", automaton);
 		mCegarLoopBenchmark.stop(CegarLoopStatisticsDefinitions.DumpTime);
 	}
 
-	protected void writeAutomataToFile(final String filename, final String atsHeaderMessage, final String atsCommands,
-			final NamedAutomaton<L, IPredicate>... automata) {
+	protected void writeAutomatonToFile(final IAutomaton<L, IPredicate> automaton, final String suffix) {
+		final TaskIdentifier tid = new SubtaskStringIdentifier(null, suffix);
+		this.writeAutomatonToFile(automaton, tid);
```
schuessf: Please omit `this` here (there are more occurences in this file)

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
```diff
@@ -198,10 +199,9 @@ public BasicCegarLoop(final DebugIdentifier name, final A initialAbstraction, fi
 			// Construct an empty file. We need this empty file in cases where
 			// the CFG does not have error location and no automaton is dumped.
 			mLogger.info("Dumping reuse automata for " + mTaskIdentifier.toString());
-			final String filename = mTaskIdentifier + "-reuse";
-			final String fullPath = mPref.dumpPath() + File.separator + filename + "."
-					+ mPrintAutomataLabeling.getFormat().getFileEnding();
-			final File file = new File(fullPath);
+			final String fullAutomataPath = getDumpAutomataPath(new SubtaskStringIdentifier(null, "reuse").toString())
```
schuessf: Is it intended that the first argument is `null` here?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/EagerReuseCegarLoop.java
```diff
@@ -119,7 +119,7 @@ private void computeDifferenceForReuseAutomaton(final int iteration,
 		final int oneBasedi = iteration + 1;
 
 		if (mPref.dumpAutomata()) {
-			writeAutomatonToFile(reuseAut, "ReusedAutomata" + oneBasedi);
+			super.writeAutomatonToFile(reuseAut, "ReusedAutomata" + oneBasedi);
```
schuessf: `super` is not necessary here, right?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/PathProgramDumpController.java
```diff
@@ -105,7 +105,7 @@ private void doDump(final IRun<L, ?> counterexample, final int iteration) {
 			throw createToolchainCanceledException(message, iteration);
 		}
 
-		final String filename = mPref.dumpPath() + File.separator + mIcfg.getIdentifier() + "_" + iteration + ".bpl";
+		final String filename = mPref.dumpAutomataDirectory() + File.separator + mIcfg.getIdentifier() + "_" + iteration + ".bpl";
```
schuessf: This line does not actually dump an automaton, but a path program. So maybe revert the name of this method to `dumpPath` (or something similar) again.

schuessf (approved):
I just had a quick look over these changes. Overall, it looks good, I have just a few comments.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
bahnwaerter: Yes, this is intended. Since we do not have any existing parent task, one has to pass `null` as argument.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
bahnwaerter: Why? I like it when the reference to the current instance is made explicitly visible, so you can see which method is being called.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/EagerReuseCegarLoop.java
bahnwaerter: Yes, in this context, it is not strictly necessary. Nevertheless, it makes it easier to recognize that a method from the superclass is being called.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
schuessf: Our [coding conventions](https://github.com/ultimate-pa/ultimate/wiki/CodingConvention) at least say so 😉 
> Use the keyword `this` only if necessary.

I also thought that our save actions actually replace such occurences.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/PathProgramDumpController.java
bahnwaerter: That is indeed a special case. However, it was already like that in the previous implementation. Back then, the setting was internally called just *dump path*, but the exposed setting was named *automata dump path*. If the `Dumper` for the specific debug text information of each CEGAR iteration is no longer needed, we can remove it entirely (most `Dumper` methods aren't used anymore). @Heizmann do you agree to a removal?

bahnwaerter: Alternatively, we could introduce a separate setting for this. But that leads to having more settings, which isn't so nice either.

schuessf: Okay, if this is not really needed anymore, you could also remove it.
I would definitely not introduce another setting for this. If we want to keep it, we could instead make this setting more general (as the previous `dumpPath` suggested; then also rename the corresponding settings label).

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/AbstractCegarLoop.java
bahnwaerter: I deleted the occurrences of the reference keywords `super` and `this`. It seems that the safe actions only take effect if the file has been modified. When reference keywords have already been added to the code (e.g., via an external editor), the safe actions no longer seem to work.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/EagerReuseCegarLoop.java
bahnwaerter: I deleted the occurrences of the reference keywords `super` and `this`. It seems that the safe actions only take effect if the file has been modified. When reference keywords have already been added to the code (e.g., via an external editor), the safe actions no longer seem to work.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/PathProgramDumpController.java
bahnwaerter: I finally removed the obsolete `Dumper` that misused the automata dump settings. I hope @Heizmann and @maul-esel really don't need it anymore.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
maul-esel: So the previous use of `mTaskIdentifier` was incorrect?

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/PathProgramDumpController.java
maul-esel: I don't remember ever using `Dumper`.

But be careful: I wouldn't expect this kind of debug functionality to be called routinely. But it may still be useful to add calls to it when debugging a specific issue. Just because there are no calls in checked-in code doesn't mean it isn't used from time to time. 🤔

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/BasicCegarLoop.java
bahnwaerter: Yes, because `mTaskIdentifier` was already instantiated with a concatenated string. However, the concept behind a `TaskIdentifier` is to be structured hierarchically. This concept allows each component to append its own suffix (via an additional sub-task identifier), which ensures consistent filenames. This was particularly problematic for BenchExec, as we do not want the input filename hard-coded into the output (similar to the output of the `WitnessPrinter`). Otherwise, it becomes difficult to reuse the input-specific filename when running BenchExec again. To my knowledge, there is no variable in the BenchExec XML run definition for the task's input filename that gets automatically substituted.

trunk/source/TraceAbstraction/src/de/uni_freiburg/informatik/ultimate/plugins/generator/traceabstraction/PathProgramDumpController.java
bahnwaerter: So we should keep the `Dumper` after all?

If necessary, I can revert the removal and either misuse the automata dump settings or introduce another setting for the `Dumper`. What’s the plan?

bahnwaerter: During a meeting with @Heizmann, we agreed to keep the `Dumper` functionality. @Heizmann might use this feature for debugging purposes in the near future. I will review all other settings to identify where additional debug information is stored. This identified location can then be used for the `Dumper` to prevent any confusion with the automata output path setting.

PR #789, opened 2026-06-26
Common settings handling for printer plugins

maul-esel:
Our various printer plugins (`BoogiePrinter`, `ChcSmtPrinter`, `ReqPrinter`) have very similar settings, and the code using these settings has been copy-pasted from one plugin to the other.

This PR removes these duplications and unifies the setting names (which differed slightly). I chose the setting names as used by `ReqPrinter`, as they seemed the sanest choice (no unnecessary special characters like `?` and `:`, cleaner naming i.e. "output" instead of "dump").

---

(The context: In the branch `wip/dk/civlized-OG` we are working on another plugin which will get the same settings and logic; and I didn't want to copy-paste the code yet another time.)


schuessf (approved):
Thanks for your work! These changes look good to me.

PR #797, opened 2026-08-13
Refactor and generalize bitvector constant folding in BitvectorUtils

the-mr-dave:
## Summary
- Replace the per-operator class hierarchy (BitvectorOperation subclasses with
  a forced constant/non-constant split) for bvand/bvor/bvxor with a single
  `bitwiseOperationHelper`: flattens nested applications, folds constant
  operands, cancels/deduplicates non-literal operands (idempotent for
  bvand/bvor, parity-based for bvxor), and applies identity/annihilation
  rules in one pass.
- Add proper n-ary (variadic) support for bvand/bvor/bvxor instead of
  binary-only simplification.
- Mirror bvugt/bvuge/bvsgt/bvsge onto bvult/bvule/bvslt/bvsle by swapping
  operands, so constant folding only needs to be implemented once per
  comparison direction.
- Add and generalize extract-over-extend simplification to partial-width
  extracts.
- Extract and substantially expand test coverage into a dedicated
  BitvectorUtilsTest.

## Test plan
- [x] BitvectorUtilsTest passes
- [x] SimplificationTest passes


the-mr-dave:
Separated it in 3 independent branches

PR #800, opened 2026-08-16
Add extract-over-extend simplification for partial-width extracts

the-mr-dave:
## Summary
- sign_extend/zero_extend only add bits above the original value, so
  extracting bits [hi:0] where hi falls between the original width and the
  full extended width always gives back a smaller extend of the original
  value (or the original value itself, if no padding bits survive).
- Add tests covering the full-undo case, partial reduction, and two guard
  cases (extract of the full extended width, extract of a plain variable)
  that must not trigger the rule.


trunk/source/Library-ModelCheckerUtilsTest/src/de/uni_freiburg/informatik/ultimate/modelcheckerutils/smt/UltimateNormalFormTest.java
the-mr-dave, line 1: Changes of the branch are here

trunk/source/Library-SmtLibUtils/src/de/uni_freiburg/informatik/ultimate/lib/smtlibutils/BitvectorUtils.java
the-mr-dave, line 1: Changes are here

Issue #810, opened 2026-09-22
Bug: precondition check for calls uses incorrect substitution


maul-esel:
When checking the precondition of a called procedure, Ultimate must substitute the actual arguments for the formal parameters of the procedure. This substitution is implemented in the (confusingly named) `WeakestPrecondition` class in `IcfgBuilder` resp. `RCFGBuilder` (the class is duplicated).

However, this substitution is not implemented properly, it does not handle quantified expressions.

## Example

Run the following example **without procedure inlining**.

```bpl
procedure test(x : int)
requires (forall foo : int :: x == foo);
{
  assert false;
}

procedure ULTIMATE.start()
{
  var foo : int;
  call test(foo);
}
```

Expected: `procedure precondition can be violated` (this is the output with inlining enabled)

Actual: `all specifications hold`

## Discussion

The (incorrect) substitution of the precondition results in `(forall foo : int :: foo == foo)`, which of course holds at the call site. However, inside the procedure, we assume the original precondition, which is equivalent to `false`.

I propose that we implement a proper substitution operation for Boogie expressions that can handle quantifiers. This class should ideally live in `Library-BoogieAST`, as it may be useful in other contexts as well (e.g. we'd like to use it in Civlizer). The code specific to call statements (and fork statements, see #809) can call this general-purpose class.

maul-esel:
cc @Heizmann @schuessf @bahnwaerter

schuessf:
Good catch, we should definitely fix this. I just pushed 817ac89, so a fix for this issue only has to touch a single file (`WeakestPrecondition` in `Library-Icfg`).

Issue #809, opened 2026-09-22
Bug: fork statements do not check precondition


maul-esel:
When a Boogie program calls a procedure with a non-free requires clause, Ultimate verifies that the precondition expressed by the requires clause holds at the call site.

However, for fork statements, the precondition is not verified. This leads to unsoundness, as the precondition is assumed at the start of the forked thread.

## Example
```bpl
procedure test(x : int)
requires false;
{
  assert false;
}

procedure ULTIMATE.start()
{
  fork 0 test(42);
  // call test(42);
}
```

Expected: `Procedure precondition can be violated` -- the same message received when uncommenting the `call` statement.

Actual: `All specifications hold.`

[...]

PR #813, opened 2026-09-23
Fix precondition checks for calls and forks

maul-esel:
This PR fixes the following soundness issues:

- #810: Implement a correct, quantifier-aware substitution operation (on Boogie expressions) for computing the condition to check when calling a procedure with a non-free `requires` clause.
- #809: Generate the same kind of check when forking a procedure with a non-free `requires` clause.

I added 2 regression tests that failed in the first nightly build of the branch, but now work correctly in the latest nightly.


schuessf (approved):
Looks good!

